PR #3, opened 2018-08-22
test/rc/abc

JarryShaw:
Migrated `pcapkit.corekit.infoclass.Info` from `builtins.dict` to `collections.abc.Mapping`


JarryShaw:
Manually merged #3

PR #373, opened 2026-09-14
Advance past the last IPv6 extension header (#348)

### LINKED ISSUE #348: IPv6 fragment header not skipped: next layer decodes it as transport
### What is wrong

The IPv6 extension-header traversal in `IPv6._decode_next_layer` (`pcapkit/protocols/internet/ipv6.py`) leaves the loop on a fragment header **before** advancing `payload` past that header, so the next layer is handed the fragment header's own eight octets and decodes them as the transport header:

```python
        payload = self.__header__.get_payload()
        while True:
            ...
            # keep record of extension headers
            self._exthdr.add(ex_proto, next_)

            # keep original data after fragment header
            if ex_proto == Enum_ExtensionHeader.IPv6_Frag:            # line 348
                ipv6.__update__({
                    'fragment': self._read_packet(header=hdr_len, payload=raw_len),
                })
                break                                                 # line 352 -- payload not advanced

            # update payload for next extension header
            payload = payload[next_.length:]                          # line 355 -- skipped
        ...
        return super()._decode_next_layer(ipv6, proto, raw_len, packet=packet,
                                          ipv6_exthdr=ipv6_exthdr, payload=payload)   # line 367
```

Every other exit from the loop is fine — the `except ValueError: break` at line 319 fires *after* line 355 has run for the previous header. Only the fragment branch skips it. `ipv6.info.fragment` is built correctly (`_read_packet` offsets it by `hdr_len`); it is the `payload` handed to `super()._decode_next_layer` at line 367 that is eight octets stale.

### Symptom

On the committed `examples/sample/ipv6.pcap` fixture, frame 12 — the first fragment of a 4778-octet UDP datagram — the UDP layer is decoded from the fragment header:

```
protochain          : Ethernet:IPv6:IPv6-Frag:UDP:Raw
bytes the UDP layer was given: 11 00 00 01 00 01 ae e4
UDP srcport         : 4352      <- 0x1100 = next-header (0x11 = UDP) and reserved (0x00)
UDP dstport         : 1         <- 0x0001 = fragment offset 0 with the M flag
UDP len             : 1         <- 0x0001 = high half of identification 110308
```

The real UDP header starts eight octets later, and `ipv6.info.fragment.payload` does hold it: `c8 22 13 89 12 aa 0f 75` — srcport 51234, dstport 5001, len 4778. So the datagram is being reported with a plausible-looking but entirely fabricated transport header, and application-layer dispatch then happens off those fake port numbers.

### Reproduction

```python
import pcapkit

ext = pcapkit.extract(fin='examples/sample/ipv6.pcap', store=True, nofile=True)
ipv6 = ext.frame[12].payload.payload
udp = ipv6.payload

print(bytes(udp.packet.header).hex(' '))     # 11 00 00 01 00 01 ae e4  <- the fragment header
print(int(udp.info.srcport), int(udp.info.dstport), udp.info.len)   # 4352 1 1
print(bytes(ipv6.info.fragment.payload)[:8].hex(' '))   # c8 22 13 89 12 aa 0f 75  <- the real UDP header
```

(`examples/sample/ipv6.pcap` is written by the generator in `examples/samples/pcap.py`; its fragmented UDP datagram is 1448/1448/1448/434 octets with identification 110308.)

### What a fix would need to touch

Advancing `payload` before the `break` — i.e. moving line 355 above the fragment branch, or repeating it inside — is the change. Patching the method that way at runtime and re-parsing the same fixture gives:

```
protochain          : Ethernet:IPv6:IPv6-Frag:UDP:Raw
hdr_len / raw_len   : 48 1448
UDP header bytes    : c8 22 13 89 12 aa 0f 75
srcport/dstport/len : 51234 5001 4778
udp payload type    : Raw
fragment payload len: 1448
```

i.e. the real UDP header, with everything else about the frame unchanged and no new warnings. But note:

**`tests/protocols/internet/test_ipv6_extension_runtime.py` currently pins this behaviour**, so fixing it requires updating that test in the same change. `test_ipv6_fragment_chain_records_extension_header_metadata` asserts the fabricated values directly:

```python
        self.assertEqual(int(udp.info.srcport), 4352)
        self.assertEqual(int(udp.info.dstport), 1)
        self.assertEqual(udp.info.len, 1)
```

Those three lines become 51234 / 5001 / 4778 once the offset is fixed, per the patched run above. The surrounding assertions (`protochain`, `hdr_len == 48`, `raw_len == 1448`, `fragment.payload` length 1448) are already correct and do not change.

Whether the first fragment's payload *should* be dispatched to UDP at all is a separate design question — a reassembler would normally decode the transport header only from the fragment at offset 0 — but the bytes it is decoded from are wrong either way.

### How it was found

While building deterministic sample captures for the test suite; the generators' module docstrings record it. `examples/samples/pcap.py` flags the 4352/1/1 values explicitly, so that they are not mistaken for a property of the fixture.

Line numbers are from `main` at 72f950d1b.

JarryShaw:
Draft. Independent of #371 and #372 — one file plus its tests.

## The defect

`pcapkit/protocols/internet/ipv6.py` broke out of the extension-header walk *before* advancing the payload, so the next layer received the bytes starting at the last extension header rather than after it. On a fragmented datagram that means the fragment header gets decoded as the transport header:

| | before | after |
|---|---|---|
| bytes handed to UDP | `11 00 00 01 00 01 ae e4` | `c8 22 13 89 12 aa 0f 75` |
| srcport / dstport / len | 4352 / 1 / 1 | 51234 / 5001 / 4778 |

Those numbers are not a malformed header: `4352` is `0x1100`, the fragment header's next-header (17) and reserved octet; `1` is its offset-and-flags hextet; the length is the top half of identification 110308.

The advance now runs before every exit from the loop. Each branch was checked rather than assumed — the fall-through to another extension header must advance (unchanged, only moved earlier), the fragment `break` must advance and did not, the `except ValueError` exit must not advance again and does not, and a datagram with no extension headers never enters the loop.

## An existing test was pinning the bug

`tests/protocols/internet/test_ipv6_extension_runtime.py` asserted the wrong ports and length; it now asserts the real header, plus an invariant (`bytes(udp.packet.header) == bytes(ipv6.info.fragment.payload)[:8]`) instead of three more magic constants. `test_ip_runtime.py` needed no change — its `ipv6.pcap` case is an ICMPv6 frame with no extension headers.

`examples/generators/pcap.py`'s module docstring documented the odd ports as expected behaviour and now describes what the fixture contains. Docstring only; the captures regenerate byte-identically.

## Regression test

A subtested table over four chains — none, fragment only, hop-by-hop plus destination options, and hop-by-hop plus a 16-octet routing header plus fragment — asserting the extension-header list, `hdr_len`, `raw_len`, and that UDP was handed exactly the UDP header. The datagrams are synthetic and parsed straight through `IPv6(...)`, because no committed capture carries more than one extension header and the fixtures are pinned byte-for-byte. The routing header is deliberately 16 octets, so a walk advancing by a fixed 8-octet stride instead of each header's own length is also caught.

Confirmed to catch the bug: with the fix reverted, the two fragment chains fail and the two non-fragment chains pass.

## Two incidental defects found and **not** fixed here

Both in `pcapkit/protocols/internet/ipv6_route.py`, both independent of this change and worth their own issues: well-formed routing headers are rejected because `_read_data_type_2` compares the schema's raw Hdr Ext Len octet against 24 (it is 2 for an RFC 6275 type-2 header), with the same shape in `_read_data_type_src` and `_read_data_type_rpl`; and a malformed extension header crashes the walk with `AttributeError: 'Raw' object has no attribute 'next'` when `_import_next_layer` degrades to `Raw`, rather than ending it.

Closes #348.


JarryShaw:
@copilot Fix the code for all comments in [this review thread](https://github.com/JarryShaw/PyPCAPKit/pull/373#pullrequestreview-5199280264).

Issue #364, opened 2026-09-14
Constructing a pcapkit warning mutates global warnings.filters


JarryShaw:
### Summary

`BaseWarning.__init__` calls `warnings.simplefilter('ignore', type(self))` when not in devmode. `warnings.simplefilter` mutates the **process-global** `warnings.filters` list and inserts at index 0, so merely constructing a pcapkit warning — or parsing one packet — permanently overrides whatever warning configuration the host application, the test runner, or `python -W` had set for that category. It also invalidates every module's `__warningregistry__`, which makes unrelated non-pcapkit warnings re-fire.

A side effect is that pcapkit's own warnings are invisible by default, and there is no way for a caller to re-enable them: any filter they install is overtaken by the next warning construction.

### file:line

`pcapkit/utilities/warnings.py:74`, in `BaseWarning.__init__` (lines 65-77):

```python
65    def __init__(self, *args: 'Any', **kwargs: 'Any') -> 'None':  # pylint: disable=useless-super-delegation
66        # log warning
67        if DEVMODE:
68            if VERBOSE:
69                logger.warning(str(self), exc_info=self, stack_info=True,
70                            stacklevel=stacklevel_calculator())
71            else:
72                logger.warning(str(self))
73        else:
74            warnings.simplefilter('ignore', type(self))
75
76        # warnings.simplefilter('default')
77        super().__init__(*args, **kwargs)
```

The offending expression, executed from a constructor:

```python
warnings.simplefilter('ignore', type(self))
```

### Observable symptom 1 — the global filter list changes

Constructing one instance. No `warn()` call, no parse. Python 3.14.7:

```
warnings.filters BEFORE any pcapkit warning:  (5 entries)
  [0] ('default', None, <class 'DeprecationWarning'>, '__main__', 0)
  [1] ('ignore', None, <class 'DeprecationWarning'>, None, 0)
  [2] ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0)
  [3] ('ignore', None, <class 'ImportWarning'>, None, 0)
  [4] ('ignore', None, <class 'ResourceWarning'>, None, 0)

--- constructing ONE pcapkit warning instance, nothing else ---
constructed: SchemaWarning('just constructing the object')

warnings.filters AFTER:  (6 entries)
  [0] ('ignore', None, <class 'pcapkit.utilities.warnings.SchemaWarning'>, None, 0)
[...]

PR #386, opened 2026-09-15
Add the pypcap and pypcapfile extraction engines

JarryShaw:
Closes the New Engines item on the Help Wanted page — both candidates it names are now selectable, each with a matching `pcapkit.toolkit` module, a `pyproject.toml` extra, docs and tests.

## Registration: built-in, not the public hook

Both are registered in `Extractor.__engine__` as `ModuleDescriptor`s, like `dpkt`/`scapy`/`pyshark`. That is not merely for consistency — **the public auto-registration path cannot work for an engine shipped with the library.** `pcapkit/__init__.py` never imports `pcapkit.foundation.engines`, so `Engine.__init_subclass__` would never fire for a bare `import pcapkit`; `Extractor(engine='pypcap')` would then fall through to "unsupported engine" and silently use the built-in parser — precisely the trap these engines exist to avoid. `ModuleDescriptor` also keeps the third-party import lazy, so `import pcapkit` stays free of `pcap`/`pcapfile`. The public path remains the documented route for user engines and keeps its own coverage.

## Capability gaps are surfaced, not papered over

**`pypcap`** is a libpcap binding with zero dissection:
- No reassembly or flow tracing — `run()` clears the flags, replaces the managers, and warns; the four toolkit adapters raise `UnsupportedCall`.
- **PCAP only.** Measured: libpcap 1.11 opens a PCAP-NG file happily, reports `datalink()==1`, and yields **0 frames** — a silently empty capture. So the engine gates on the magic number and raises `FormatError` instead.
- File-on-disk only (`pcap_open_offline` opens by name), so non-file input raises `UnsupportedCall`.
- `follow_tcp_stream(engine='pypcap')` now falls back with an `EngineWarning`, as it cannot trace.

**`pypcapfile`** decodes Ethernet/IPv4/TCP/UDP only:
- No IPv6 — `run()` clears the flag and warns, and `ipv6_reassembly` **raises** rather than returning `None`, which would be indistinguishable from "no fragment here".
- Design note: it loads with `layers=0` and decodes to depth 2 itself, because pypcapfile's decoders *replace* payload bytes; stopping at the network layer keeps the TCP segment verbatim so the header/payload split is exact rather than reconstructed. The IPv4 header *is* reconstructed, and that reconstruction is asserted byte-exact against real packets including options and fragment flags.

## Neither installs cleanly from PyPI, and `pep.rst` says so

- **`pypcapfile`** 0.12.0 imports `imp`, removed in Python 3.12, so it cannot be imported at all on 3.12+. Upstream master (0.12.1, unreleased) fixes it, and that is what the engine was verified against.
- **`pypcap`** 1.3.0 ships a pre-generated `pcap.c` from an old Cython that no longer compiles against the 3.12+ C API (`ob_digit`, `_PyLong_AsByteArray`), and its `setup.py` searches a fixed list of prefixes for `pcap.h`. It builds once the headers are visible and `pcap.c` is regenerated with Cython 3.

Both were **genuinely executed** here, on a locally rebuilt wheel — neither is claimed as unverified. The caveats are about installability, not coverage, and the note in `pep.rst` states them rather than claiming both are simply done.

## Parity, and proving the engine actually ran

`tests/foundation/engines/test_new_engine_parity.py` runs over `in.pcap`, `arp.pcap`, `tcp.pcap`, `ipv4.pcap` (1137 frames) and asserts, per frame: capture length, original length, timestamp, and the full Ethernet header — all exact. Every parity test goes through a helper that patches `warn`, asserts **no `EngineWarning` was raised**, and asserts `extractor._exnam` is the engine requested — because `Extractor` does not raise when an engine's package is missing, it warns and silently falls back, so a naive test would happily report parity for work the built-in parser did. `pypcapfile` IPv4 reassembly is additionally asserted byte-identical to the default engine.

## One pre-existing crash fixed, because nothing worked without it

`Extractor(trace=True)` died with `AttributeError: 'dict' object has no attribute 'packet'`. The guard at `extraction.py` listed only `pyshark` **and** only matched `trace_format == 'pcap'`, while `TraceFlow.__init__` substitutes `'pcap'` for `None` — so it never fired. It now matches `('pcap', 'cap', None)`, which incidentally fixes `pyshark` too. **`dpkt` and `scapy` remain broken there by design** and are called out in a code comment; that is its own change.

Also worth knowing for future tests: `BaseWarning.__init__` calls `warnings.simplefilter('ignore', type(self))` outside devmode, so `assertWarns` can never fire on a pcapkit warning. These tests patch `warn` and inspect the call instead. (That behaviour is #364.)

## Verification

56 new test functions across 5 files. **Full suite on this branch: 550 passed, 18 skipped, 335 subtests** — the 18 skips are exactly the availability-gated engine tests, since the project venv deliberately does not carry either library. In a venv with both installed the same suite runs 8 more tests.


JarryShaw:
@copilot Fix the code for all comments in [this review thread](https://github.com/JarryShaw/PyPCAPKit/pull/386#pullrequestreview-5205334685).

JarryShaw:
Copilot's four typo comments were already fixed by its own commit `1cbb9e726`; I fast-forwarded onto it rather than replaying over it. One heads-up though, because it changes what an extra means.

**`pypcap` had to come out of the `all` extra.** `deploy-pages` failed on this branch, and it is not a flake:

```
Getting requirements to build wheel ... error
  Found pcap headers in /Applications/.../MacOSX26.sdk/usr/include/pcap.h
  None of the following found: ['libpcap.a', 'libpcap.so', 'libpcap.dylib', 'wpcap.lib']
ERROR: Failed to build 'pypcap' when getting requirements to build wheel
```

`pypcap` 1.3.0 publishes **no wheels at all** — PyPI has only an sdist — so `pip install pypcapkit[all]` now compiles a C extension against libpcap and needs the headers *and* the shared library. The macOS runner has the header and no library, which is exactly the failing case.

`deploy-pages` was only the first to hit it. `.[all]` is installed by five workflows:

| workflow | occurrences |
|---|---|
| `deploy-pages` | 1 |
| `cron-conda` | 3 |
| `cron-vendor` | 1 |
| `create-release` | 2 |

So the release and conda paths would have broken on merge too, and every user running `pip install pypcapkit[all]` would need a compiler and libpcap dev files.

`pypcap` stays as its own opt-in extra — `pip install pypcapkit[PyPCAP]` — documented as needing libpcap present. `pypcapfile` is sdist-only as well but pure Python, so it installs anywhere and stays in `all`. Also dropped from the `Pipfile` dev-packages, where it would break `pipenv install --dev` and even `pipenv lock` on any machine without libpcap — this one included.

Worth noting `pep.rst` already said `pip install pypcapkit[PyPCAP]` "can fail to build" while also claiming both extras were in `all`; that contradiction is now resolved in favour of what the packaging does.

Verified: `pip install --dry-run '.[all]'` on this libpcap-less host resolves and would install only `pypcapfile`. Unit tier 438 passed / 14 skipped; engines and toolkit 74 passed / 14 skipped.

Unrelated and left alone: `seperated` also appears pre-existing in `toolkit/scapy.py`, `toolkit/dpkt.py`, `protocols/misc/pcapng.py`, `protocols/link/ethernet.py` and `protocols/link/arp.py` — the new modules had copied it from their siblings. Out of scope here.

PR #390, opened 2026-09-15
Report each warning once per channel; stop mutating global warning state (#362-#364)

### LINKED ISSUE #362: MultiDict.get() on an absent key logs an ERROR record per lookup
### Summary

`MultiDict.get()` / `OrderedMultiDict.get()` emit a `logging` record at **ERROR** level every time a key is absent, then correctly return the default. On a successful parse of a capture containing unfragmented IPv6 with reassembly enabled, this puts one ERROR line on stderr per IPv6 frame, for what the calling code itself describes as the ordinary case.

### file:line

`pcapkit/utilities/exceptions.py:112-113`, in `BaseError.__init__` (lines 100-117):

```python
100    def __init__(self, *args: 'Any', quiet: 'bool' = False, **kwargs: 'Any') -> 'None':
101        # log error
102        if not quiet:
103            if DEVMODE:
104                logger.critical('%s: %s', type(self).__name__, str(self),
105                                exc_info=self if VERBOSE else False,
106                                stack_info=VERBOSE, stacklevel=-stacklevel())
107            else:
108                logger.critical("%s: %s", type(self).__name__, str(self))
109
110            # logger.error('%s: %s', type(self).__name__, str(self), exc_info=self,
111            #              stack_info=True, stacklevel=-stacklevel())
112        else:
113            logger.error('%s: %s', type(self).__name__, str(self))
```

The offending expression is line 113:

```python
logger.error('%s: %s', type(self).__name__, str(self))
```

`quiet=True` selects that branch. So `quiet` currently means "log at ERROR instead of CRITICAL", not "do not log".

### Intent vs behaviour

`docs/source/pcapkit/utilities/exceptions.rst:14` documents the parameter as:

```
   :param quiet: If :data:`True`, suppress exception message.
```

### How an ordinary lookup reaches it

`pcapkit/corekit/multidict.py:210-214` — `get` is defined on `MultiDict` and inherited by `OrderedMultiDict`:

```python
210    def get(self, key: '_KT', default: '_VT | _T' = None) -> '_VT | _T':
211        try:
212            return self[key]
213        except MissingKeyError:
214            return default
```

Both `__getitem__` implementations raise with `quiet=True`:

```python
183        raise MissingKeyError(key, quiet=True)   # MultiDict.__getitem__        (169-183)
514        raise MissingKeyError(key, quiet=True)   # OrderedMultiDict.__getitem__ (511-514)
```

The IPv6 caller is `pcapkit/toolkit/pcap.py:101-104` (identical code in `pcapkit/toolkit/pcapng.py:104-107`), whose own comment says a miss is expected:

```python
101        if (ipv6_frag := ipv6.extension_headers.get(  # type: ignore[call-overload]
102            Enum_ExtensionHeader.IPv6_Frag
103        )) is None:  # dismiss not fragmented frame
104            return None
```

### Observable symptom

`pcapkit/utilities/logging.py:49-55` attaches a `StreamHandler(sys.stderr)` at level INFO unconditionally, so these records reach the terminal with no logging configuration at all.

On `examples/captures/ipv6.pcap` (16 frames, parse succeeds):

```
[INFO] 09/14/2026 12:24:22 AM - IPv6 reassembly enabled
[ERROR] 09/14/2026 12:24:22 AM - MissingKeyError: <ExtensionHeader.IPv6_Frag: 44>
[ERROR] 09/14/2026 12:24:22 AM - MissingKeyError: <ExtensionHeader.IPv6_Frag: 44>
[ERROR] 09/14/2026 12:24:22 AM - MissingKeyError: <ExtensionHeader.IPv6_Frag: 44>
... (12 in total)
[WARNING] 09/14/2026 12:24:22 AM - EOF reached
frames parsed: 16
reassembled ipv6 datagrams: 1
```

Counted programmatically (Python 3.14.7):

| capture | frames | records by level | `MissingKeyError` ERROR records |
|---|---|---|---|
| `examples/captures/ipv6.pcap` | 16 | `{'INFO': 1, 'ERROR': 12, 'WARNING': 1}` | 12 |
| `examples/captures/http6.cap` | 26 | `{'INFO': 1, 'ERROR': 26, 'WARNING': 3}` | 26 |

Every one of them is the identical string `MissingKeyError: <ExtensionHeader.IPv6_Frag: 44>`.

Scope, measured on `ipv6.pcap` — IPv6 reassembly has to be on:

```
config                             frames  MissingKeyError ERRORs
default (no reassembly)                16        0
reassembly=True only                   16        0
reassembly=True, ip=True               16       12
reassembly=True, ipv6=True             16       12
reassembly=True, ipv4=True             16        0
```

This corrects the way it was first described to me as happening on "any capture containing unfragmented IPv6" — reassembly must be enabled, and the CLI exposes no reassembly flag, so it is a library-API path. The minimal form below has no preconditions at all, though.

### Minimal reproduction

No capture and no reassembly needed:

```python
import logging

from pcapkit.corekit.multidict import OrderedMultiDict


class Sink(logging.Handler):
    def __init__(self):
        super().__init__()
        self.records = []

    def emit(self, record):
        self.records.append(record)


sink = Sink()
logging.getLogger('pcapkit').addHandler(sink)

result = OrderedMultiDict().get('absent')

print('returned:', repr(result))
for r in sink.records:
    print(r.levelname, '|', r.getMessage())
```

Output:

```
returned: None
ERROR | MissingKeyError: 'absent'
```

The capture-level form:

```python
import pcapkit

pcapkit.extract(fin='examples/captures/ipv6.pcap', nofile=True, store=True,
                reassembly=True, ipv6=True)
```

### Related side effect in the same constructor

Lines 115-116 run on **both** branches, so the same absent-key `.get()` also sets `sys.tracebacklimit = 0` for the rest of the process, truncating tracebacks for entirely unrelated exceptions:

```
sys.tracebacklimit before: <unset>
traceback lines for an unrelated ValueError, before: 14
OrderedMultiDict().get("absent") returned: None
sys.tracebacklimit after : 0
traceback lines for the same unrelated ValueError, after: 1
```

### What a fix would need to touch

Two defensible places; the second is arguably the root cause.

1. `pcapkit/utilities/exceptions.py`, `BaseError.__init__`, the `else` branch at 112-113 — make `quiet=True` mean what the docs say (emit nothing, or emit at `DEBUG`). Also decide whether `sys.tracebacklimit = 0` at 115-116 should apply on the quiet path.
2. `pcapkit/corekit/multidict.py` — `MultiDict.get` (210-214) with `MultiDict.__getitem__` (169-183) and `OrderedMultiDict.__getitem__` (511-514): a `.get()` miss is control flow, not an error, and need not construct a logging exception at all. Fixing here removes the per-lookup cost as well as the noise and leaves `quiet` semantics alone.

Other callers passing `quiet=True`, which a semantics change would affect: `pcapkit/corekit/multidict.py:183`, `pcapkit/corekit/multidict.py:514`, `pcapkit/protocols/protocol.py:835`.

Tests pinning current behaviour: `tests/utilities/test_exceptions_warnings.py:22-25` and `:27-38`.

### Observed vs inferred

- **Observed by running code:** the single ERROR record from a bare `.get()` miss on both `MultiDict` and `OrderedMultiDict`; the correct `None` return; 12 and 26 ERROR records on `ipv6.pcap` and `http6.cap`; that they appear on stderr with no logging configuration; the trigger matrix; the `sys.tracebacklimit = 0` side effect and its effect on an unrelated traceback.
- **Inferred by reading code:** that `docs/source/pcapkit/utilities/exceptions.rst:14` is the authoritative statement of what `quiet` is meant to do; that `pcapkit/toolkit/pcapng.py:104-107` behaves the same as the `pcap.py` path (identical code, not separately exercised).

### Related

Two other defects in the same family, filed separately because no single change fixes more than one of them: the double/triple emission in `pcapkit/utilities/warnings.py:52-54`, and the process-global `warnings.filters` mutation at `pcapkit/utilities/warnings.py:74`. Note the structural rhyme with the latter — both are constructors with unscoped process-global side effects.

### LINKED ISSUE #363: warn() emits each warning twice, and three times in devmode
### Summary

`pcapkit.utilities.warnings.warn()` writes to the `pcapkit` logger and then calls `warnings.warn()`. Constructing the warning category runs `BaseWarning.__init__`, which logs a **second** time when `PCAPKIT_DEVMODE` is set. One `warn()` call therefore produces three observable emissions in devmode. All 97 call sites in the package inherit this.

The underlying question is a design choice — which channel should own a pcapkit warning — not a mechanical correction. See "What a fix would need to touch" below; this issue deliberately does not prescribe an answer.

### file:line

`pcapkit/utilities/warnings.py:52-54`, in `warn()` (lines 39-54):

```python
49    if stacklevel is None:
50        stacklevel = stacklevel_calculator()
51
52    logger.warning(message, exc_info=VERBOSE, stack_info=VERBOSE,
53                   stacklevel=stacklevel)
54    warnings.warn(message, category, stacklevel)
```

The third emission comes from `pcapkit/utilities/warnings.py:69-72`, in `BaseWarning.__init__` (lines 65-77):

```python
67        if DEVMODE:
68            if VERBOSE:
69                logger.warning(str(self), exc_info=self, stack_info=True,
70                            stacklevel=stacklevel_calculator())
71            else:
72                logger.warning(str(self))
```

### Observable symptom

Measured counts for **one** `warn()` call, in separate processes, Python 3.14.7:

| category passed | `PCAPKIT_DEVMODE` | logging records | `warnings`-module emissions | total |
|---|---|---|---|---|
| `SchemaWarning` (a `BaseWarning`) | unset | 1 | 0 | 1 |
| `SchemaWarning` | `1` | 2 | 1 | **3** |
| `UserWarning` (not a `BaseWarning`) | unset | 1 | 1 | **2** |
| `UserWarning` | `1` | 1 | 1 | 2 |

Devmode, with each emission attributed to its source line:

```
=== ONE call: warn("probe", SchemaWarning) ===

logging records: 2
  message='probe'
    emitted from pcapkit/utilities/warnings.py:52 in warn()  ->  logger.warning(message, exc_info=VERBOSE, stack_info=VERBOSE,
  message='probe'
    emitted from pcapkit/utilities/warnings.py:72 in __init__()  ->  logger.warning(str(self))
warnings-module emissions: 1
    pcapkit/utilities/warnings.py:54  SchemaWarning: probe
```

Three emissions from lines 52, 72 and 54.

**One nuance worth knowing before measuring this yourself, and a correction to how it was first reported to me.** The reported "two records normally" is **not** reproducible: in non-devmode with a `BaseWarning` category the count is **1**, not 2, because the `warnings`-module emission is swallowed. That is not because the duplication is absent but because `warnings.simplefilter('ignore', type(self))` at `pcapkit/utilities/warnings.py:74` runs during `category(message)` construction, which CPython performs *before* consulting `warnings.filters`. Measured in one process: `len(warnings.filters)` went `5 -> 6` while the `warnings` display hook was called 0 times. Passing a category that is not a `BaseWarning` subclass bypasses that and shows the raw duplication:

```
--- ONE warn("...", UserWarning) ---
  logging records          : 1 ['probe for UserWarning']
  warnings-module emissions: 1 ['UserWarning: probe for UserWarning  (pcapkit/utilities/warnings.py:54)']
  warnings.filters len     : 5 -> 5
  TOTAL observable         : 2

--- ONE warn("...", SchemaWarning) ---
  logging records          : 1 ['probe for SchemaWarning']
  warnings-module emissions: 0 []
  warnings.filters len     : 5 -> 6
  TOTAL observable         : 1
```

So fixing the filter mutation in isolation would raise the non-devmode count from 1 to 2. Devmode's "three" is confirmed exactly.

### This is already worked around in-tree

`examples/generators/pcapng.py:865-876`:

```
class _LogCapture(logging.Handler):
    """Collect pcapkit's own log records rather than letting them print.

    pcapkit reports a parse complaint twice -- once through :mod:`warnings` and
    once through its ``pcapkit`` logger -- but only the logger call is reliable:
    it passes a computed ``stacklevel`` to :func:`warnings.warn` that can point
    outside the stack, and the record then never reaches a
    :func:`warnings.catch_warnings` recorder. So the logger is what this listens
    to. Duplicates are dropped, because one complaint per block in a capture
    with sixty-four packets is not sixty-four pieces of information.
    """
```

### Blast radius

97 call sites of `warn()` across 32 files in `pcapkit/`, counted by AST (`ast.Call` whose `func` is an `ast.Name` named `warn`), all in modules that import `pcapkit.utilities.warnings.warn`. Heaviest: `protocols/misc/pcapng.py` 14, `foundation/extraction.py` 12, `protocols/internet/ipv4.py` 10, then `protocols/internet/hopopt.py`, `protocols/internet/ipv6_opts.py`, `protocols/internet/mh.py` and `vendor/default.py` at 5 each. Every call site passes a `BaseWarning` subclass. Only the three lines quoted above are at fault; no call site needs editing.

### Minimal reproduction

```python
import logging
import warnings as stdlib_warnings

from pcapkit.utilities.warnings import SchemaWarning, warn


class Sink(logging.Handler):
    def __init__(self):
        super().__init__()
        self.records = []

    def emit(self, record):
        self.records.append(record)


sink = Sink()
logging.getLogger('pcapkit').addHandler(sink)

shown = []
stdlib_warnings.showwarning = lambda msg, cat, fn, ln, file=None, line=None: \
    shown.append('%s: %s (%s:%d)' % (cat.__name__, msg, fn, ln))

warn('probe', SchemaWarning)

print('logging records          :', len(sink.records))
print('warnings-module emissions:', len(shown), shown)
print('total                    :', len(sink.records) + len(shown))
```

Run it twice, in separate processes — pcapkit reads its env flags at import time:

```
$ python repro.py
logging records          : 1
warnings-module emissions: 0 []
total                    : 1

$ PCAPKIT_DEVMODE=1 python repro.py
logging records          : 2
warnings-module emissions: 1 ['SchemaWarning: probe (.../pcapkit/utilities/warnings.py:54)']
total                    : 3
```

### What a fix would need to touch — a design choice, not a one-liner

The fault is three lines, but choosing among the fixes settles which channel owns a pcapkit warning, and each answer costs something real. This needs a maintainer decision rather than a patch chosen by whoever gets there first.

Files: `pcapkit/utilities/warnings.py` — `warn()` (39-54) and `BaseWarning.__init__` (65-77). No call site changes under any option.

- **Option A — the logger owns it.** Drop line 54. Cost: `-W`, `python -W error`, pytest's `filterwarnings`, and `warnings.catch_warnings` stop seeing pcapkit warnings entirely, and pcapkit stops participating in the standard Python warning contract.
- **Option B — the `warnings` module owns it.** Drop 52-53 and the devmode logging at 69-72, and point users at `logging.captureWarnings(True)`. Cost: consumers that rely on the `pcapkit` logger lose the complaints — including `examples/generators/pcapng.py:865`, which listens to the logger precisely *because* the `warnings` channel is unreliable here. That unreliability (the computed `stacklevel` from `stacklevel_calculator()` can point outside the stack) would have to be fixed first.
- **Option C — keep both channels, remove only the triple.** Drop the logging in `BaseWarning.__init__` (69-72). Cheapest; leaves the normal double in place.

Tests pinning current behaviour: `tests/utilities/test_exceptions_warnings.py:57-79` asserts one recorded warning *and* a `logger.warning` call, and separately that `BaseWarning('dev warning')` logs exactly once. Options B and C break it.

The counts observed today also depend on the process-global filter mutation at `pcapkit/utilities/warnings.py:74`, filed separately. Decide the two together; they are still two changes in two functions.

### Observed vs inferred

- **Observed by running code:** every row of the count table, in separate processes; the per-line attribution of the three emissions to `warnings.py:52`, `:72` and `:54`; that `BaseWarning.__init__` runs exactly once per `warn()` call and before the filter search (`filters` 5 -> 6 with 0 display-hook calls); the 97 call sites across 32 files, counted by AST.
- **Inferred by reading code:** that fixing the filter mutation alone would take the non-devmode count from 1 to 2 (inferred from the `UserWarning` control, which bypasses `BaseWarning.__init__`, not from a patched build); the costs listed against each option; the `examples/generators/pcapng.py` claim about `stacklevel` pointing outside the stack, which was read but not independently measured.

### Related

Same family, filed separately: the `warnings.filters` mutation at `pcapkit/utilities/warnings.py:74`, and the ERROR-level logging of `quiet=True` exceptions at `pcapkit/utilities/exceptions.py:112-113` (#362).

### LINKED ISSUE #364: Constructing a pcapkit warning mutates global warnings.filters

JarryShaw:
Three entangled defects, fixed together because they are one subsystem: how many times a warning is reported, on which channel, and what `quiet=True` means.

## The emission model

**A warning is reported once per channel; construction is not a report.**

| | records | channel | governed by |
|---|---|---|---|
| `warn()` → logging | exactly **1**, `WARNING` | unconditional, ignores `warnings.filters` | `logging.getLogger('pcapkit')` |
| `warn()` → warnings | exactly **1** | filtered normally | `filterwarnings`, `catch_warnings`, `-W`, pytest |
| constructing `SchemaWarning(...)` | **0** | — | — |

Counts are now **identical in devmode and out** — `PCAPKIT_DEVMODE` changes the *richness* of the record (`exc_info`, `stack_info`), never the count.

A logger-only model was rejected because suppression must remain available through the standard `warnings` machinery; a warnings-only model was rejected because `examples/generators/pcapng.py` deliberately listens to the *logger*, and it would depend on first fixing `stacklevel()` (see below).

## #364 — constructing a warning mutated `warnings.filters`

`BaseWarning.__init__` called `warnings.simplefilter('ignore', type(self))` outside devmode, inserting an ignore filter at **index 0** of the process-global list. So it overrode the host application's own `-W` configuration, invalidated `__warningregistry__` so unrelated third-party warnings re-fired, and made pcapkit's own warnings invisible by default — which cannot have been the intent. The constructor is gone; no filter is ever installed.

Measured, non-devmode:

| probe | before | after |
|---|---|---|
| filters after constructing all 23 categories | +22 entries | **+0, `filters == snapshot`** |
| app's `simplefilter('always')` then `warn()` | 0 recorded | **1 recorded** |
| app's `simplefilter('error')` then `warn()` | no exception | **raises** |
| unrelated third-party warning re-fires | **True** | **False** |
| `extract()` end-to-end on http6.cap | re-fired, filters +1 | not re-fired, filters unchanged |

I confirmed the headline independently on this branch: 23 categories constructed, `filters unchanged: True`, app filter wins with 1 record.

## #363 — each warning emitted more than once

Both `warn()` and the constructor logged. Non-devmode was **1** rather than 2, because #364's own `simplefilter` swallowed the `warnings` emission (CPython constructs the category before consulting the filters) — so the two issues were entangled and fixing one changed the other's observable behaviour. Devmode was exactly 3. Both are now 2: one log record, one warning.

## #362 — `quiet=True` still logged

It meant "log at `ERROR` instead of `CRITICAL`", which is why `MultiDict.get()` on an absent key logged an error *per lookup* — **12 on one `ipv6.pcap` reassembly, 26 on `http6.cap`**. It now means what it says: no record at any level, and `sys.tracebacklimit` left alone, since a quiet error is pcapkit's own control flow that it catches itself. A loud error is unchanged: one `CRITICAL`, and `tracebacklimit = 0` outside devmode. Verified here: a quiet miss produces empty log output.

## Behaviour changes a consumer could notice

1. **pcapkit warnings now reach the `warnings` channel** — 43 new warning sites in pytest's summary. Restore with `warnings.filterwarnings('ignore', category=BaseWarning)`.
2. **The application's filter now wins, including `-W error`.** A project running `-W error::UserWarning`, or pytest with `filterwarnings = error`, will now get **exceptions** where pcapkit warnings previously vanished. The sharpest edge; documented under an `.. attention::` block.
3. **Devmode: 2 log records per warning → 1.**
4. **Constructing a `BaseWarning` subclass logs nothing** (was 1 in devmode).
5. **`quiet=True` exceptions emit no `ERROR` record** — not restorable by configuration, so anything grepping logs for `MissingKeyError` must catch the exception instead.
6. **`sys.tracebacklimit` is no longer set by a quiet exception**, so an uncaught `MultiDict()['missing']` prints a normal traceback instead of one line.
7. `pcapkit.utilities.warnings.DEVMODE` re-export removed (never in `__all__`; canonical home is `pcapkit.utilities.logging.DEVMODE`).

`BaseWarning.__init__` no longer existing was checked for behavioural equivalence: `args`, `str()`, pickle round-trip, MRO, and even the `TypeError` message for kwargs are unchanged.

## Verification

20 tests across three new files under `tests/utilities/`, plus helpers. Two details worth noting: the re-fire probe carries a **control arm** asserting it can distinguish the two cases, so a failure is attributable to pcapkit rather than to the probe; and the devmode bootstrap **raises** if `PCAPKIT_DEVMODE` did not take effect, so a test cannot pass for the wrong reason.

Before, against `main`'s copies of both files: **25 failed, 31 passed**. After: **42 passed**. Suite **508 → 529 passed**, **315 → 332 subtests**. isort clean; pylint 10.00/10 on the eight files authored here.

`tests/utilities/test_logging.py` was deliberately **not** touched — it belongs to #384.

## Merge note for #384

#384 (logging) also touches both files, and there are two reconciliations at merge time: its `logging.rst` note saying these three issues are "deliberately left alone for now" becomes stale, and its `logger = get_logger(__name__)` conflicts textually — in `warnings.py` keep `get_logger(__name__)` and drop the now-unused `DEVMODE` import; in `exceptions.py` `BaseError.__init__` is rewritten around the logger calls here.

## Reported, not fixed — a pre-existing defect this makes visible

`stacklevel()` (`exceptions.py:55-73`) returns an **absolute** stack index rather than a depth relative to the caller, so `warnings.warn` receives values like `0` and every pcapkit warning is attributed to `pcapkit/utilities/warnings.py` rather than the user's call site (it also keys the dedup registry there). It was invisible while the channel was swallowed. Left deliberately: it is shared by both channels *and* by every `BaseError` log record, so fixing it would relocate the reported source of every warning and error — a fourth behavioural change nobody asked for. Wants its own issue.

Also corrected in the docs: `-W ignore::pcapkit.utilities.warnings.BaseWarning` **does not work** — CPython rejects it, because `-W` resolves the category before `site` adds `site-packages` to `sys.path`. Verified generic. The docs now recommend a standard base category on the CLI or a programmatic filter.

Closes #362, closes #363, closes #364.


JarryShaw:
@copilot Fix the code for all comments in [this review thread](https://github.com/JarryShaw/PyPCAPKit/pull/390#pullrequestreview-5205297417).

PR #397, opened 2026-09-15
utilities: make stacklevel() relative, as its callers require

JarryShaw:
One of the deferred findings from this session. The defect is real and turned out worse than the note described.

## The reported bug

`stacklevel()` returned an **absolute index** into `traceback.extract_stack()`, but every caller feeds it to something that counts frames *outward from the call site* — both `warnings.warn(stacklevel=)` and `logger.*(stacklevel=)` do. So the deeper the *outer*, non-pcapkit part of the stack, the larger the number and the further the attributed frame drifted.

One warning raised from six different depths, all from the same source line, wanting `probe.py:68` every time:

| outer frames | `warnings` said |
|---|---|
| 0 | `io.py:284` |
| 1 | `probe.py:68` ✓ |
| 3 | `chain.py:20` |
| 7 | `chain.py:16` |
| 15 | `chain.py:8` |
| 21 | `chain.py:2` |

Correct at exactly one depth, by coincidence. After the fix: `probe.py:68` at all six.

## Two worse consequences found while proving it

**Every pcapkit error was blamed on `logging` itself.** `BaseError.__init__` passed `stacklevel=-stacklevel()` — negated. `logging` treats a level ≤ 0 as "do not walk", so the log record pointed at `logging/__init__.py:1624` at *every* depth, unconditionally. The docstring bullet that motivated the negation described Python 2.7 behaviour.

**The -1 edge is reached in ordinary use, not as a corner case.** `traceback.extract_stack()` honours `sys.tracebacklimit`, which `BaseError.__init__` sets to `0` for loud errors outside devmode. After the first such error the stack comes back **empty** for the rest of the process, the `for`/`else` branch fires, and the function returns **-1** from then on:

```
after a loud error outside devmode has set sys.tracebacklimit = 0:
  BEFORE: raw stacklevel() = -1 | extract_stack() frames = 0 | warn says warnings.py:116
  AFTER:  raw stacklevel() =  1 | extract_stack() frames = 0 | warn says probe.py:68
```

This also explains a symptom that would otherwise look like flakiness: the new tests pass alone and fail in a full-directory run, because `test_decorators.py` raises a loud `StructError` first.

## The fix

The walk now uses `inspect.currentframe()`/`f_back` — immune to `sys.tracebacklimit`, and it skips building `FrameSummary` objects and doing `linecache` lookups in a function called once per warning.

Frames are numbered **outward from `stacklevel()`'s own frame**, so the number *is* the relative level a consumer wants:

```python
return max(1, min(boundary + 1, outermost))
```

Equivalent to `len(tb) - index` in the old outermost-first terms, but measured from the inside out, so it depends only on how deep the pcapkit frames run and never on the outer stack. Lower clamp because `logging` blames itself at ≤ 0; upper clamp for when the whole stack is inside pcapkit (running the package as a script), where walking past the top makes `warnings.warn` blame `sys`.

## The 47 call sites, fixed in one place

`warn(..., stacklevel=stacklevel())` appears 47 times across `corekit/`, `foundation/`, `protocols/`, `interface/`, `toolkit/`, `vendor/` and `__main__.py`. Each evaluates `stacklevel()` in its *own* frame, but `warnings.warn` is invoked one frame deeper inside the `warn` wrapper — so all 47 were one frame short even with correct arithmetic.

Fixed in `warn()` rather than at 47 call sites: its `stacklevel` argument is now documented as relative to *its caller*, the ordinary wrapper convention, and it adds its own frame back. Confirmed no site bypasses the wrapper with a direct `warnings.warn(..., stacklevel=stacklevel())`. The two direct `logger.*(..., stacklevel=stacklevel())` sites have no wrapper frame and were already correct.

## Verification

`tests/utilities/test_stacklevel.py` — 16 tests, 65 subtests, unit tier, no captures or optional deps. Three groups: synthetic-frame arithmetic (both clamps, an exhaustive "never below 1" over 25 stack shapes, interleaved stacks resolving to the outermost boundary); end-to-end attribution at depths 0/1/2/5/12/30 asserting the same *source line* each time, compared by text via `linecache` rather than by offset arithmetic; and the `warn()` wrapper convention. Plus an explicit regression for the `sys.tracebacklimit = 0` blinding.

Against the unfixed code: **68 failed, 7 passed**. Against the fix: 16 passed, 65 subtests.

| selection | result |
|---|---|
| `tests/utilities` | 86 passed, 100 subtests |
| fixture-free CI selection | **558 passed, 17 skipped**, 520 subtests |
| same on HEAD without the new module | 539 passed, 17 skipped, 455 subtests |

Delta is exactly the 16 new tests and 65 subtests — no regressions, nothing newly skipped. A lot of existing tests assert on warning and exception behaviour, so that is the number that matters here.

## Deliberately left alone

- The six literal `stacklevel=2` sites in `pcapkit/vendor/default.py` and `vendor/__main__.py`. Under the new wrapper convention `2` means "my caller", as the classic idiom reads, so their attribution moves out one frame — a fix rather than a regression, in network-bound crawler internals not exercised by the suite. Flagging it as the one deliberate behaviour shift.
- `BaseError`'s `sys.tracebacklimit = 0`, which `test_loud_error_still_limits_the_traceback` asserts as a feature. Worked around rather than changed.


JarryShaw:
Both looked at; one fixed, one declined with evidence.

**Fixed** — "Three groups of test here" → "tests". Straight grammar error, `fff7b0c35`.

**Declined** — "inside a ... path" → "an". `MARKER` is not a literal in that string, it is interpolated:

```python
MARKER = f'{os.path.sep}pcapkit{os.path.sep}'
```

so the message renders as:

```
this file is inside a '/pcapkit/' path, so the boundary
```

`'/pcapkit/'` is spoken with a consonant sound — "pee-cap-kit" — so **"a" is correct**; "an" would be wrong. The suggestion appears to have judged the `{MARKER!r}` placeholder rather than its value, which is a reasonable thing to get wrong from the source text alone.

`tests/utilities`: 86 passed, 100 subtests.

PR #402, opened 2026-09-15
interface: dispatch follow_tcp_stream on the engine, and fix the dpkt crash

### LINKED ISSUE #399: follow_tcp_stream ignores the engine: crashes on dpkt, silently returns no streams on scapy
`pcapkit.interface.misc.follow_tcp_stream` picks its reassembly adapter by comparing an object to a string, so the comparison is never true and the engine argument is effectively ignored.

## The bug

`pcapkit/interface/misc.py:93-99`:

```python
fallback = False
if extraction.engine == 'dpkt':      # type: ignore[comparison-overlap]
    from pcapkit.toolkit.dpkt import tcp_reassembly
elif extraction.engine == 'scapy':   # type: ignore[comparison-overlap]
    from pcapkit.toolkit.scapy import tcp_reassembly
else:
    from pcapkit.toolkit.pcap import tcp_reassembly
    fallback = True
```

But `Extractor.engine` is a property returning `self._exeng` — an `Engine` **instance**, not the engine's name:

```python
>>> ex = pcapkit.extract(fin='examples/captures/in.pcap', store=False, nofile=True, engine='dpkt')
>>> ex.engine
<pcapkit.foundation.engines.dpkt.DPKT object at 0x7f270826b4d0>
>>> ex.engine == 'dpkt'
False
>>> ex.engine.name
'DPKT'
```

So both branches are dead, `fallback` is **always** `True`, and `pcapkit.toolkit.pcap.tcp_reassembly` is always used — against frames produced by whichever engine actually ran. The two `# type: ignore[comparison-overlap]` comments are mypy having already reported this and being silenced.

## Impact — measured on `examples/captures/in.pcap`

```
engine=default  OK   streams=3
engine=dpkt     FAIL AttributeError: 'dict' object has no attribute 'packet'
engine=scapy    OK   streams=0
```

- **dpkt raises outright.** The pcap adapter is handed a dpkt frame and fails on the shape mismatch.
- **scapy is worse: it silently returns 0 streams** where the same capture yields 3. No error, no warning — just an empty result that looks like "this capture has no TCP streams".

The second is the dangerous one, since a caller has no way to tell a wrong answer from a real one.

`fallback` also changes the call: the fallback branch calls `tcp_reassembly(frame)` while the engine-specific branches would pass `count=index`, so even where the shapes happen to line up the frame numbering differs.

## Suggested fix

Compare something that is actually a string — the engine's own name, e.g. `extraction.engine.name` (`'DPKT'`, `'Scapy'`) or its class — and drop the `type: ignore` comments, which exist only to hide the defect. A `test_follow_tcp_stream` case per engine asserting the same stream count as `default` would pin it; note the counts genuinely differ for engines with capability gaps, so the assertion should be per-engine rather than uniform.

Worth checking whether the same always-false comparison pattern appears elsewhere — the `type: ignore[comparison-overlap]` marker is a good grep.

## Environment

`main` at `23c46071c`, Python 3.14, dpkt 1.9.8, scapy 2.7.0.

JarryShaw:
Closes #399 — with a correction to that issue's diagnosis, which was mine and was wrong about half of it.

## The dispatch bug

`follow_tcp_stream` chose its reassembly adapter with `extraction.engine == 'dpkt'`, but `Extractor.engine` returns the `Engine` **instance**, so both branches were dead code and `pcapkit.toolkit.pcap`'s adapter was always used — against frames from whichever engine actually ran.

The two `# type: ignore[comparison-overlap]` comments were mypy reporting precisely this and being silenced. Both are gone and mypy is clean on the file without them. Note it was only "clean" before because the always-false comparison made the `count=` call statically **unreachable**.

**Dispatch is now on the engine type**, chosen over `.name` deliberately: for a third-party engine registered as `class Foo(Engine, name='foo')` the registry key (`'foo'`) and the `.name` property (`'Foo'`) diverge, so a name comparison is unreliable — and a third-party engine that *subclasses* a built-in now routes to the right adapter. An engine matching none of the built-ins **warns and returns `()`** rather than silently using the wrong one, which is the failure mode this issue is about.

`fallback` became `pass_count`, set per adapter rather than per branch. The pcapkit-native adapters read the frame number off the dissected frame and take no `count`; the third-party ones have no intrinsic number and must be passed `count=index`. The old code had these backwards.

## The second defect, and my mistake in the issue

I filed #399 claiming the dpkt `AttributeError: 'dict' object has no attribute 'packet'` came from the reassembly adapter. It does not. Traced on `main`:

```
File "pcapkit/dumpkit/pcap.py", line 97, in __call__
    self._append_value(value, file, name or '')
File "pcapkit/dumpkit/pcap.py", line 139, in _append_value
    packet=value.packet,
AttributeError: 'dict' object has no attribute 'packet'
```

It fires during `Extractor` construction, in the PCAP **trace-flow dumper**: dpkt and scapy hand the tracer plain `dict` frames, and `follow_tcp_stream` lets `trace_format` default to `'pcap'`, which that dumper cannot re-serialise.

This is a **known, deliberately deferred** defect — `pcapkit/foundation/extraction.py:836-846` says so in a `NOTE`, and the guard on line 847 covers only `'pyshark'` and `'pypcapfile'`:

> The DPKT and Scapy engines report the frame as a mapping too and are deliberately *not* listed here: they are affected by the same defect on this revision, but they are outside the scope of this change.

Handled here inside `interface/misc.py` — upgrading to `'json'` for dpkt and scapy, silently when unset and with a `FormatWarning` when `'pcap'` was requested explicitly — rather than editing shared `extraction.py`, to respect that recorded deferral. **Still open afterwards:** direct `Extractor` use is untouched by this workaround, so extending the `extraction.py:847` guard remains worth doing separately.

## scapy returning 0 streams is not a bug

Also a correction to the issue. scapy does not recognise `in.pcap`'s link-layer type (`unknown LL type [1]`) and yields all six frames as opaque `Raw`, so no frame carries TCP and there is genuinely no stream to follow. The test asserts that **cause**, so a future scapy that learns the link type fails loudly instead of silently returning 0.

## Verification

| engine | before | after |
|---|---|---|
| `default` | 3 streams | 3 streams |
| `dpkt` | **crash** | **3 streams**, byte-identical to default incl. per-stream packet counts |
| `scapy` | 0 streams | 0 streams (cause now asserted) |

The old tests **mocked `engine` as a string** — which is exactly why they never caught this — and that mocking cannot survive type-based dispatch, so they are rewritten against real extraction of the committed `in.pcap`. Per-engine expected counts rather than a uniform number, since engines with genuine capability gaps legitimately differ. dpkt/scapy cases are gated on the optional extras so a fresh `[test]` clone skips rather than fails.

`tests/interface`: 12 passed. Fixture-free unit tier: **577 passed, 17 skipped** against a 572/17 baseline measured on the unmodified worktree — the delta is exactly the new tests.

## Also checked

`grep -rn "comparison-overlap"` finds two more, at `protocols/internet/hopopt.py:1004` and `ipv6_opts.py:1016`. Those are a **different, benign** pattern — `schema.seed is not NoValue`, where the field is typed `int` but defaults to the sentinel at runtime, so the guard is correct and the ignore hides an annotation gap rather than a dead comparison. Left untouched. No other object-vs-string comparisons exist in `pcapkit/`.


JarryShaw:
Copilot's finding is correct, and the gap is **pre-existing** rather than introduced — this PR merely made it observable, because the dpkt/scapy trace workaround has to compare `trace_format` against `'cap'`. Fixed in `ea800298d`.

`Formats` declared four values:

```python
Formats = Literal['pcap', 'json', 'tree', 'plist']
```

while both registries accept **eight**, checked against the live objects rather than by reading the source:

```
Extractor.__output__ : ['cap', 'json', 'pcap', 'plist', 'text', 'tree', 'txt', 'xml']
TraceFlow.__output__ : ['cap', 'json', 'pcap', 'plist', 'text', 'tree', 'txt', 'xml']
same set? True
```

So `'cap'` and the `'text'`/`'txt'`/`'xml'` aliases were accepted at runtime but unspellable for a type checker — a caller passing `format='cap'` got a spurious error against a value the library documents in its own registry. Both declarations (`foundation/extraction.py` and `interface/misc.py`) now match the registries exactly, and I asserted that programmatically against `__output__` rather than trusting the literal.

One deliberate call worth recording: two of those eight keys are **defective for output** — `'pcap'`/`'cap'` raise `TypeError: PCAPIO.__init__() missing 1 required keyword-only argument: 'protocol'`, and `'text'` raises `AttributeError: module 'dictdumper' has no attribute 'Text'`. I still included them, because they are *registered*, so they are part of the declared API surface; a type should describe what the API accepts, not which paths happen to work. Narrowing the type to hide broken formats would misreport the interface and silently change what a checker permits. Both defects are pre-existing and unfiled — worth their own issue.

`tests/interface` + `tests/foundation/test_extraction.py`: 19 passed, 2 subtests.

PR #407, opened 2026-09-16
gitignore: add .bench-venvs to ignore list

JarryShaw:
docs: update engine support table and test environment details

examples: enhance error handling for pcap-ct/libpcap engine availability

## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

- [ ] [Searched](https://github.com/JarryShaw/PyPCAPKit/search?q=is%3Apr&type=Issues) for similar pull requests
- [ ] Followed PEP8 coding style
- [ ] Tested with proper test samples

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [ ] New feature

---

### Description of your *pull request* and other information

&emsp; Explanation of your *pull request* in arbitrary form goes here. Please make sure the description explains the purpose and effect of your *pull request* and is worded well enough to be understood. Provide as much context and examples as possible.


JarryShaw:
Recommendations on re-doing the benchmark, since the current table can't be completed on any one machine.

## The blocker is not the hardware — it's that no single environment holds all seven engines

This is the constraint that shapes everything else, and it was measured rather than assumed:

- **`pypcap` and `pcap_ct` are mutually exclusive.** Both distributions own the import name `pcap`; install both and `pcap-ct` wins the import while upstream's extension module is shadowed and unreachable.
- **`pypcap` needs a compiler plus libpcap headers *and* the shared library**, and its 1.3.0 sdist won't build past Python 3.11 (pre-generated `pcap.c` from Cython 0.29.32).
- **`pyshark` needs the `tshark` binary** and can't run on 3.14 at all (`asyncio.get_event_loop_policy().get_event_loop()` raises).
- **`pypcapfile` 0.12.0 can't be imported on 3.12+** (`imp` removed).

So **3.11 is the last interpreter where every engine runs, and even there `pypcap` and `pcap_ct` cannot coexist.** Any complete table needs at least two environments. That's not a macOS problem — macOS just also fails the libpcap half.

## Recommendation 1 — publish ratios, not absolute ms/packet

This is the highest-value change and it makes the rest tractable.

Absolute ms/packet is only meaningful on the machine that produced it. Measured concretely: this Linux host disagrees with the README's existing M2 Pro figures **in both directions** — `dpkt` and `pcapkit` about 2× slower, `scapy` about 1.5× *faster*. So the two tables can't be mixed row-for-row no matter how carefully either was taken, and patching in new rows beside old ones is misleading by construction.

Normalise every engine to the **`default` engine measured in the same run**:

```
Engine        relative to ``default``
============  =======================
``dpkt``                        0.06x
``scapy``                       0.13x
``pcapkit``                     1.00x   (baseline)
``pcap_ct``                     0.22x
``pyshark``                       67x
```

Ratios survive hardware, OS and interpreter differences far better than absolute times, and they answer the question a reader actually has: *how much faster is engine X than the built-in parser?*

Keep one absolute number — the `default` engine's ms/packet — as a footnote with its environment, so the ratios can be turned back into wall-clock if someone wants.

## Recommendation 2 — Docker, for reproducibility and to sidestep disclosure

A pinned image solves three problems at once:

- **Reproducibility.** Publish the `Dockerfile` (or its digest) and anyone can re-run and get the same *ratios*. Right now the numbers are unfalsifiable.
- **Availability.** You control whether `libpcap-dev` and `tshark` are present, which is exactly what neither macOS nor a bare runner gives you.
- **Disclosure.** You report the image and a CPU class, not the host. That is the cleaner answer to "run it on this Linux box but don't say which" — don't obscure the host, make the host irrelevant.

Two images (or one image, two venvs — which is what the `.bench-venvs/` in this PR already gestures at):

| environment | Python | engines |
|---|---|---|
| A | 3.11 | `default`, `dpkt`, `scapy`, `pyshark`, `pypcapfile`, **`pypcap`** |
| B | 3.11 | `default`, `dpkt`, `scapy`, `pyshark`, `pypcapfile`, **`pcap_ct`** |

`default` appears in both, so it stitches them together: express everything as a ratio to `default` and the two runs become comparable even though they are different environments. Label which environment produced each row.

**What Docker does not fix:** container timing on a loaded host is still noisy. It buys reproducibility, not precision — which is fine, because ratios tolerate noise far better than absolute times.

## Recommendation 3 — publish the noise floor

Whatever the method, measure repeatability and state it. On back-to-back runs here the spread was **1.4–3.3%**. Without that figure a reader will over-read a 5% gap; with it, they know which differences are real. The `pypcap` vs `pcap_ct` gap (~18%, the Cython binding beating the ctypes one) is comfortably signal; anything under ~5% is not.

Also pin the capture and the iteration count in the table caption, not just in the harness — the existing methodology (1,000 timed extractions of `in.pcap`, `perf_counter_ns`, first round discarded) is good and worth stating where the numbers are read.

## On GitHub Actions

Worth considering as the *publishable* option: free, public, and anyone can re-run the workflow. The vCPUs are shared and noisy, so absolute numbers would be poor — but for a ratio table with a stated noise floor it is defensible, and it has the significant advantage that the numbers stop depending on any one person's laptop. If you go this way, run it on a schedule rather than per-PR so it doesn't gate merges.

## What I'd do concretely

1. Re-measure the **whole** table in the new scheme rather than patching rows; the existing macOS figures can't be mixed with anything.
2. Two venvs on one Linux image, Python 3.11, `default` in both as the stitch.
3. Publish ratios plus one absolute baseline, the repeatability spread, the capture, the iteration count, and the image reference.
4. Keep the engine × Python **support** matrix separate from the speed table — it is host-independent and the more useful of the two for most readers.

Happy to implement whichever shape you prefer.

PR #408, opened 2026-09-16
docs: prune the Help Wanted page, fix the uninstallable docs extra, quiet 5 Sphinx warnings

JarryShaw:
Three things: the `[docs]` extra could not install at all, `pep.rst` was asking for help with work already finished, and a handful of Sphinx warnings were ours rather than pre-existing.

## The `[docs]` extra was uninstallable

`pip install -e '.[docs]'` failed outright, so **nobody could build the documentation from a clean checkout**:

- `sphinx-opengraph` **404s on PyPI**. The real package is `sphinxext-opengraph`, which is what `docs/source/conf.py:61` imports.
- `sphinxcontrib-mermaid` was missing from the extra entirely, though `conf.py:65` loads it.

Both fixed; the extra now resolves (verified with `pip install --dry-run`, exit 0).

## `pep.rst` was soliciting help for finished work

Which is worse than saying nothing — a contributor reads it and starts on something already shipped. Every claim was checked against the code before editing.

**Solicitations removed**, each verified present: SCTP (all 13 RFC 9260 chunk types, 8 parameters, 13 error causes, real CRC32c, PPID dispatch), both new engines, the logging integration, the test suite.

**Narrowed rather than deleted**, which is more useful than a stale "help wanted":

| Item | Now says |
|---|---|
| ESP | the four missing algorithms, confirmed absent; 5 encryption and 5 integrity algorithms actually applied |
| Mobility Header | FMIPv6 **has** landed, so the vague "requires some help" became: 10 of 24 message data types, 51 of 71 options, 3 of 4 CGA extensions |
| SCTP | narrowed to the registered-but-unimplemented surplus: 17 of 30 chunk types, 24 of 32 parameters, 10 of 23 error causes |
| `NotImplemented` list | kept, with NDP's stub corrected to `link/` (not `internet/`) and QUIC dropped — no stub exists |

**Three places the page was simply wrong:**

1. It claimed `BaseWarning` still calls `warnings.simplefilter`. That was removed in `487d3da82`, with tests. Deleted; the `warn()` double-reporting item is still true and stays.
2. "84 test modules" — the real count is 86.
3. "bundled with the distribution" — `[tool.setuptools.packages.find]` excludes `test*`, so an installed PyPCAPKit has **no** `tests` package. Now recorded as a still-wanted item rather than a boast.

The page also claimed to mirror discussion 106 while carrying 4 of its 6 comments; the missing two are added.

**The `pcap-ct` paragraph said "a way out that has not been adopted yet".** It has been adopted — as its own `PCAP_CT` engine in #405. It now describes what shipped, why a separate engine rather than a second backend (both distributions own the import name `pcap`; with both installed pcap-ct wins and upstream is shadowed), and the two caveats that remain: both are betas, and a runtime `libpcap.so.1` is still required because the `libpcap` wheel's config sets `LIBPCAP = None`.

## Sphinx warnings: 268 → 263

**The TOC suspicion did not pan out, and that is worth recording** so nobody re-investigates it. Across 128 pages: **0 orphans, 0 broken toctree entries, 0 `automodule` targets pointing at nothing, 0 unresolvable `:doc:` targets, 0 pages with multiple toctree parents.** All 19 newly added modules resolve to a reachable page, and the const-enumeration convention is correctly followed — `esp.rst` and `sctp.rst` live under `docs/source/pcapkit/const/`, and the ESP protocol page only cross-references them rather than `automodule`-ing them.

What was actually fixed:

- `sphinx.ext.autodoc.typehints` removed from `extensions` — an internal submodule with no `setup()`, so it only ever produced a warning. The real work is done by the third-party `sphinx_autodoc_typehints`.
- The duplicate-object warnings on `esp.rst` and `context.rst`, the only two pages in the protocols/corekit tree using `automodule`. The cause needed two attempts: `:no-index:` alone removed only half of them, because the **module docstring itself** carries `.. module::` — as 362 pcapkit modules do. Dropping the redundant `.. module::` from the `.rst` and keeping `:no-index:` leaves the docstring as the single declaration. Both files carry a comment explaining why they deviate.

**The 263 remaining are pre-existing and none are in the new modules**, itemised so the next person doesn't have to re-count:

| count | cause |
|---|---|
| 158 | ambiguous cross-references — short names like ``:class:`TransType``` matching both `const.*` and `vendor.*` |
| ~40 | unresolvable forward references — `TYPE_CHECKING`-only names in annotations |
| 38 | duplicate object descriptions on established pages (`ipv6_opts`, `hopopt`, `tcp`, `mh`, …), **0** involving esp/sctp/context |
| 22 | guarded-import failures — environment-only, absent in CI with `.[all]` |
| 5 | autodoc signature crashes on `__protocol_type__`/`__schema__` |

Left alone deliberately: those pages render correctly today, and fixing the duplicates means judgement calls about rendered output on pages this PR has no other reason to touch.

## Not done here

The 158 ambiguous cross-references and ~40 forward references need docstring changes across `pcapkit/`, which is a separate sweep. Same for the 5 autodoc crashes.


docs/source/pcapkit/corekit/context.rst
```diff
@@ -1,10 +1,15 @@
 Parsing Context
 ===============
 
-.. module:: pcapkit.corekit.context
+.. Unlike its sibling pages, this one renders the module docstring through
```
JarryShaw: Prefer the format of its sibling pages

docs/source/pcapkit/protocols/internet/esp.rst
```diff
@@ -1,10 +1,15 @@
 ESP - Encapsulating Security Payload
 ====================================
 
-.. module:: pcapkit.protocols.internet.esp
+.. Unlike its sibling pages, this one renders the module docstring through
```
JarryShaw: Same, prefer the format of its sibling pages

JarryShaw:
Fixed in `6a62710be` — the review was right, and the contradiction was sharper than it looked.

The intro said proposals are recorded *"here in the discussion thread"* and then *"documented and maintained here"*, conflating the page with the thread. And because this branch added an `.. important::` block that carefully distinguishes the two, that paragraph ended up **contradicting the note three lines above it**.

Reworded so "the discussion thread" and "this page" are always distinct: the proposals were *raised* in the thread; the page is where they are *kept up to date*. Two things fixed while there:

- The tense — 16k lines and 800 commits are long past, so "As PyPCAPKit reaches" became "reached".
- "please leave a note either in this thread" had no referent on a rendered page, since the page is not the thread. It now links the discussion and the Q&A category explicitly.

`rstcheck` clean on the file.

PR #420, opened 2026-09-16
perf: cut extraction time by ~46% on an HTTP capture, with byte-identical output

JarryShaw:
A profiling pass before 1.5.0, per the ask to "see if there's anything to improve/accelerate the library". Four optimisations, one reverted, and a 643-line `PROFILING-NOTES.md` carrying the measurements — including the suspects that turned out **not** to be worth touching, which is the half that stops the next person re-treading it.

## Measured, and independently re-measured

`http.pcap` (1117 frames), best-of-7, `store=False, nofile=True`, separate processes:

| capture | before | after | |
|---|---|---|---|
| **`http.pcap`** | **1042.7 ms** | **560.7 ms** | **−46.2%** |
| `test.pcap` | 28.3 ms | 17.0 ms | −39.8% |
| `many_interfaces.pcapng` | 49.6 ms | 39.2 ms | −20.9% |
| `profile.pcapng` | 43.9 ms | 35.2 ms | −19.7% |
| `ipv6.pcap` | 5.06 ms | 4.37 ms | −13.6% |
| `ipv4.pcap` | 1.90 ms | 1.66 ms | −12.6% |

The headline row is my own A/B against `origin/main` on the same interpreter, which came out slightly better than the −44.8% originally reported.

## Correctness first

**All 14 committed captures, serialised to both `tree` and `json`, with reassembly and flow tracing on — 30 files, byte-for-byte identical to `main`.** `diff -rq` clean. Full suite: 782 passed, 17 skipped, zero failures. `pylint` and `mypy` message sets identical to pristine.

Two things in the diff could have been subtly wrong, and both were checked rather than inferred from the passing tests:

- `FieldBase.__copy__` replaces the generic `copy.copy` path with `__new__` plus a `__dict__` update. That is only equivalent if nothing in `corekit/fields/` defines `__slots__`, `__getstate__`, `__setstate__`, `__reduce__` or `__deepcopy__` — grep confirms none does.
- Hoisting `length = field.length` above `field.unpack(...)` in `Schema.unpack` would skew the later `consumed = length - field.option_padding` arithmetic **if** `unpack` could change `_length`. It cannot: `_length` is assigned only in the constructor and in `__call__` (which runs before the hoist), while `unpack` touches `_option_padding` only.

## The four wins

- **`fa0920f97` — charset detection was uncached, and was 30% of an HTTP extraction.** 3011 `chardet.detect()` calls over **163 distinct bytestrings**, 94.6% of them redundant: an HTTP-heavy capture asks for the encoding of `b'Connection'` once per message and gets the same answer every time. `chardet.detect` is a pure function of its bytes, so the verdict is now memoised on them, bounded at 1024 entries so a capture full of never-repeating text cannot retain all of it. Wired into both call sites — attribution mattered, since `http.pcap` only reaches `ProtocolBase.decode` and `many_interfaces.pcapng` only `StringField.post_process`. **−30.3%.**
- **`5ce03cd77` — `FieldBase.__copy__`, which helps every capture shape.** Every field of every protocol is copied once per packet, 63207 times per extraction, and it was falling through `__reduce_ex__`/`_reconstruct`. Microbenchmark 1.66 µs → 0.43 µs. **−11% to −16% everywhere.**
- **`0c1387460` — `Field.length` read once instead of two to four times.** It recomputes `struct.calcsize()` on every read: 176908 calls per extraction. **−2%.**
- **`0aca00690` — `Schema.__setattr__` was re-entering itself once per field assigned.** 180297 of 254043 calls were that round trip, each one missing the `__fields__` test and falling through to `object.__setattr__`. **−3.7%.**

## One optimisation reverted

`faef5d649` gives back a measured −1.0%: hoisting an `isinstance` into a local cost mypy its type narrowing and added four `attr-defined` errors. Not a trade worth making in this codebase.

## Three larger wins found and deliberately not taken

All outside the parse hot path, all wanting their own review:

1. **`dumpkit/pcap.py:94,128` — the flow dumper reopens the output file per frame and rebuilds a whole `Frame` to get bytes it already holds.** 80% of the flow-tracing cost (+849 ms → +168 ms in a counterfactual), with **330 of 331 output files byte-identical**. Best-evidenced and lowest-risk change available; it is not here only because it is a separate concern from the field machinery.
2. **`reassembly/ip.py:189` — `analyze()` runs eagerly on every frame, fragmented or not**, because `toolkit/pcap.py:53` filters only on DF. `http.pcap` contains zero actual fragments and still yields 1117 "datagrams". 86% of the IP-reassembly cost plus a 133 ms GC bill. Making `packet` lazy is mechanical; **whether unfragmented frames should be emitted at all is a design call for you**, not something to decide inside a perf pass.
3. **`collections.py:262-270` parses every option twice** — 2274 schema unpacks for 1137 options, the pre-parse always discarded. ~15% of a PCAP-NG extraction, higher risk than the rest.

## Two non-performance defects found while profiling

Neither is fixed here; both will be filed:

- `schema_final`'s generated typed `__init__` is **dead code** — the `schema.py:77` guard is always False because `Schema.__init__` is `__update__`, so `__post_init__` never runs.
- `TCP(**kwargs)` **reliably raises** `AttributeError` at `tcp.py:479`.

## Dismissed with data, so nobody repeats it

`logging.py` (zero logger calls during an extraction; 0.03% in the heaviest shape — all lazy `%s`, the one f-string is in `vendor/__main__.py`), `multidict.py` (<1%), `ProtoChain` (0.3%), `_import_next_layer` (0.035 s tottime; its large cumtime is just recursion), **`Info`/`InfoMeta` (0.0% of construction, 4.4% of parse)** — the most-suspected component turned out not to be the problem — `store=True` (1.3%), and `in.pcap` as a profiling target at all (6 frames, setup dominates).

Two measurement traps worth knowing, both recorded in the notes: `reassembly=True` and `trace=True` are **no-ops without** `ip=`/`tcp=`, so a benchmark passing only the switch measures nothing; and timing several capture shapes in one process inflates them by up to 73%.


pcapkit/corekit/fields/strings.py
```diff
@@ -17,6 +18,33 @@
...
+
+    Args:
+        value: Bytestring whose encoding is to be detected.
+
+    Returns:
+        Name of the detected encoding, or ``'utf-8'`` where detection declines
+        to name one.
+
+    """
+    return chardet.detect(value)['encoding'] or 'utf-8'
```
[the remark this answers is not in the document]
JarryShaw: Right, and fixed in 572f13b. `lru_cache` bounds entry *count*, not entry *size*, and `ProtocolBase.decode` being public means a caller can hand it a whole payload — so 1024 of those would be retained for the life of the process.

Rather than guess a threshold I measured what actually reaches detection, across three captures:

```
http.pcap               calls=163  distinct=39  max=116  p95=52
http6.cap               calls= 11  distinct= 7  max=101  p95=101
many_interfaces.pcapng  calls=  7  distinct= 6  max= 27  p95=27
```

So everything a real capture presents is well under 256 octets, and 256 is now the cap. That keeps every repeating string the optimisation was aimed at while bounding worst-case retention to 256 KiB instead of unbounded. A value large enough to bypass is unlikely to recur, so it loses nothing it was not already paying.

Verified: short values still cached (1 miss then 2 hits), long values leave `currsize` untouched, and both paths return exactly what `chardet.detect` returns. `http.pcap` best-of-7 is 565.5 ms against 560.7 ms before, so the win is intact. Full suite 782 passed, 17 skipped.

PROFILING-NOTES.md
```diff
@@ -0,0 +1,643 @@
...
+
+The editable install in `.venv` points at the **main** checkout, so `PYTHONPATH`
+must name the tree under test or you profile the wrong code:
+
+```bash
+PY=/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python
+WT=/local/home/jarryx/GitHub/PyPCAPKit/.claude/worktrees/agent-a14f0c4dddc25f516
+
+$PY examples/generators/make_samples.py          # build the un-committed fixtures
+PYTHONPATH=$WT $PY -m pytest -q                  # 782 passed, 17 skipped
```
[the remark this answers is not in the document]
JarryShaw: Fair, and fixed in 572f13b. Those two lines were the only absolute paths in the file, and they named this machine's worktree, which is useless to anyone else.

The snippet now derives the tree it is testing from `git rev-parse --show-toplevel` and leaves the interpreter as a plain `python`, so it is copy-pasteable from any checkout. I also added a line saying the figures were taken on CPython 3.14 on Linux and that only the ratios are meant to carry across machines — the same reasoning that stopped the absolute benchmark numbers being pasted into `README.rst` from this host.

pcapkit/corekit/fields/strings.py
```diff
@@ -17,6 +18,64 @@
...
+
+    Args:
+        value: Bytestring whose encoding is to be detected.
+
+    Returns:
+        Name of the detected encoding, or ``'utf-8'`` where detection declines
+        to name one.
+
+    """
+    if len(value) > DETECT_CACHE_MAX_BYTES:
```
JarryShaw: what about directly cache on the first MAX_BYTES bytes?

JarryShaw: Tried that first, and it is not safe — `chardet` is statistical over the whole byte sequence, so a prefix can disagree with the full value. Measured:

```
case                                   full    first 256  agree?
ascii prefix, utf-8 body              utf-8        ascii  NO
ascii prefix, latin-1 body     Windows-1252        ascii  NO
ascii prefix, cp1251 body      Windows-1251        ascii  NO
utf-8 prefix, ascii body              utf-8        utf-8  yes
utf-16 whole                         UTF-16       UTF-16  yes
long pure ascii                       ascii        ascii  yes
```

Three of six disagree, and they disagree the wrong way round: anything with an ASCII header and a non-ASCII body — an HTTP response with `Content-Type` lines followed by UTF-8 text is the obvious one — would be detected as `ascii` and decoded wrongly. That is a correctness regression bought with a memory saving, so it is the one trade not worth making here.

Keying on the prefix while storing the *full* verdict has the same problem from the other direction: two payloads sharing their first 256 bytes but differing later would collide on one cache entry and one of them would get the other's answer.

**There is a correct version of your idea**, if you want long values cached: key on `hashlib.blake2b(value).digest()` instead of on the bytes. That is 64 octets regardless of input size, so memory stays bounded *and* the answer stays exact, and the hash is O(n) against a `chardet.detect` that is also O(n) but far more expensive — so it would be nearly free.

I left it out because it buys nothing measurable. Across `http.pcap`, `http6.cap` and `many_interfaces.pcapng` the largest value reaching detection was **116 octets**, p95 **52** — nothing came close to the threshold, so no real capture has a long value to cache, let alone a recurring one. Adding a hash to the hot path for zero measured gain seemed the wrong side of the trade, but say the word and it is a small change.

PROFILING-NOTES.md
```diff
@@ -0,0 +1,647 @@
+# PyPCAPKit profiling notes (pre-1.5.0)
```
JarryShaw: prefer .rst and this shouldn't be kept on top level or maybe not even in the commit.

pcapkit/corekit/fields/strings.py
```diff
@@ -17,6 +18,64 @@
...
+
+    Returns:
+        Name of the detected encoding, or ``'utf-8'`` where detection declines
+        to name one.
+
+    """
+    return chardet.detect(value)['encoding'] or 'utf-8'
+
+
+def _detect_charset(value: 'bytes') -> 'str':
```
JarryShaw: make promote to be a utility function? since protocol.py uses it as well.

PROFILING-NOTES.md
JarryShaw: Dropped it entirely in 427a89d4, which I think is the right end of your "or maybe not even in the commit".

Converting to `.rst` at top level would have made it *worse* rather than better: `MANIFEST.in` carries `global-include *.rst`, so a top-level `.rst` newly ships in every sdist — developer notes delivered to every user. Moving it under `docs/` avoids that (`docs/` is pruned from the sdist) but then it is a Sphinx source file, so it either needs a toctree entry or it becomes an orphan warning.

And it turned out to be largely duplicative: `docs/source/pep.rst`'s performance section, merged in #419, already carries the findings, the four wins, the three remaining ones, the two predictions the profiling contradicted, and the measurement traps — in `.rst`, user-facing, properly placed. The exact call counts and per-capture figures live in this PR's description permanently.

pcapkit/corekit/fields/strings.py
JarryShaw: Agreed, and done in 427a89d4 — it is now `pcapkit.utilities.chardet`, exporting `detect_charset`, with a docs page in the utilities toctree. You are right that a private name imported across modules was the wrong shape for something two callers share; I had flagged the same smell reviewing the original commit and should have acted on it then.

Named `chardet` per your convention note, matching `utilities/logging.py` and `utilities/warnings.py`, which are named for the stdlib modules they extend. Worth recording that this does **not** shadow the real package: absolute imports mean `import chardet` inside `pcapkit/utilities/chardet.py` still resolves the third-party one, verified at 7.6.0. The now-unused `import chardet` in `strings.py` went with it.

Full suite 782 passed, 17 skipped.

pcapkit/utilities/chardet.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Returns:
+        Name of the detected encoding, or ``'utf-8'`` where detection declines
+        to name one.
+
+    """
+    return chardet.detect(value)['encoding'] or 'utf-8'
+
+
+def detect_charset(value: 'bytes') -> 'str':
```
JarryShaw: maybe let's name is `detect` as well?

JarryShaw: Done in 01a39eb9 — it is `pcapkit.utilities.chardet.detect` now, which reads the way `utilities/warnings.py` exports `warn` for `warnings.warn`. The inner memoised helper follows as `_detect_cached`.

One thing I got wrong and then undid, worth recording. I talked myself into believing a module named `chardet` would make the `:func:`chardet.detect`` references in its own docstrings ambiguous — Sphinx's Python domain resolves an unqualified target by suffix match, and `pcapkit.utilities.chardet.detect` does end with `chardet.detect`, so the reasoning looked sound. I replaced them with plain literals and wrote a Note explaining the hazard.

The rendered HTML disagrees:

```
chardet.detect()  ->  https://chardet.readthedocs.io/en/latest/api/index.html#chardet.detect
detect()          ->  #pcapkit.utilities.chardet.detect
```

Intersphinx resolves all three upstream references correctly, and only the bare `detect` binds locally. So the roles are restored, the Note describing the non-problem is gone, and the module keeps working links. Inference lost to measurement — the docs build was the thing to check before editing, not after.

`import chardet` inside the module still resolves the third-party package (absolute imports; verified 7.6.0). Full suite 830 passed, 17 skipped.

pcapkit/corekit/fields/strings.py
JarryShaw: let's do the hash still. that's a right thing to do. sample files do not produce real world statistics.

JarryShaw: Done in 9f1b56b1 — and you were right about *why*, which is the part that matters. My case for the threshold rested on "the largest value reaching detection was 116 octets across three sample captures", and sample captures are not a sample of real traffic. A capture full of large repeated text is exactly the case that most wants the cache, and the threshold excluded precisely that case.

Keyed on a 32-octet `blake2b` digest now, so it is bounded by count *and* size while every value benefits regardless of length:

```
1524 distinct 5 KB values  ->  1024 entries, ~32 KiB retained (not 5 MiB)
a 2 MB value               ->  cached under a 32-byte key
```

Verified identical to `chardet.detect` on ASCII-prefixed UTF-8, Latin-1 and UTF-16 inputs. It is an `OrderedDict` with `move_to_end`/`popitem` rather than `lru_cache`, since the key is a digest of the argument rather than the argument itself and `lru_cache` cannot express that.

**The cost, stated plainly:** `http.pcap` best-of-7 goes 565.5 ms → **588.0 ms**, about 4%, from hashing the 3011 short strings the previous version keyed directly. Against the 1042.7 ms pre-optimisation baseline that is still −43.6%.

If you want that 4% back, a hybrid — raw bytes below 256 octets, digest above — is correct and still bounded, at the price of two key paths in one function. I have left it out on the grounds that one path is easier to reason about than 4%, but it is a five-line change if you disagree.

Full suite 830 passed, 17 skipped.

pcapkit/utilities/chardet.py
```diff
@@ -0,0 +1,91 @@
...
+#: 256 bits, so a collision -- which would hand one bytestring another's verdict
+#: -- is not a thing that happens; halving :func:`~hashlib.blake2b`'s default 64
+#: halves what the cache retains per entry for no practical loss.
+DETECT_DIGEST_SIZE = 32
+
+#: Detected encodings, keyed by digest of the bytes they were detected from, least
+#: recently used first. An :class:`~collections.OrderedDict` rather than
+#: :func:`functools.lru_cache` because the cache key is a digest of the argument
+#: rather than the argument itself, which ``lru_cache`` cannot express.
+_cache = collections.OrderedDict()  # type: OrderedDict[bytes, str]
```
JarryShaw: Since lru_cache actually uses hash value to key, maybe we don't have to rebuild the wheel then?

JarryShaw: It caches by hash *and* keeps the key object, which is the part that bites — a dict has to retain the key to settle equality on a hash collision. Measured:

```
fed 20 distinct 1000 KB values = 19 MB total
  lru_cache keyed on the bytes  ->    19.1 MB retained
  digest-keyed OrderedDict      ->     0.0 MB retained
```

So `lru_cache` holds every byte it was keyed on, for the lifetime of the entry. `functools._make_key` confirms why:

```python
>>> k = functools._make_key((b'A' * 10,), {}, False)
    type=tuple  holds the value=True
```

`_make_key` short-circuits to the bare argument only for `int` and `str` (its `fasttypes`); anything else — `bytes` included — is wrapped in a tuple that holds the argument. The cached hash saves *recomputing* the hash, not *storing* the value.

Two notes on how I got here, since the first answer I gave you was wrong. My initial check counted surviving objects with `gc.get_objects()` and reported 0 MB retained for both, which would have supported your reading. That was a broken measurement: plain `bytes` contain no references, so they are not GC-tracked and never appear in `gc.get_objects()` at all. `tracemalloc` measures the allocation directly and gives the numbers above.

And I am not rebuilding `lru_cache` so much as sidestepping the one thing it cannot express: the key here is a *digest of* the argument rather than the argument, and `lru_cache` has no hook for that — it keys on what you pass it. The `OrderedDict` is doing the LRU eviction `lru_cache` would otherwise do, in six lines, which is the cost of that.

If the six lines are the objection rather than the memory, the alternative is `lru_cache` on an inner function taking the digest, with the outer function passing the bytes through a module-global — which is worse in every way I can see.

JarryShaw: Done in bcc9fea8 — plain `lru_cache` over `chardet.detect`, and it is the fastest of the three variants as well as the simplest:

```
lru_cache on the value   559.8 ms
size threshold           565.5 ms
digest-keyed             588.0 ms
baseline (no cache)     1042.7 ms
```

The module is ~70 lines shorter, and `detect.cache_clear()` comes free — which is a genuine gain over the hand-rolled `OrderedDict`, since a long-running consumer now has a way to release the cache.

The trade-off is written into the docstring rather than left in this thread, with the numbers behind it, so nobody re-runs the experiment: the bound is on entry *count* rather than footprint, because `lru_cache` keeps the argument alongside its cached hash (20 × 1 MB retains 19.1 MB, where the digest version retained 0.0 MB). Both rejected alternatives are named there too — prefix keys unsound, digest keys sound but inexpressible with `lru_cache`.

PR #424, opened 2026-09-16
reassembly: make the four IPv6 adapters agree, and analyse datagram payloads lazily

### LINKED ISSUE #415: IPv6 reassembly: the default/pcapng adapters keep the Fragment header in ihl and header, dpkt/scapy do not
The four toolkit adapters disagree about whether the 8-octet Fragment header
belongs to the IPv6 reassembly packet's `ihl` and `header`. Measured on
`examples/captures/ipv6.pcap`, frame 13:

| adapter | `ihl` | `len(header)` | `tl` | Fragment header inside `header`? |
|---|---|---|---|---|
| `toolkit/pcap.py` (default) | 48 | 48 | 1496 | **yes** |
| `toolkit/pcapng.py` | 48 | 48 | 1496 | **yes** |
| `toolkit/dpkt.py` | 40 | 40 | 1488 | no |
| `toolkit/scapy.py` | 40 | 40 | 1496 | no |

Note `tl` is a three-way disagreement: `scapy` and the default agree at 1496
while `dpkt` reports 1488.

## Why the default path includes it

`pcapkit/protocols/internet/ipv6.py:341` adds each extension header's length to
`hdr_len` *before* the Fragment-header check breaks the loop at line 354:

```python
hdr_len += next_.length          # line 341
...
if ex_proto == Enum_ExtensionHeader.IPv6_Frag:   # line 354
    ipv6.__update__({'fragment': self._read_packet(header=hdr_len, payload=raw_len)})
    break
```

So by the time `_read_packet(header=hdr_len, ...)` cuts the header, `hdr_len`
already counts the Fragment header — and `ipv6_info.fragment.header`, which
`toolkit/pcap.py:119` passes as `header`, therefore contains it. The data
model's own comment agrees that this is `hdr_len`'s meaning: "Header length
(including extension headers)".

`dpkt` and `scapy` compute the boundary independently and land on the other
side: `hdr_len + ipv6_frag.__hdr_len__` is where `dpkt` starts the payload, and
`scapy` uses `bytes(ipv6)[:-len(ipv6_frag)]`.

## The user-visible consequence

The reassembled datagram presents a Fragment header on a datagram that is by
definition no longer a fragment:

```
reassembled datagram, default engine:
  len(header)=48  next-header byte=44 (44 = IPv6-Frag)
  payload len=4778
```

:rfc:`8200#section-4.5` — "The Fragment header is not present in the
reassembled packet." So the default and pcapng adapters are the wrong ones
here, and `dpkt`/`scapy` are right.

## Suggested fix

Have `toolkit/pcap.py` and `toolkit/pcapng.py` report the header up to but not
including the Fragment header, matching the other two adapters and the RFC.
`ipv6_info.hdr_len` itself should keep its documented meaning (it is correct as
a header length; it is just not the right value for this field), so the
adapters need the pre-Fragment length rather than a change to the data model.
`tl` wants settling in the same pass, since no two adapters currently agree on
it either.

Worth a test asserting all four adapters produce the same `ihl`, `header` and
`tl` for the same capture — the inconsistency survived because each adapter is
only ever tested against itself.

Found while fixing the documentation for these fields in #413, where the
docstring's "only headers before IPv6-Frag" comment turned out to be wrong for
the adapter it describes. That comment is repeated verbatim in all four
adapters and should be corrected alongside the behaviour.

JarryShaw:
Closes #415. Also lands the second of the three performance findings from #420, which lives in the same files.

## The correctness fix — and a data-corruption defect the issue did not name

#415 reported that the four toolkit adapters disagree about whether the 8-octet IPv6 Fragment header belongs to `ihl`, `header` and `tl`. They now agree. Measured on `examples/captures/ipv6.pcap` frame 13, and independently re-measured:

| adapter | `ihl` | `len(header)` | `tl` | → | `ihl` | `len(header)` | `tl` |
|---|---|---|---|---|---|---|---|
| `toolkit/pcap.py` | 48 | 48 | 1496 | → | **40** | **40** | **1488** |
| `toolkit/pcapng.py` | 48 | 48 | 1496 | → | **40** | **40** | **1488** |
| `toolkit/dpkt.py` | 40 | 40 | 1488 | → | 40 | 40 | 1488 |
| `toolkit/scapy.py` | 40 | 40 | **1496** | → | 40 | 40 | **1488** |

**That `scapy` row was silently corrupting reassembled payloads.** Its `ihl` was right, but `tl` was `len(ipv6)`, which counts the Fragment header — and the machinery writes each fragment over the span `tl - ihl` (`reassembly/ip.py:99-100`), so the span was 8 octets too wide. On `main`:

```
default  payload=4778
dpkt     payload=4778
scapy    payload=4786      <-- eight stray zeroes per fragment
```

A reassembled datagram that no other engine agreed with, and nothing reported it. `tl` is now derived from the payload actually handed over, so `tl - ihl == len(payload)` holds structurally rather than by coincidence.

`ipv6_info.hdr_len` keeps its documented meaning ("including extension headers") and `protocols/internet/ipv6.py` is untouched — `pcap`/`pcapng` subtract `ipv6_frag.length`, undoing exactly the addition `ipv6.py:342` makes before the loop breaks.

## A second defect #415 conflated with the first

Excluding the Fragment header's *octets* still left the field *pointing* at it, which is why `header[6] == 44` on all four engines, `dpkt` and `scapy` included. :rfc:`8200#section-4.5` also requires the Fragment header's Next Header value to move into the last header of the unfragmentable part.

That is a **reassembly** rule rather than a per-fragment fact — an individual fragment's header genuinely does say 44 — so it went into `IP._rectify_header`, an identity hook that `ipv6.py` overrides, applied once for all four engines. It walks the extension-header chain rather than assuming offset 6, since Hop-by-Hop, Routing or Destination Options headers may precede the Fragment header, and it handles AH's different length encoding.

Verified on every engine:

```
default  len(header)=40  header[6]=17 (UDP)  payload=4778
dpkt     len(header)=40  header[6]=17 (UDP)  payload=4778
scapy    len(header)=40  header[6]=17 (UDP)  payload=4778
```

## The performance fix — `analyze()` was running on every frame

`Datagram.packet` is now analysed on first read. `http.pcap`, best of 5, one interpreter per shape:

| shape | before | after | |
|---|---|---|---|
| `reassembly=True, ip=True` | 1881.1 ms | **1111.2 ms** | −40.9% |
| the feature's own cost | +834.4 ms | **+77.3 ms** | **−90.7%** |
| GC bill | +79.8 ms | ~0 | |

`test.pcap`: 40.7 → 29.7 ms, feature cost +12.8 → +1.3 ms. Forcing every `packet` gives 1793.8 ms, confirming the work is **moved, not lost**.

The mechanism worth knowing: listing `packet` in `__additional__` makes `Info` store it under a mangled key and map it back, so it stays in `to_dict()`, `keys()` and `repr()` under its own name while attribute reads fall through to `__getattr__`. `__contains__` is overridden so `'packet' in datagram` does not trigger a full parse just to answer.

**Behaviour-neutrality checked three ways**: every observable of a `Datagram` — attribute read, `len`, iteration, `to_dict`, `dict()`, `keys`, `items`, `get`, `in`, `str`, `repr`, `hasattr`, `copy`/`deepcopy`/`pickle` — identical before and after, for complete *and* incomplete datagrams, with `packet` read first and not at all; all reassembly results over all 14 fixtures in both `ip` and `tcp` shapes identical (1509 lines); and all 14 fixtures serialised to `tree` and `json` ± reassembly byte-identical over 60 files / 33 MB.

## Tests

- `tests/integration/test_reassembly_engine_parity.py` — all four adapters over identical octets. **Fails on the pristine tree with 9 subtest failures reproducing #415's table verbatim.**
- `tests/foundation/reassembly/test_ipv6.py` — chain walk, AH encoding, idempotence, short headers.
- `tests/toolkit/test_pcap_unit.py` is the **only** existing test changed: two assertions pinned `v6.header == b'V' * 48`, i.e. a header with the Fragment header still in it — exactly the value #415 reports as wrong. Now 40, with `ihl`, `tl` and the `tl - ihl` invariant pinned alongside. Its IPv6 fake gained the `length` attribute the real `IPv6_Frag` has, and its `raw_len` now agrees with its own payload.

**Full suite: 851 passed, 17 skipped, 0 failed.** mypy identical to pristine (124 errors). pylint message multiset identical — `cyclic-import` excluded from that comparison because it is non-deterministic: two consecutive runs of the *unchanged* pristine tree gave 115 then 124.

## The design change deliberately not made — your call

Filtering unfragmented frames out entirely means adding `if not flags.mf and not offset: return None` beside the existing DF check in `toolkit/pcap.py:56` and `pcapng.py`.

- **It buys** another two-thirds of what laziness left: `reasm-ip` on `http.pcap` 1111.2 → **1050.6 ms**, feature cost +77.3 → **+24.5 ms**. Most of that is the per-frame `bytearray(8191)` + `bytearray(65535)` allocation, which is also where the residual GC pressure lives.
- **It costs** consumers this: across all 15 fixtures IP reassembly currently yields **1237 datagrams, of which exactly one is genuinely fragmented** (`ipv6.pcap` frames 13–16). Anyone reading `extractor.reassembly.ipv4` as "one entry per IP packet, reassembled where needed" would see `http.pcap` go 1117 → 0, `test.pcap` 21 → 0, `many_interfaces.pcapng` 34 → 0.

It may well be intended — a trivially-complete datagram *is* the reassembly of a one-fragment packet — which is why it is untouched.

Two follow-ups, neither in scope here: `docs/source/pep.rst:293-298` still lists this finding as fully open, and its "making `packet` lazy is mechanical" half is now done; and `reassembly/tcp.py:298` has the same eager `analyze`, though it is FIN/RST-driven (222 submits per `http.pcap` pass against 1117), so it wants its own change.


pcapkit/foundation/reassembly/data/ip.py
```diff
@@ -25,6 +25,51 @@
 BufferID: 'TypeAlias' = Tuple[_AT, _AT, int, 'TransType']
 
 
+class Deferred:
```
JarryShaw: only applies to IP protocols, not to TCP and other future protocols?

JarryShaw: Right on both counts — done in a9b8603c.

**It was misfiled.** `Deferred` holds an analyser, a protocol and a payload and calls the analyser; nothing in it is IP-specific. It only sat in `data/ip.py` because IP reassembly was the caller that needed it. `ReassemblyData` had the same problem from the other side — a class living in a package initialiser.

Both are now in `pcapkit/foundation/reassembly/data/data.py`, following `pcapkit/protocols/data/data.py`, and `data/__init__.py` is re-exports only. `Deferred` stays importable from `data.ip` for anything that already had it.

Its docstring now separates the mechanism from the case that motivated it, and records that **TCP reassembly can use it unmodified** — `TCP.submit` builds its `packet` eagerly too. It is a much smaller cost there, being FIN/RST-driven rather than per-frame (222 submits per `http.pcap` pass against 1117), so it is left for its own change rather than widened into this one.

**Flow tracing needs nothing, though, and that is worth recording rather than leaving as a maybe.** I looked before answering: its data models carry no parsed protocol at all — `Packet` holds the already-extracted `frame` it was handed, and `Buffer`/`Index` hold a dumper, an index list and a label — and `grep` finds no `analyze()` call anywhere in `foundation/traceflow/`. So there is no second parse to postpone. Its per-packet cost was the dumper rebuilding a whole `Frame` to get bytes it already held, which is a different defect and is fixed on `perf/dumper-and-options` (82% of tracing overhead, +1283 ms → +233 ms).

One asymmetry left deliberately: `TraceFlowData` still lives in `traceflow/data/__init__.py`, exactly as `ReassemblyData` did. Same one-file move if you want the two packages symmetric — say so and I will fold it in, but it is unrelated to this PR's subject.

Full suite 851 passed, 17 skipped.

pcapkit/foundation/reassembly/data/data.py
```diff
@@ -0,0 +1,87 @@
...
+    from pcapkit.const.reg.transtype import TransType
+    from pcapkit.foundation.reassembly.data.ip import Datagram as IP_Datagram
+    from pcapkit.foundation.reassembly.data.tcp import Datagram as TCP_Datagram
+    from pcapkit.protocols.protocol import ProtocolBase as Protocol
+
+
+class Deferred:
+    """A postponed analysis of a reassembled payload.
+
+A reassembled datagram's ``packet`` is a second, full parse of the payload
```
JarryShaw: indentation?

pcapkit/foundation/reassembly/ipv6.py
```diff
@@ -10,11 +10,66 @@
...
+#: Offset of the Next Header field within the fixed IPv6 header.
+_IPV6_NEXT_HEADER = 6
+
+#: Next Header value of the IPv6 Fragment header (:rfc:`8200#section-4.5`).
+_NH_IPV6_FRAG = 44
+
+#: Next Header value of the Authentication Header. It is the one extension
+#: header that does not measure its length in 8-octet units
+#: (:rfc:`4302#section-2.2`), so the header walk below has to special-case it.
+_NH_AH = 51
```
JarryShaw: lets ref the const enum directly.

pcapkit/foundation/reassembly/ipv6.py
```diff
@@ -10,11 +10,66 @@
...
 
+#: Length of the fixed IPv6 header, i.e. the offset of the first extension
+#: header (:rfc:`8200#section-3`).
+_IPV6_HDR_LEN = 40
+
+#: Offset of the Next Header field within the fixed IPv6 header.
+_IPV6_NEXT_HEADER = 6
+
+#: Next Header value of the IPv6 Fragment header (:rfc:`8200#section-4.5`).
+_NH_IPV6_FRAG = 44
```
JarryShaw: let's ref the const enum directly.

pcapkit/foundation/reassembly/ipv6.py
```diff
@@ -10,11 +10,66 @@
...
+    from pcapkit.const.reg.transtype import TransType
+
 __all__ = ['IPv6']
 
+#: Length of the fixed IPv6 header, i.e. the offset of the first extension
+#: header (:rfc:`8200#section-3`).
+_IPV6_HDR_LEN = 40
+
+#: Offset of the Next Header field within the fixed IPv6 header.
+_IPV6_NEXT_HEADER = 6
```
JarryShaw, line 31: lets ref the const enum directly.

pcapkit/foundation/reassembly/ipv6.py
```diff
@@ -10,11 +10,66 @@
...
 from pcapkit.protocols.internet.ipv6 import IPv6 as IPv6_Protocol
 
+if TYPE_CHECKING:
+    from pcapkit.const.reg.transtype import TransType
+
 __all__ = ['IPv6']
 
+#: Length of the fixed IPv6 header, i.e. the offset of the first extension
+#: header (:rfc:`8200#section-3`).
+_IPV6_HDR_LEN = 40
```
JarryShaw, line 27: avoid using the global const if its available from the protocol class' definition

pcapkit/foundation/reassembly/ipv6.py
JarryShaw: Done in 6f43873 — `44` and `51` are now `Enum_TransType.IPv6_Frag` and `Enum_TransType.AH`. Having the number in a comment was the wrong way round; the enum says it in the code.

`_NH_IPV6_FRAG` is gone entirely, since the comparison reads better against the member directly. `_NH_AH` survives as a *name* bound to `Enum_TransType.AH`, because its comment carries the reason the walk singles that header out: AH is the one extension header that does not measure its length in 8-octet units (:rfc:`4302#section-2.2`). Losing the name would lose that.

pcapkit/foundation/reassembly/ipv6.py
JarryShaw: Done in 6f43873 — `44` and `51` are now `Enum_TransType.IPv6_Frag` and `Enum_TransType.AH`. Having the number in a comment was the wrong way round; the enum says it in the code.

`_NH_IPV6_FRAG` is gone entirely, since the comparison reads better against the member directly. `_NH_AH` survives as a *name* bound to `Enum_TransType.AH`, because its comment carries the reason the walk singles that header out: AH is the one extension header that does not measure its length in 8-octet units (:rfc:`4302#section-2.2`). Losing the name would lose that.

pcapkit/foundation/reassembly/ipv6.py
JarryShaw: Done in 6f43873 — `44` and `51` are now `Enum_TransType.IPv6_Frag` and `Enum_TransType.AH`. Having the number in a comment was the wrong way round; the enum says it in the code.

`_NH_IPV6_FRAG` is gone entirely, since the comparison reads better against the member directly. `_NH_AH` survives as a *name* bound to `Enum_TransType.AH`, because its comment carries the reason the walk singles that header out: AH is the one extension header that does not measure its length in 8-octet units (:rfc:`4302#section-2.2`). Losing the name would lose that.

pcapkit/foundation/reassembly/ipv6.py
JarryShaw: Checked that before choosing, and there is nothing on the protocol class to prefer: `__protocol_type__` is `None` on `IPv6`, `IPv6_Frag` and `AH` alike, so the const enum is the most specific thing available. Fixed in 6f43873 using `Enum_TransType`.

Two of the four constants stay literal, and now say why in the code rather than leaving you to wonder: `_IPV6_HDR_LEN` (40) and `_IPV6_NEXT_HEADER` (6) are **byte offsets into the header**, not protocol numbers, so no enumeration carries them. 40 is derivable by summing the IPv6 schema fields, but doing that at import time to avoid writing a number the RFC fixes seemed worse than citing :rfc:`8200#section-3` next to it.

If you would rather they came from the schema too, say so — it is a small change, just one I judged as obscuring rather than clarifying.

pcapkit/foundation/reassembly/data/data.py
JarryShaw: Already fixed, in 8ae95a3a, before this comment landed — the pushed file now reads:

```
    A reassembled datagram's ``packet`` is a second, full parse of the payload
    the datagram just reassembled. Nothing about postponing it is specific to any
```

That was my damage: when I rewrote the docstring to separate the mechanism from the IP-specific motivation, I left the first body line flush against the margin while its continuations kept their four spaces. Caught it while wiring the TCP side.

pcapkit/foundation/reassembly/ipv6.py
```diff
@@ -10,11 +10,66 @@
...
+    position = _IPV6_HDR_LEN
+    proto = header[offset]
+
+    # NOTE: ``header[position]`` is the Next Header field of the extension
+    # header starting at ``position`` and ``header[position + 1]`` its length,
+    # but which of the two length encodings applies is decided by the *previous*
+    # header's Next Header value -- hence ``proto`` trailing one step behind.
+    while position + 1 < len(header):
+        offset = position
+        if proto == _NH_AH:
```
JarryShaw: lets directly reference the const enum here then. no need for the global const literal.

pcapkit/foundation/traceflow/data/data.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""shared data models for flow tracing"""
```
JarryShaw, line 2: TraceFlow's docs not updated.

pcapkit/foundation/reassembly/ipv6.py
JarryShaw: Done in 454bce78 — `_NH_AH` is gone and the comparison reads `Enum_TransType.AH` directly. The constant existed only to carry the "why AH is special" comment, so that comment moved to the use site instead:

```python
# NOTE: AH is the one extension header that does not measure its
# length in 8-octet units (:rfc:`4302#section-2.2`), which is why it
# is singled out rather than falling to the common case below.
if proto == Enum_TransType.AH:
```

Four-adapter parity re-checked after the change: 40/40/1488 with `header[6] == 17` on default, dpkt and scapy alike.

pcapkit/foundation/traceflow/data/data.py
JarryShaw: Fixed in 454bce78 — and the gap was larger than the comment suggested, in two ways I had missed.

**`DeferredPacket` was undocumentable.** It was in `data/data.py`'s `__all__` but never re-exported from the package, so autodoc failed outright: `AttributeError: module 'pcapkit.foundation.reassembly.data' has no attribute 'DeferredPacket'`. Now re-exported.

**`Deferred` was already documented in the wrong place.** `reassembly/ip/ip.rst:49` carried an entry for `...data.ip.Deferred` from when the class lived there, so the entry I added beside `ReassemblyData` made it a *duplicate registration*. The stale `ip.rst` entry is removed — the class moved, so its documentation moves with it — and the two `:class:` references in `ipv4.rst` and `ipv6.rst` are repointed.

**`TraceFlowData` itself was already documented**, in `traceflow/index.rst`; what was missing was the same treatment for the module it now lives in.

One thing I tried and reverted: `.. module::` directives for both `data.data` modules. They double-register everything the autoclass paths already cover, which is precisely what produced the duplicate warning — so the build told me to take them out again.

Verified by a clean build rather than by eye:

```
total WARNING/ERROR:            85   (exactly main's baseline)
duplicate Deferred:             0
DeferredPacket import failure:  0
any warning naming data.data:   0
rendered: Deferred, DeferredPacket, ReassemblyData, TraceFlowData
```

Full suite 857 passed, 17 skipped.

Issue #425, opened 2026-09-16
Option, chunk and block registries leak on lookup miss, the same way __proto__ did


JarryShaw:
#421 covered the `__proto__` next-layer registries. The `__option__` / `__chunk__` /
`__block__` / `__param__` / `__cause__` family has the identical defect and was
out of scope for the fix in #426.

Reproduced, on a tree that already has #421's fix:

```python
before = set(TCP.__option__)
hdr = bytes.fromhex('005001bb00000000000000006002ffff00000000') + bytes([156, 2, 0, 0])
TCP(hdr, len(hdr))                      # one packet, one unknown option kind

set(TCP.__option__) - before            # {<Option.Reserved_156: 156>}
register_tcp_option(Option(156), ...)   # 'option 156 already registered, overwriting'
```

So parsing a single TCP packet carrying an unrecognised option kind makes a later,
entirely legitimate registration report an overwrite that never happened — #421's
exact symptom, on a different registry.

## Scope

Eight `defaultdict` registries share the shape, enumerated by inspection rather
than by grep:

```
TCP.__option__        MH.__option__         HOPOPT.__option__
IPv6_Opts.__option__  PCAPNG.__option__     PCAPNG.__block__
SCTP.__chunk__        SCTP.__cause__
```

Read sites include `tcp.py:669,1948,1979`, `mh.py:1536,3114,3128`,
`hopopt.py:470,1278,1309`, `ipv6_opts.py:481`, `pcapng.py:953,1013,2043,3928,3955`,
plus the SCTP chunk/parameter/cause paths.

## Fix

#426 adds `ProtocolBase._lookup_next_layer(registry, proto)`, which resolves a hit
and memoises it while returning `registry.default_factory()` on a miss **without**
recording it. These registries hold method-name pairs rather than protocol classes,
so the helper is not directly reusable, but the shape of the fix is the same and
the same helper could be generalised.

Worth a test per registry family asserting a miss leaves the key set unchanged —
that is what caught the `__proto__` cases and what is missing here.

## Consequences, same as #421

- A spurious `RegistryWarning` pointing at a conflict that does not exist.
- One entry accreted per distinct unknown code seen, for the process lifetime, on
  an attribute shared by every instance of the class.
- "Is this option registered?" becomes un-answerable by inspection, since the
  answer depends on what has been parsed.

Found while fixing #421; deliberately not folded into #426 because three of the
files involved were outside that change's scope.

---

[...]

PR #426, opened 2026-09-16
protocols: stop next-layer dispatch writing to shared registries, and fix two transport defects

### LINKED ISSUE #418: Unknown next-layer codes keep their enumeration on SCTP and IP but not on TCP/UDP
`Data_Raw.protocol` is documented as "Original enumeration of this protocol", but
whether it is actually populated for an *unknown* code depends on which layer
asked. Measured on the generated fixtures:

| layer | unknown code | `Raw.info.protocol` | protochain |
|---|---|---|---|
| SCTP | PPID 4243 | `4243` | `SCTP:Unassigned_4243` |
| IPv4 | protocol 253 | the enum | `IPv4:Use_for_experimentation_and_testing_253` |
| TCP | port 22, unregistered | **`None`** | `Ethernet:IPv4:TCP:Raw` |

```python
ex = extract(fin='examples/captures/tcp.pcap', nofile=True, store=True)
tcp = ex.frame[5].payload.payload.payload
tcp.info.dstport            # ssh [22 - tcp|udp|sctp]
tcp.payload.info.protocol   # None
```

## Why

`Protocol._import_next_layer` passes `alias=proto` to whatever it constructs, and
`Raw.read` turns that into `Data_Raw.protocol`. `Internet.__proto__` and
`SCTP.__proto__` fall through to `Raw` *via that call*, so the code survives.

`Transport._decode_next_layer` resolves ports through `cls.__proto__[ports[0]]` /
`[ports[1]]` (`pcapkit/protocols/transport/transport.py:112-114`) and reaches
`Raw` by a route that never carries the alias, so TCP and UDP anonymise the port
they could not place.

## Why it matters

The value is most useful precisely when the protocol is unknown — it is the only
record of what the payload claimed to be. A consumer walking frames cannot tell
"TCP payload on some port we do not decode" from "TCP payload on port 22" today,
while the equivalent SCTP and IPv4 cases both say so.

## Note

Not a regression, and not what #417 changed. #417 makes the *failure* path
(`beholder`, i.e. a registered protocol whose parse raised) forward the alias
uniformly, which it previously did not — necessary there because SCTP's
unregistered path does keep the enumeration, so registering NGAP on PPID 60 would
otherwise have made a failed parse *less* informative than leaving the PPID
unregistered. That leaves the failure path consistent across layers and these
unknown-code paths still inconsistent with each other, which is this issue.

Fixing it means giving `Transport._decode_next_layer` the same alias forwarding
`_import_next_layer` does. `tests/protocols/transport/test_tcp_runtime.py:85` and
`tests/protocols/transport/test_sctp_unit.py:966` pin the two current behaviours
and are where the change would show.

### LINKED ISSUE #421: Parsing a packet pollutes the shared __proto__ registry, so a later register_* warns about a protocol nobody registered
`ProtocolBase._import_next_layer` indexes a `defaultdict` at
`pcapkit/protocols/protocol.py:1300`, so a lookup **miss inserts the key**. The
registry is class-level and shared, so ordinary parsing mutates it.

Reproduced on `main`:

```
ICMP in Internet.__proto__ before: False
ICMP in Internet.__proto__ after:  True
register_transtype now warns: ['protocol 1 already registered, overwriting']
```

Parsing a single 24-byte IPv4 packet carrying `proto=1` inserts
`TransType.ICMP -> Raw` into `Internet.__proto__`. A subsequent, entirely
legitimate `register_transtype(TransType.ICMP, …)` then reports that ICMP was
"already registered, overwriting" — about a registration that never happened.

Same pattern at `pcapkit/protocols/internet/internet.py:248` and
`pcapkit/protocols/internet/ipv6.py:409`.

## Consequences

- A spurious `RegistryWarning` that sends the reader looking for a conflicting
  registration that does not exist.
- The registry grows one entry per distinct unknown code seen, for the process
  lifetime, on a class attribute shared by every instance.
- It makes "is this protocol registered?" un-answerable by inspection, because
  the answer depends on what has been parsed.

## There is already a template for the fix

`SCTP._import_next_layer` overrides the base method precisely to avoid this and
documents why at `pcapkit/protocols/transport/sctp.py:818-831`. The same
treatment is wanted for `Frame`, `PCAPNG`, `Link`, `Internet`, `IPv6`, `TCP` and
`UDP` — or, better, in the base method so each subclass does not have to
remember.

Note this is related to but distinct from the `__output__` `defaultdict`
discussed earlier: there the injection is load-bearing memoisation, deliberately
kept. Here nothing depends on the insertion — the value inserted is just `Raw`.

Found while auditing for #419.

### LINKED ISSUE #423: TCP(**kwargs) raises AttributeError instead of constructing a packet
Constructing TCP from keyword arguments fails:

```python
>>> from pcapkit.protocols.transport.tcp import TCP
>>> TCP(srcport=80, dstport=443)
AttributeError: 'int' object has no attribute 'port'
```

Raised at `pcapkit/protocols/transport/tcp.py:479`, which reads `.port` off a
value it was handed as a plain `int`.

Reproduced on `main`. This is the documented construction path for every other
protocol, so either the argument handling needs to accept the integer form it
advertises, or the expectation needs stating — but an `AttributeError` naming
`int` tells the caller nothing about what it should have passed.

Found while profiling for #420 (not a performance issue — noticed in passing).

JarryShaw:
Closes #421. Closes #418. Closes #423.

Five commits, each its own concern. Verified independently against `origin/main` — I re-ran every reproduction rather than taking the numbers on trust.

## #421 — parsing a packet polluted a shared class-level registry

`ProtocolBase._import_next_layer` indexed a `defaultdict`, so a lookup **miss inserted the key**. Before → after:

```
ICMP in Internet.__proto__ after parsing:  True  →  False
register_transtype(ICMP, …) warns:         'protocol 1 already registered,
                                            overwriting'  →  nothing
TCP.__proto__ after parsing:               [21, 80, None]  →  [21, 80]
```

**The fix went in the base**, as `ProtocolBase._lookup_next_layer(registry, proto)`: hit → resolve and memoise as before; miss → return `registry.default_factory()` *without* recording it. Six read sites route through it — `ProtocolBase.analyze`, `ProtocolBase._import_next_layer`, `Internet._import_next_layer`, `IPv6._import_next_layer`, `Transport.analyze`, and the ethernet schema callback. `Frame`, `PCAPNG`, `Link`, `TCP` and `UDP` inherit it.

Two of those the issue did not name: **`Transport.analyze`'s `ports[1]` branch** had the same miss, and **`schema/link/ethernet.py:27`** subscripted `Ethernet.__proto__` — which *is* `Link.__proto__` — inside a `PayloadField` callback, so one frame with an unregistered EtherType leaked into the shared registry. Verified fixed:

```
EtherType 0x88B5 registered after parsing:  True  →  False
```

That callback also gains the `ModuleDescriptor` memoisation it never had, so it stops re-importing per frame.

It takes the registry as an argument rather than reading `cls.__proto__`, and that is load-bearing rather than stylistic: `test_protocol_base_unit.py:317` sets `__proto__` on an *instance*, which the old `self.__proto__` honoured. A `classmethod` broke that test; the argument preserves the semantics.

**`SCTP._import_next_layer` is now redundant and removed.** It differed from the base only in the non-inserting lookup (now in the base) and a *missing* `__context__=self._exctx` — so removing it also restores parsing-context propagation to SCTP payloads.

A sweep confirmed six is the whole set for `__proto__`: of nine `__proto__[` hits, seven are assignments inside `register()` and one is a plain `dict`. `grep` across `protocols/schema/` returns that one ethernet line and nothing else.

## #418 — TCP and UDP anonymised the port they could not place

```
tcp.info.dstport          : ssh [22 - tcp|udp|sctp]
tcp.payload.info.protocol : None  →  22
frame.protochain          : Ethernet:IPv4:TCP:Raw   (unchanged)
```

`Transport._decode_next_layer` now forwards `sort_port[0]` — the lower port, already its primary lookup key — instead of `None`. The protochain deliberately does *not* become `ssh_22`: ports arrive as plain `int`, because `AppType` is a `StrEnum` and `sorted()` over members would order them by string and break the lower-port rule. A plain int has no `.name`, so `Raw` still labels itself `Raw`, and `Data_Raw.protocol` is declared `Optional[int]`, so an int is exactly the contract.

`test_tcp_runtime.py:85` asserted `assertIsNone(...)` — that was the defect rather than a contract, and is now `22`. `test_http_runtime.py:105` still asserts `None` and still passes: that payload reaches `Raw` through `beholder`, a different path.

**A pointer correction for anyone reading #418**: its reference to `test_sciop_unit.py:966` is wrong. At the commit the issue was filed against, line 966 was `test_register_rejects_a_non_protocol`; the test it means is `test_unregistered_ppid_does_not_mutate_the_class_registry`, now at 951–1016 since #417 grew the file. Find it by name. SCTP's behaviour is the reference here and is unchanged.

## #423 — `TCP(**kwargs)` raised

**Decision: the integer is accepted and coerced.** `TCP.make` advertises `srcport: 'Enum_AppType | int'`, so an int is documented input, not misuse. The schema field converted one only on the way *out* to bytes and never wrote it back, and `ProtocolBase.unpack` reuses the schema `make` built rather than re-parsing — so `read` saw an int where a parsed packet carries an `AppType`.

```
>>> TCP(srcport=80, dstport=443)
srcport=<AppType.www-http: 80 [tcp|udp]>  dstport=<AppType.https: 443 [tcp|udp|sctp]>
```

Checking the siblings widened it: **UDP raised identically**, and **SCTP did not raise only because it keys its next layer on the DATA chunk PPID and never reads `.port`** — it simply left an int in a field declared `AppType`. All three now normalise through `Transport._make_port(port, proto)`. The transport protocol is a parameter because it distinguishes TCP/1 (`tcpmux`) from SCTP/1 (unassigned); a test pins that.

## Fifth commit — one exception-type outlier

`Transport.register` raised bare `TypeError` where five sibling `register()` methods raise `RegistryError` with the *identical* message (`internet.py:141`, `link.py:119`, `misc/pcapng.py:782`, `misc/pcap/frame.py:139`, `foundation/registry/protocols.py:148`) — and the line above it already raised `UnsupportedCall`, so the method was inconsistent with itself. `RegistryError` is `(BaseError, TypeError)`, so anything catching `TypeError` still catches it.

## Verification

- **Byte comparison** (15 captures × `tree`+`json`, `ip=True, tcp=True, reassembly=True`, against a pristine `git archive` of the base): the #421 commit alone is **byte-identical**, 30/30. The full stack differs in 24 of 30 files, and **every differing line without exception** is a `Raw` payload's `protocol` going `null`→N / `NIL`→N — 120 lines in each format, no protochain change, no other field. Ports now recorded: 67, 53, 443, 123, 22, 5353, 7000, 17500, 12345, 10352, 12089, 39645, 5001, 0.
- **Full suite: 836 passed, 17 skipped, 0 failed** (808 subtests). The pristine baseline is 812 passed / 35 skipped; the 18 extra skips there are `test_tier_guard.py` declining to run outside a git checkout, so 812 + 18 + 7 new tests = 836. Nothing that passed before fails.
- New tests fail on a pristine tree — checked rather than assumed. `test_internet_unit.py`'s reproduction gives `Items in the first set but not the second: <TransType.ICMP: 1>`.
- mypy: same pre-existing errors as baseline, none on new lines. `isort` and `pylint` clean.
- No perf regression from dropping the miss-memoisation: `http.pcap` ×3 at 7.50 s against 7.61 s baseline.

## Not folded in

**#426 — the sibling registry family has the identical defect.** `__option__`, `__chunk__`, `__block__`, `__param__` and `__cause__` leak the same way; eight registries share the shape, and one TCP packet with option kind 156 makes `register_tcp_option(156, …)` warn about an overwrite that never happened. Three of the files involved are outside this change's scope, so it is filed separately.

**Not rebased, deliberately.** `origin/main` moved to #419 while this was being verified, so the branch is 5 ahead / 1 behind. Rebasing would rewrite four already-pushed commits and need a force-push. #419 touches only `docs/source/pep.rst`, which this never touches; `git merge-tree --write-tree origin/main HEAD` exits 0 with no conflicts, which I confirmed.


JarryShaw:
this only applies to protocol registry? what about all other defaultdict registries. seems like they're all in scope. are they covered?

pcapkit/protocols/protocol.py
```diff
@@ -1235,6 +1237,50 @@ def _make_payload(cls, data: 'Data') -> 'ProtocolBase':
 
         return proto.from_data(data[name])
 
+    @staticmethod
+    def _lookup_next_layer(registry: 'DefaultDict[int, ModuleDescriptor[ProtocolBase] | Type[ProtocolBase]]',
```
JarryShaw, line 1241: this only applies to protocol registry? what about all other defaultdict registries. seems like they're all in scope. are they covered?

JarryShaw: They are all in scope, they are not all covered here, and that gap is filed as #425 with a reproduction.

**Covered by this PR — the `__proto__` family, all six read sites.** A sweep settled that six is the whole set: of nine `__proto__[` hits, seven are `cls.__proto__[code] = protocol` *writes* inside `register()` (assignments cannot inject), one is a plain `dict` in `protocols/__init__.py`, and the last was the ethernet schema callback, fixed in 8918c809. `grep` across `protocols/schema/` returns that one line and nothing else.

**Not covered — the option/chunk/block family.** Eight registries share the shape, enumerated by inspection rather than by grep:

```
TCP.__option__      MH.__option__       HOPOPT.__option__   IPv6_Opts.__option__
PCAPNG.__option__   PCAPNG.__block__    SCTP.__chunk__      SCTP.__cause__
```

Reproduced end to end on a tree that already has this PR's fix:

```python
before = set(TCP.__option__)
TCP(hdr, len(hdr))                      # one packet, one unknown option kind 156
set(TCP.__option__) - before            # {<Option.Reserved_156: 156>}
register_tcp_option(Option(156), ...)   # 'option 156 already registered, overwriting'
```

Same symptom as #421, different registry.

Left out deliberately rather than overlooked, for two reasons. Three of the files involved — `mh.py`, `hopopt.py`, `ipv6_opts.py` — were outside this change's file scope while other work was in flight in the same tree. And these registries hold *method-name pairs* rather than protocol classes, so `_lookup_next_layer` is not directly reusable: the shape of the fix carries over, but the helper needs generalising first.

#425 records the read sites and suggests a test per registry family asserting that a miss leaves the key set unchanged — which is what caught the `__proto__` cases, and exactly what was missing here.

PR #435, opened 2026-09-17
foundation: give IP reassembly the RFC timeout, and trace TCP flows bidirectionally

JarryShaw:
Closes the two reassembly and traceflow items on the Help Wanted page, `docs/source/pep.rst` under *"Reassembly Beyond IP and TCP"* (added by #419, mirroring discussion #106 `#discussioncomment-18470233`):

> The TCP tracer itself closes a flow on FIN but never on RST, which is not in its packet model at all, and **treats each direction of a connection as a separate flow**.

> **Nothing ever times a partial datagram out.** RFC 791 gives IP reassembly a 15-second timer and RFC 8200 gives IPv6 60 seconds; neither is implemented, and neither can be until the buffer models carry a timestamp.

## What the timeout means for an offline parser

**The clock is the capture's own timestamps.** Wall clock says nothing about a file, and `time.time()` would make the same capture reassemble differently on every run — which would break both the fixtures and any reproducible comparison. A fragment arriving is the only evidence that capture time advanced, so the sweep runs when a packet is handed over rather than on a timer.

That has a consequence worth documenting rather than hiding: the clock only advances while that particular reassembler is fed. For IPv4 and TCP that is nearly every frame; for IPv6 it is only fragments. So a stalled IPv6 buffer with nothing behind it reports `PARTIAL`, not `TIMEOUT` — the honest answer, since the capture never shows the deadline passing.

**60 seconds for both IPv4 and IPv6, and the page's 15 seconds was wrong.** RFC 1122 §3.3.2 supersedes RFC 791 here: the timeout "SHOULD be a fixed value, **not set from the remaining TTL**… between 60 seconds and 120 seconds". RFC 791's 15 s is an *initial* setting that `MAX(TIMER, TTL)` raises toward the 4.25-minute TTL ceiling, so it is not a deadline — and TTL is not in the packet model anyway. `pep.rst` is corrected in this PR. **TCP gets no timeout by default**: no specification gives stream reassembly one and an idle connection is ordinary, but `timeout=` enables one.

## Two behaviour changes, both opt-out-able

**1. `Datagram.completed` widens from `bool` to a `Completion` enum** — `COMPLETE` / `PARTIAL` / `TIMEOUT`, same field, same channel, so a caller learns why a datagram is incomplete without a parallel vocabulary. Only `COMPLETE` is truthy, so `if datagram.completed:` and `assertTrue`/`assertFalse` are unchanged. **`datagram.completed == True` no longer holds** — that is the breaking shape.

**2. Bidirectional TCP flow tracing, on by default**, with `trace_bidirectional=False` on `extract()`, `Extractor` and `follow_tcp_stream()`, and `bidirectional=False` on `TraceFlow`. `TCP.make_bufid()` orders the endpoints canonically so both halves reduce to one key; the label still comes from the packet that opened the flow, so labels are unchanged. `Index` gains `forward` and `reverse`. A flow closes once **both** halves have FINed — closing on the first would cut the peer's FIN and final ACK out, and they would then open a second flow under the same ID, which is the very split being removed.

Default-on because `pep.rst` lists per-direction keying as a **defect**, beside "closes on FIN but never on RST", rather than as a missing option — and because the module is called "Follow TCP Stream", which everywhere else means the conversation.

## Flow output changes, and the argument for it

**An earlier revision of this description had these numbers wrong, and the way they were wrong is worth recording**: it reported `http.pcap` at 220 flows, of which 109 were single-frame, and read those 109 as one-directional conversations. They were not — they were the defect. Each conversation was a 9-frame flow *plus a 1-frame stray tail*, created when the flow was finalised on the teardown and the final acknowledgement then opened a second buffer under the same ID. Fixing the close condition removes the tails.

| capture | main | this PR | single-frame flows |
|---|---|---|---|
| `http.pcap` | 331 | **111** | 109 → **0** |
| `tcp.pcap` | 4 | 2 | — |
| `in.pcap` | 3 | 2 | — |
| **corpus** | **355** | **122** | — |

`http.pcap` has 111 conversations and every one of them now carries both directions in a single flow. `bidirectional=False` reproduces `main` exactly at **355 flows**.

`traced_frames` is identical in every capture — **1222 total**, nothing lost and nothing double-counted, which is what makes the flow-count drop a merge rather than a loss. Concretely on `in.pcap`: frame 3 is `123.129.210.135:80 → 192.168.1.100:55232` and frame 4 is the same connection reversed; they now form one flow with `forward=(3,)` and `reverse=(4,)`. Same address pair, same port pair, opposite directions — one TCP connection.

Verified independently through `follow_tcp_stream` on `in.pcap`, with payload conserved in every configuration:

```
origin/main                     3 streams: 70 + 70 + 70 = 210 octets
this PR, default                2 streams: 140 + 70      = 210 octets
this PR, trace_bidirectional=0  3 streams: 70 + 70 + 70  = 210 octets   (identical to main)
```

So the merged stream carries both directions' payload, which is what following a stream means, and the opt-out reproduces `main` exactly.

## Also fixed, beyond the brief

- **Three `strict=False` cases, not two, and they are not symmetric — an earlier draft of this description got that wrong.**
  - **TCP**: payload unchanged — the buffer with its holes zero-filled, which is what `follow_tcp_stream` reconstructs a stream from. Only `completed` stopped claiming a holed datagram was complete.
  - **IP, final fragment received** (`TDL > 0`): payload unchanged, holes zero-filled, as TCP does. Only `completed` became honest.
  - **IP, final fragment never received** (`TDL` still `-1`): **the payload changed.** It was `datagram[:-1]` — 65534 octets of preallocated buffer, reported *complete*. It is now the contiguous prefix, offset zero to the first hole; not `b''`, because those octets really did arrive. `strict=True`, the default, still lists the runs.

  The original wording, kept for the record of what was wrong: TCP reported a hole-ridden buffer as complete, passing zero-filled gaps off as received data; there the payload *shape* really is unchanged, a test pins it and `follow_tcp_stream` depends on the blob, so only `completed` became honest. **IP is different.** It sliced `datagram[:TDL]` with `TDL` still `-1`, handing back **65534 octets of preallocated buffer** — almost all of them zeros the sender never sent — and calling it complete. That payload could not be left as it was. It now reports the **contiguous prefix**: every octet from offset zero up to the first hole. That is the longest run whose extent is known without knowing the total length, it is all genuinely received, and it is the part a caller asking for one blob can actually use, since a parser reading from the start cannot use a run that begins after an unmeasured gap. Reporting `b''` was considered and rejected as discarding data that really arrived; anything past the first hole is still reported by `strict=True`, which lists the runs precisely because their offsets cannot be conveyed in a blob. `RCVBT` records receipt in 8-octet units, so the prefix ends on that boundary.
- **Scapy's traceflow reported `time.time()`** as the flow timestamp, putting the parse moment into every label and output filename. Now `float(packet.time)`, i.e. deterministic.
- **`Datagram.__init__` overloads removed.** They promised complete ⇒ `bytes` and incomplete ⇒ `tuple`, which only held because loose mode was lying, and could not be selected from a runtime-computed value anyway.
- **DPKT is the one adapter that cannot self-supply a timestamp** — it returns `(ts, buf)` and the engine keeps only the packet — so its three reassembly adapters take it positionally. `follow_tcp_stream` binds `0.0` there via `functools.partial`, safe because TCP has no timeout by default, and documented at the call site.

## Deliberately not done

**Wiring traceflow's `analyze` is not in this PR, and the record says it should not be.** PR #424 `#discussion_r4031491063` settled it: traceflow buffers no parsed protocol at all — only a dumper, frame indices and a label — and `grep` finds no `analyze()` call anywhere under `foundation/traceflow/`. So there is no second parse to postpone; adding one is a *new capability* needing a design decision first (does the tracer grow a per-direction payload buffer, or delegate to `TCP_Reassembly` as `follow_tcp_stream` already does?). Recorded in `pep.rst` as the remaining open item there.

## Verification

Suite **893 passed / 17 skipped / 848 subtests / 0 failed**, against a branch-point baseline of 859 / 35 / 844 — collection 910 versus 894, exactly the +16 tests added. The 18 extra baseline skips are all `tests/test_tier_guard.py`, because a `git archive` scratch tree is not a checkout; worth knowing for anyone measuring a baseline that way.

All 15 captures produce byte-identical `tree`, `json` **and** reassembly output with `ip/tcp/reassembly=True`; **no capture hits a timeout** — 1479 datagrams, all `COMPLETE`. `make_samples.py` regenerates byte-identically. mypy 127 errors versus 128 on the baseline; pylint 4689 versus 4706, rating 8.76 → 8.77. Docs build clean for the changed content.


pcapkit/foundation/traceflow/tcp.py
```diff
@@ -109,33 +174,58 @@ def trace(self, packet: 'Packet[_AT]', *, output: 'bool' = False) -> 'Dumper | s
...
+
+        # when the session is over, submit its buffer
+        #
+        # NOTE: A bidirectional flow is a whole connection, and a connection is
+        # not finished while either direction still is -- so it takes a FIN from
+        # both endpoints, not the first FIN seen. Closing on the first would cut
+        # the peer's FIN and the final acknowledgement out of the flow, and they
+        # would then open a *second* flow under the same buffer ID: precisely the
+        # split this is here to remove.
+        closed = len(buffer.fin) >= 2 if self._bidir else FIN
```
JarryShaw: **Bidirectional close fires one packet too early, and the leftover buffer becomes a magnet for a later, unrelated connection that reuses the same port pair.**

`closed = len(buffer.fin) >= 2` (line 216) pops the buffer as soon as the *second* FIN is seen. In a normal 4-way close (`FIN A→B`, `ACK B→A`, `FIN B→A`, `ACK A→B`) that second FIN is not the last packet of the connection — the final ACK still follows. By the time that ACK arrives, `BUFID` has already been popped, so `BUFID not in self._buffer` is true again and the ACK opens a **new**, one-packet buffer under the same canonical key. That stray buffer never gets a second FIN, so it sits open in `self._buffer` until something else lands on the same key — including a later, completely unrelated TCP connection that happens to reuse the same 4-tuple, which then gets silently appended onto it.

Reproduced directly against `TCP.trace()` (synthetic packets, same shape as the existing `test_a_conversation_closes_only_once_both_halves_have_finished`, extended with the final ACK and a second, later connection reusing the exact 4-tuple):

```
1   A->B SYN
2   B->A SYN,ACK
3   A->B ACK
6   A->B FIN
7   B->A ACK
8   B->A FIN        # closes here: len(buffer.fin) == {A, B} == 2
9   A->B ACK         # the connection's OWN final ACK -> lands in a brand-new buffer
...
100 A->B SYN         # a genuinely different, later connection, same 4-tuple
101 B->A SYN,ACK
102 A->B ACK
103 A->B data
```

After frame 103, `trace._buffer` has exactly **one** open entry:
```
(192.0.2.1, 12345, 198.51.100.2, 443) label=..._2.3 index=[9, 100, 101, 102, 103]
```
Frame 9 (connection #1's trailing ACK) and the entirety of connection #2 have been merged into one flow. This is precisely the "two genuinely different connections reusing a port pair after close must not collapse into one flow" case, and it does collapse — via `follow_tcp_stream` this would feed both connections' segments into one `TCP_Reassembly` object keyed on the same 4-tuple, mixing two unrelated sequence-number spaces into one reconstructed "stream."

The same stray-buffer artifact also exists with `bidirectional=False` (I checked: the trailing ACK reopens a fresh per-direction buffer there too), but the blast radius is smaller — only the one direction's later traffic can land on it, not an entire subsequent connection's both directions, since a per-direction buffer only reuses that one direction's key.

I could not find a test that sends any packet after both FINs have closed a bidirectional flow, which is why this doesn't surface in the added suite (`test_a_conversation_closes_only_once_both_halves_have_finished` stops right at the second FIN). Given this directly undermines the "one flow per connection" guarantee the bidirectional default exists to provide, and does so on an ordinary graceful-close-plus-reuse pattern rather than an exotic one, I'd treat this as a real defect rather than a documentation gap.

pcapkit/foundation/reassembly/ip.py
```diff
@@ -194,11 +218,21 @@ def submit(self, buf: 'Buffer[_AT]', *, bufid: 'tuple[_AT, _AT, int, TransType]'
...
+            # NOTE: ``max(TDL, 0)``, not ``TDL``. ``TDL`` is still its initial
+            # ``-1`` until the fragment with **MF** clear arrives, so a datagram
+            # whose final fragment never came reached this branch under
+            # ``strict=False`` and sliced ``datagram[:-1]`` -- handing back 65534
+            # octets of the preallocated buffer, almost all of them zeros the
+            # sender never sent, and calling it complete. The length of such a
+            # datagram is simply not known, so there is nothing honest to report
+            # but an empty payload; ``strict=True``, the default, reports the runs
+            # that did arrive instead.
+            payload = bytes(datagram[:max(TDL, 0)])
```
JarryShaw: **The PR description's "payload shape is deliberately unchanged" claim doesn't hold for this fix — the IP `strict=False` + never-finished-datagram case now discards even the legitimately-received prefix.**

Before this PR, `payload = bytes(datagram[:TDL])` with `TDL` still `-1` (final fragment never arrived) sliced `datagram[:-1]` — 65534 bytes, mostly the zero-filled preallocated buffer, but *including* whatever real bytes had actually arrived at the front. This line now does `datagram[:max(TDL, 0)]`, i.e. `datagram[:0]` == `b''` in the same situation: empty, discarding the real data along with the garbage.

Verified directly, one fragment sent (`mf=True`, offset 0, 100 bytes of payload), final fragment never arrives, `strict=False`, then `.fetch()`:
```
main (44aa38ae8):   completed=True,    len(payload)=65534
branch (a939c4f3b): completed=partial, len(payload)=0
```
So both `completed` *and* the payload bytes changed for this path — the PR body's "only completed became honest" doesn't describe what actually happens here (it's accurate for the TCP `strict=False` fix in the same commit, where `buffer.raw` is genuinely unchanged and only `completed` flips).

This is a reasonable design choice on its own (65534 bytes of near-garbage reported as "complete" was worse), but as written it throws away real, received data that a caller in best-effort/`strict=False` mode presumably wants — the same PR keeps the received-so-far blob for the analogous TCP defect rather than emptying it. I couldn't find a test in `test_ip.py` or the new `test_timeout.py` that pins the new payload for this exact case, unlike the TCP fix, which is pinned by `test_submit_incomplete_strict_complete_strict_false_and_empty_buffers`. Worth either correcting the PR description, or reconsidering the empty-payload choice (e.g. reporting the received extent instead of an empty blob), plus a regression test either way.

JarryShaw:
## Review of #435 at `a939c4f3b3364649c6980794ec94d869d555e57b`

Reviewed as a stand-in for Copilot (out of tokens). Checked out the branch in an isolated worktree, cloned a pristine `main` (`44aa38ae8`) for comparison, and read the full 41-file diff.

**What I ran**
- Generated the sample corpus (`python examples/generators/make_samples.py`) — 15 captures, matching the PR body.
- Full suite: `PYTHONSAFEPATH=1 PYTHONPATH=<worktree> .venv/bin/python -m pytest tests/ -q` → **893 passed, 17 skipped, 848 subtests passed, 0 failed** — matches the PR body exactly.
- Extracted all 15 generated captures with `ip=True, tcp=True, reassembly=True` → **1479 datagrams, all `COMPLETE`**, matching the body's "no capture hits a timeout" claim.
- Verified the RFC quotes against the actual RFC text (fetched `rfc-editor.org/rfc/rfc791.txt`, `rfc1122.txt`, `rfc8200.txt`): the RFC 791 "15 seconds initial timer, `TIMER <- MAX(TIMER,TTL)`, ~4.25 min ceiling" description, the RFC 1122 §3.3.2 "SHOULD be a fixed value, not set from the remaining TTL... between 60 seconds and 120 seconds" quote, and the RFC 8200 §4.5 "within 60 seconds of the reception of the first-arriving fragment" quote are all **verbatim-accurate**. The claim that RFC 1122 supersedes RFC 791's scheme is a fair reading of the text. No finding here — this part of the PR description checks out.
- Independently constructed synthetic packets (bypassing the toolkit adapters) to probe: `Completion` truthiness across every `.completed` consumer in the repo (grepped the whole tree — every test uses `assertTrue`/`assertFalse`/`assertIs`, none compare `== True`/`== False`, all consistent with the enum's `__bool__`), timestamps going backwards (deterministic, no crash), an infinite/negative/explicit timeout, PCAP-NG differing per-interface timestamp resolution (already normalized upstream by `_get_resolution`, unaffected by this PR), and the DPKT `functools.partial(..., timestamp=0.0)` binding used by `follow_tcp_stream` (safe, since TCP's default timeout is `math.inf`).
- Reproduced the two `strict=False` fixes directly against both trees (see the two inline comments) and the TCP bidirectional-tracing closing/merging behaviour against synthetic 4-way-close + port-reuse sequences.

**Findings posted inline (2):**
1. `pcapkit/foundation/traceflow/tcp.py:216` — the bidirectional close condition fires on the connection's second FIN, before its final ACK typically arrives, so that ACK opens a stray one-packet buffer under the same canonical key; a **later, unrelated connection reusing the same 4-tuple then merges into that stray buffer**, undermining the "one flow per connection" property the bidirectional default exists to provide. Reproduced with concrete frame numbers.
2. `pcapkit/foundation/reassembly/ip.py:233` — for the IP `strict=False` + never-completed-datagram fix, the PR description's "payload shape is deliberately unchanged" doesn't hold: the payload goes from 65534 (mostly garbage) bytes to 0 bytes, discarding the legitimately-received prefix too, and no test pins the new value the way the analogous TCP fix is pinned. Verified against both trees.

**What I did not fully verify:** a byte-for-byte corpus-wide diff of `tree`/`json` dump output between `main` and the branch (I spot-checked the reassembly completion counts across all 15 captures instead, which matched); the mypy/pylint counts (explicitly out of scope per the review brief); and I did not exercise the RST-not-handled gap beyond confirming it's still accurately documented (no `rst` field exists in the traceflow `Packet` model, as claimed).

**Verdict:** Not clean. Finding 1 is a real correctness gap in the PR's own headline feature — a common capture pattern (graceful close followed by port reuse) can silently merge two unrelated TCP connections into one reported flow / one reconstructed `follow_tcp_stream` conversation. I would not call this good-to-merge as-is; I'd want finding 1 fixed (or at least explicitly documented as a known limitation) before merging, and finding 2 addressed or the description corrected. Everything else I checked — the RFC claims, the enum truthiness change, the timeout determinism, the test-suite and datagram-count claims — held up exactly as described.

pcapkit/foundation/traceflow/tcp.py
JarryShaw: Fixed in `b937abca2`, and the fix goes wider than the finding — the close condition is gone rather than corrected.

**A teardown no longer ends a flow.** Seeing a connection close is not proof that no more of its packets can arrive: the acknowledgement completing the close follows the second FIN, and duplicates of that acknowledgement can follow in turn. So the teardown is now *recorded*, and a flow is finalised only by proof that nothing more can come — a **new connection appearing under the same buffer ID**, or the end of the capture via `TraceFlow.finish()`. Callbacks run at finalisation, so they now see the whole conversation.

That removes the class of defect rather than moving it one packet later, which is what tuning `len(buffer.fin) >= 2` would have done. It also means the reasoning in my own NOTE — that closing early cuts the peer's FIN and final ACK out and reopens a second flow under the same ID — is now enforced instead of merely described.

Six tests cover the edges, including the ones this finding and my brief called out:

```
test_a_four_way_close_keeps_its_final_acknowledgement
test_a_reused_port_pair_does_not_join_the_closed_connection
test_a_reset_ends_the_connection_as_a_close_does
test_a_half_open_connection_is_still_reported_at_end_of_capture
test_the_peers_syn_ack_joins_the_flow_rather_than_splitting_it
test_both_halves_of_a_connection_are_one_flow
```

The RST case is worth noting specifically: `docs/source/pep.rst` records that RST "is not in its packet model at all", so that is a gap closed alongside this rather than an incidental extra.

pcapkit/foundation/reassembly/ip.py
JarryShaw: Both halves fixed in `b937abca2` — the behaviour and the description, which were wrong in different ways.

**The description first, since that was the sharper part of the finding.** The PR body claimed "the payload *shape* is deliberately unchanged" for both `strict=False` fixes. That is true of TCP and was false of IP, and the body has been corrected to say so explicitly rather than glossing the two together.

**The behaviour is no longer `b''` either.** Reporting nothing was the other extreme from returning 65534 octets of preallocated buffer, and it discarded data that really did arrive. It now reports the **contiguous prefix** — every octet from offset zero up to the first hole:

> That is the longest run whose extent is known without knowing the total length, it is all genuinely received, and it is the part a caller asking for one contiguous payload can actually use — a parser reading a payload from its start cannot use a run that begins after an unmeasured gap anyway. Anything past the first hole is still reported by `strict=True`, which lists the runs precisely because their offsets cannot be conveyed in a blob.

`RCVBT` records receipt in 8-octet units, so the prefix ends on that boundary. That is the design question the finding raised — whether an empty payload was genuinely the best answer — answered in the code rather than left implicit, which is where it belonged.

The branch also gains `tests/foundation/reassembly/test_timeout.py`, so this path is now covered where before only the analogous TCP fix was.

docs/source/pep.rst
```diff
@@ -529,16 +529,105 @@ Two smaller items in the same subsystem:
...
+  :class:`~pcapkit.foundation.reassembly.tcp.TCP` the way
+  :func:`~pcapkit.interface.misc.follow_tcp_stream` already does.
+
+  One case remains undecided rather than solved: a capture that *starts* in the
+  middle of a connection, sees no teardown, and then has its endpoints reused. The
+  reuse is indistinguishable from a continuation without the ACK flag on
+  :class:`~pcapkit.foundation.traceflow.data.tcp.Packet`, which would make
+  ``syn and not ack`` a definitive new-connection test on its own.
+
+  What is still wanted here is **wiring the application layer into flow
```
JarryShaw: Reviewing `a206459f3` (head of this PR, CI green: 21 checks SUCCESS, 2 SKIPPED, `pyup.io/safety-ci` SUCCESS).

The "wiring the application layer into flow tracing" paragraph is pasted twice, verbatim, in the flow-tracing bullet — once right after "What ends a bidirectional flow is worth stating..." (lines 563-572) and again right after "One case remains undecided rather than solved..." (lines 580-589). Confirmed with:

```
$ grep -n "wiring the application layer" docs/source/pep.rst
563:  What is still wanted here is **wiring the application layer into flow
580:  What is still wanted here is **wiring the application layer into flow
```

This looks like a copy-paste artifact from drafting rather than an intentional repeat, and it lands in exactly the spot the task brief asked me to scrutinize: the "one case left deliberately undecided" disclosure now reads as sandwiched between two identical copies of an unrelated paragraph, which undercuts the disclosure it's meant to support. Suggest dropping the first copy (563-572) and keeping the one that follows the "undecided" paragraph (or vice versa) so the section reads as one coherent bullet.

Not a code defect — the code matches the (de-duplicated) disclosure correctly, see PR-level comment.

JarryShaw:
Reviewed at head `a206459f3` (the sha named in the review brief; the PR has since gained an unrelated merge to `87099110f` bringing in HIP/IPv4 protocol changes from `main` — `git diff a206459f3 8709911` touches only `docs/source/ext.rst`, `docs/source/pcapkit/protocols/internet/{hip,ipv4}.rst`, `pcapkit/foundation/registry/protocols.py`, `pcapkit/protocols/internet/{hip,ipv4}.py` and their tests; nothing under `traceflow/`, `reassembly/`, or `pep.rst`, so everything below still applies unchanged).

**CI at `a206459f3`**: all green — 21 checks `SUCCESS`, 2 `SKIPPED` (`Docs test gate`, `Gate (full suite, Python 3.14)`, both legitimately skipped, not failed), plus `pyup.io/safety-ci` `SUCCESS`. Confirmed via `gh pr view 435 --json statusCheckRollup` directly against the API, not a local suite.

## What I checked

**Finding 1 (traceflow redesign).** Read `pcapkit/foundation/traceflow/tcp.py`, `traceflow/data/tcp.py`, `traceflow/traceflow.py`, `foundation/extraction.py`'s wiring of `finish()`/`trace_bidirectional`, and `interface/core.py` / `interface/misc.py`'s parameter threading. Traced the logic by hand through every scenario called out in the brief:

- **Capture starting mid-connection, no SYN ever seen**: buffer just accumulates under whatever `origin` the first-seen packet gives it; never superseded without a SYN, sits until `finish()`. No defect — matches the disclosed "undecided" case.
- **Retransmitted SYN** (either side): `_ended()` is false on a live buffer, so the supersede branch never fires; the retransmit just joins the buffer as an ordinary packet. Correct.
- **Simultaneous open / SYN arriving before any teardown recorded**: `_ended()` false → merged into the existing buffer rather than treated as a new flow. This is exactly the behaviour `pep.rst` discloses as deliberately undecided (no ACK flag to disambiguate a genuine reuse from a continuation).
- **RST followed by more packets of the same connection**: `reset=True` is sticky and `_ended()` goes true immediately, but nothing pops the buffer until a superseding SYN or `finish()` — later packets keep landing in the same buffer/forward/reverse lists as intended.
- **`finish()` called twice, or with nothing buffered**: it drains `self._buffer` with `list(self._buffer)`, so a second call (or a call on an empty tracer) is a no-op. Verified against `tests/foundation/traceflow/test_tcp.py:278-285`, which asserts this directly.
- **`make_bufid` canonicalisation**: verified algebraically that both directions of a conversation always reduce to the same tuple regardless of which endpoint is numerically smaller.
- **Long-lived-connection memory question**: a completed bidirectional flow now stays in `self._buffer` (not moved to `self._stream`) until superseded or `finish()`, rather than closing as soon as its own FIN went out. I could not turn this into a demonstrable regression: `dictdumper.Dumper.__call__` opens/writes/closes the output file on every call (verified by reading the installed `dictdumper` source), so no file descriptor is held open across the delay, and `self._stream` already accumulated every finished flow for the life of the object in the pre-PR code too, so total retained data is the same order either way. Flagging only as something to keep an eye on for very large captures with many long-lived, never-superseded connections, not as a defect.

**Finding 2 (IP `strict=False` contiguous prefix).** Read `pcapkit/foundation/reassembly/ip.py`'s new `submit()` branch and the `RCVBT` bit-table semantics. `TDL > 0` path is untouched (zero-filled buffer, as before). `TDL == -1` path walks `RCVBT` from the front and stops at the first clear bit, `stop = received * 8` — confirmed this returns `b''` when the very first 8-octet unit is missing (not an error), and correctly returns nothing past the first hole. `strict=True` branch's payload-building loop is untouched by the diff (only `completed` changed type); confirmed by diff inspection.

**Everywhere `Completion` flows**: read `reassembly/data/data.py`'s enum, `__bool__`/`__str__`, and every `submit()` in `ip.py`/`tcp.py` that constructs it. `TCP`'s `strict=False` path is explicitly documented as keeping the old zero-filled-whole-buffer payload, only `completed` changes — confirmed by diff, matches `follow_tcp_stream`'s use of it.

**Toolkit adapters**: confirmed all 6 engines that support TCP trace (dpkt, pcap, pcapng, pypcapfile, pyshark, scapy) now populate `Packet.rst`, reading the right flag bit/attribute for each engine's packet model. Noticed in passing (not a finding, a good catch already in this diff) that `toolkit/scapy.py`'s `tcp_traceflow` used to stamp `timestamp=time.time()` — the host clock — and now uses `float(packet.time)`, the capture's own clock, consistent with the rest of the PR's "capture's own clock" design.

## Claims independently reproduced

Checked out `a206459f3` in a clean tree, regenerated `examples/captures/` via `make_samples.py`, and ran with `pcapkit.__file__` printed and asserted to be under the tree being tested (`PYTHONSAFEPATH=1`, tree inserted at `sys.path[0]`):

- `pytest tests/ -q` → **917 passed, 17 skipped, 862 subtests passed, 0 failed** (550s). Matches exactly.
- Flow counts across the full 15-capture corpus, both modes:
  - default (bidirectional): **122 flows, 1222 traced frames** total; `http.pcap` **111 flows, 0 single-frame** (was 331/109 in `nobidir` mode below).
  - `trace_bidirectional=False`: **355 flows, 1222 traced frames** total — and this reproduces, file-for-file, an actual `git clone` + checkout of `origin/main` (`84233da3a`) run through the same script: 355/117-single/1222, identical per capture. Not just self-consistent with the PR's own claims — checked against real old code.
- Reassembly: **1479 datagrams, all `Completion.COMPLETE`**, across the same corpus. Matches.
- `mypy pcapkit --config-file mypy.ini`: **127 errors** on `a206459f3` vs **128** on `origin/main`. Diffed the two error lists line-by-line: the delta is exactly one removed error (`tcp.py`'s old `@overload`-based `Datagram.__init__` mismatch, which the PR deliberately removed) plus line-number shifts of pre-existing errors. Matches, and the mechanism checks out.
- pylint `cyclic-import` on the **same unchanged tree**, two consecutive runs: 130 then 117 lines. Confirms the count is non-deterministic run-to-run on identical code, consistent with the "28-message swing on an unchanged baseline" claim — spot-checked, not fully re-derived.

## Findings posted

One inline comment on `docs/source/pep.rst` (line 580, anchored on `a206459f3`, auto-remapped by GitHub to the current head `87099110f` since that file is untouched between the two): the "wiring the application layer into flow tracing" paragraph is pasted twice, verbatim, in the flow-tracing bullet (lines 563-572 and 580-589). It's a documentation duplication, not a code defect — the code matches the (de-duplicated) disclosure. Worth a cleanup pass before merge but not blocking.

## Not covered

Did not review: the `docs/source/*.rst` API-doc changes beyond `pep.rst` (mechanical, tracks the code); `tests/` file contents beyond spot-reading `test_tcp.py` (traceflow) and skimming `test_timeout.py`'s structure — I verified their *outcomes* by running the suite rather than auditing every assertion; the `RCVBT` marking arithmetic in `reassembly/ip.py`'s `reassembly()` method itself (start/stop-on-8-octet-boundary bit-setting) is pre-existing code untouched by this diff, so I didn't re-litigate it even though the new contiguous-prefix code depends on it being accurate.

## Verdict

No blocking defects found. One minor, non-blocking documentation issue posted inline (duplicate paragraph in `pep.rst`). CI is green at `a206459f3`. **Good to merge**, modulo the doc cleanup at the reviewer's discretion.

docs/source/pep.rst
JarryShaw: Confirmed and fixed in `9b3da9996`. The paragraph was byte-identical in both places, so this was a drafting artefact rather than two similar passages.

I dropped the **first** copy, which is the ordering you suggested and also the one that reads correctly: the "one case remains undecided" paragraph now follows directly from the finalisation explanation it qualifies — SYN versus SYN-ACK, and why the recorded teardown is what distinguishes them — and the "what is still wanted" paragraph closes the bullet as the forward-looking item. Previously the disclosure was sandwiched between two identical copies of an unrelated paragraph, which is exactly the reading problem you identified.

```
before:  563  What is still wanted here is **wiring the application layer …
         574  One case remains undecided rather than solved: …
         580  What is still wanted here is **wiring the application layer …   (duplicate)

after:   563  One case remains undecided rather than solved: …
         569  What is still wanted here is **wiring the application layer …
```

11 lines removed, docs only. `docutils` parses the page with zero messages and no nested inline markup.

Two notes on your review, since both were the right call. Anchoring to `a206459f3` while the head had moved to `87099110f` was correct — that merge touches only `ext.rst`, the HIP/IPv4 docs and code, and `registry/protocols.py`, nothing under `traceflow/` or `reassembly/` or this file, so nothing in your review was invalidated. And your independent re-derivation of the flow counts against a real `git clone` of `origin/main` matters more than it might look: **an earlier report of mine had those numbers wrong** — 355 → 234 with `http.pcap` at 220 — because it was measuring the defect, counting the 109 one-frame stray tails as real flows. Your 122 / 1222 with `http.pcap` at 111 and zero single-frame flows confirms the corrected figures from a second, independent direction.

The watch-item you raised rather than filed is also the right shape: completed bidirectional flows now stay in `self._buffer` until superseded or `finish()`, which sounds like a memory concern for a long capture, but `dictdumper.Dumper` opens, writes and closes per call and `self._stream` already retained everything for the object's lifetime before this PR. So it is a property of the design rather than a regression it introduces. Recording it here so it is findable if a long-capture memory question ever comes up.

pcapkit/foundation/reassembly/data/data.py
```diff
@@ -18,6 +19,58 @@
     from pcapkit.protocols.protocol import ProtocolBase as Protocol
 
 
+class Completion(enum.Enum):
```
JarryShaw: worth using StringEnum? from the compat module.

docs/source/pep.rst
```diff
@@ -529,16 +529,94 @@ Two smaller items in the same subsystem:
...
+  <pcapkit.foundation.traceflow.traceflow.TraceFlowBase.finish>`. Distinguishing
+  that SYN from the peer's SYN-ACK is what the recorded teardown is for.
+
+  One case remains undecided rather than solved: a capture that *starts* in the
+  middle of a connection, sees no teardown, and then has its endpoints reused. The
+  reuse is indistinguishable from a continuation without the ACK flag on
+  :class:`~pcapkit.foundation.traceflow.data.tcp.Packet`, which would make
+  ``syn and not ack`` a definitive new-connection test on its own.
+
+  What is still wanted here is **wiring the application layer into flow
```
JarryShaw: let's add this to the PR's scope with Deferred wired in.

pcapkit/foundation/reassembly/ip.py
```diff
@@ -87,6 +90,13 @@ def reassembly(self, info: 'Packet[_AT]') -> 'None':
         IHL = info.ihl      # Internet Header Length
         MF = info.mf        # More Fragments flag
         TL = info.tl        # Total Length
+        TS = info.timestamp  # Capture timestamp, i.e. the only clock we have
```
JarryShaw, line 93: `#` comment line alignment. needs to add a whitespace to all those lines above.

pcapkit/foundation/traceflow/tcp.py
```diff
@@ -27,17 +27,49 @@
 logger = get_logger(__name__)
 
 
-class TCP(TraceFlow[BufferID, Buffer, Index, Packet[_AT]], Generic[_AT]):
+class TCP(TraceFlow[BufferID, 'Buffer[_AT]', Index, Packet[_AT]], Generic[_AT]):
```
JarryShaw: is `Buffer[_AT]` or `'Buffer[_AT]'` here?

pcapkit/interface/misc.py
```diff
@@ -157,7 +165,21 @@ def follow_tcp_stream(fin: 'Optional[str]' = None, verbose: 'bool' = False,
         tcp_reassembly, pass_count = cast('ReassemblyAdapter', tk_pcapng.tcp_reassembly), False
     elif isinstance(exeng, DPKT_Engine):
         from pcapkit.toolkit import dpkt as tk_dpkt  # isort: skip # pylint: disable=import-outside-toplevel
-        tcp_reassembly = cast('ReassemblyAdapter', tk_dpkt.tcp_reassembly)
+        # NOTE: DPKT is the one adapter that cannot read a frame's capture
+        # timestamp off the frame -- DPKT hands ``(timestamp, bytes)`` back from
```
JarryShaw: wait, since the DPKT returns the timestamp, that's actually the capture timestamp we need.

JarryShaw: You were right, and the fix was one layer below where you commented. Done in `c4934656c`.

`DPKT.read_frame` was discarding the timestamp its reader hands back beside the octets — so the value was never unavailable, just thrown away, and `functools.partial(timestamp=0.0)` in `interface/misc.py` was papering over that. The engine now attaches it (`attach_timestamp` / `packet2timestamp` in `pcapkit.toolkit.dpkt`), and for a frame that never came through the engine it **raises** rather than defaulting — a zero there would silently misdate rather than fail. `grep` for `timestamp=0.0` in `misc.py` now returns **0 occurrences**, verified.

The NOTE that asserted DPKT "cannot read a frame's capture timestamp off the frame" is corrected, along with three adapter docstrings that repeated the same framing. That wording is what produced the wrong fix in the first place, so it mattered more than the code.

No behaviour change, checked rather than assumed: TCP reassembly has no timeout by default, so the value reaches no decision, and `follow_tcp_stream` on the DPKT engine now returns identical streams and conversations to the default engine — pinned by a new test. The timestamps attached to `in.pcap` are the capture's own (`1511106545.471719 …`, monotonic), not zeros.

docs/source/pep.rst
JarryShaw: In scope and done in `c4934656c`, with the design question answered rather than deferred.

**The tracer delegates to `pcapkit.foundation.reassembly.tcp.TCP` rather than growing a per-direction payload buffer.** The argument is that a buffer concatenating payloads in capture order is *silently* wrong on the first retransmission or reordered segment, whereas RFC 815's hole-descriptor algorithm already in the reassembler is not. That is pinned by a test rather than asserted — `test_analyse_gives_a_flow_its_application_layer_per_direction` delivers `seq=106` before `seq=101` and asserts `b'helloworld'` comes back, which the naive design would have returned as `b'worldhello'`.

`Packet` carries the four fields the reassembler needs — `seq`, `ack`, `header`, `payload` — read from exactly where each engine's sibling `tcp_reassembly` adapter already reads them. Each flow owns a reassembler, and `Index.packet` postpones **twice**: reading it flushes the flow's reassembler, and each datagram's own `packet` is parsed later still. On `http.pcap` the field holds a `Deferred` until read, then yields 2 datagrams, one per direction, both complete at 3703 and 361 bytes, whose payload analyses to an `HTTP` object with a real `HTTP/1.1 200 OK`.

**Opt-in, via `analyse=True`** (`trace_analyse=True` on `Extractor`, `extract()` and `follow_tcp_stream()`). Buffering every traced payload is a cost tracing does not otherwise pay, and your own #424 note is precisely about tracing's per-packet cost — so off by default nothing is buffered, reassembled or parsed, and flow counts are unchanged at 122 bidirectional / 355 unidirectional with 1222 traced frames either way. It is refused on `pyshark`, which reports dissected fields rather than octets — the same reason that engine has no reassembly adapter.

`pep.rst` no longer lists this as wanted.

pcapkit/foundation/reassembly/data/data.py
JarryShaw: Adopted in `c4934656c` — and the name is `StrEnum`, not `StringEnum`. It is house convention already: `httpv1.Type` and `pcapng.TLSKeyLabel` both use it.

I verified the truthiness contract myself on the branch rather than taking it on trust, because that contract is what makes this change safe:

```
COMPLETE   bool=True    ==True:False   ==False:False   json={"c": "complete"}
PARTIAL    bool=False   ==True:False   ==False:False   json={"c": "partial"}
TIMEOUT    bool=False   ==True:False   ==False:False   json={"c": "timeout"}
```

So `if datagram.completed:` and `assertTrue`/`assertFalse` behave exactly as they did when the field was a `bool`, while `== True` stays broken — which was already the documented break. The agent confirmed the same on both the stdlib and `aenum` paths.

It also **gains** something that matters here: `json.dumps` works. A plain `Enum` raises `TypeError`, and `to_dict()` hands this field straight out, so a dumped datagram would have failed. And `completed == 'timeout'` now reads naturally. The redundant `__str__` is gone since `StrEnum` provides it.

One cost, now carried as a `Warning` in the docstring: it is a non-empty string that tests false, so `bool(x)` and `bool(str(x))` disagree. Worth having written down where someone will find it.

pcapkit/foundation/traceflow/tcp.py
JarryShaw: Unquoted, in `c4934656c`. The line now reads:

```python
class TCP(TraceFlow[BufferID, Buffer[_AT], Index, Packet[_AT]], Generic[_AT]):
```

Both `Buffer` and `Packet` are imported at module scope from `pcapkit.foundation.traceflow.data.tcp`, so neither needed the quotes — the inconsistency was the whole of the problem. This also matches the sibling `reassembly.ip.IP`.

pcapkit/foundation/reassembly/ip.py
JarryShaw: Fixed in `c4934656c` — and in `reassembly/tcp.py` as well as `ip.py`, since the same misalignment had been introduced in both. Fixing only the file you happened to comment on would have left the asymmetry.

pcapkit/foundation/traceflow/tcp.py
```diff
@@ -19,27 +20,86 @@
...
+
+        Pass ``bidirectional=False`` for the older per-direction behaviour: a flow
+        is one direction, and closes on that direction's FIN. That mode reproduces
+        what flow tracing did before conversations became one flow, RST included --
+        which is to say it ignores RST, as it always did.
+
+    Note:
+        With ``analyse=True`` a flow also carries its **application layer**:
+        :attr:`Index.packet <pcapkit.foundation.traceflow.data.tcp.Index.packet>`
+        holds one reassembled datagram per direction, each with its payload parsed
```
JarryShaw: Reviewing at head `c4934656c`.

`Index.packet` does not hold "one reassembled datagram per direction" in the general case — it holds one datagram per **distinct ACK value** observed in that direction. `_make_segment` feeds each traced packet straight into `pcapkit.foundation.reassembly.tcp.TCP.reassembly()` unmodified, and that method buckets segments by `(BUFID, ACK)` (`pcapkit/foundation/reassembly/tcp.py`, the `if ACK not in self._buffer[BUFID].ack:` branch around line 168 — a new `Fragment` per distinct ACK number, never merged with a prior one under the same BUFID). That bucketing predates this PR and is fine for the reassembler's original job (one burst, one ACK), but it is now reachable through `analyse=True` for a whole multi-round-trip conversation, where it isn't fine.

For a single request/response where each side's ACK number never changes — the only shape `test_analyse_gives_a_flow_its_application_layer_per_direction` exercises — this is invisible: everything lands in one `Fragment`. But a real ACK number advances every time a side receives more from its peer, so any connection with more than one exchange per direction (HTTP keep-alive, RPC, chat — the normal case for "follow a TCP stream") produces one `Datagram` per exchange, not one per direction.

Reproduced against this exact tree (`PYTHONSAFEPATH=1`, `PYTHONPATH` at the checkout of `c4934656c`, `pcapkit.__file__` confirmed under it) with a 3-round-trip exchange on one connection:

```
client -> GET1 (ack=9001), GET2 (ack=9006), GET3 (ack=9011)
server -> RESP1 (ack=1005), RESP2 (ack=1009), RESP3 (ack=1013), FIN
```

`flow.packet` returns **6** `Datagram` entries, not 2 — `GET1`/`GET2`/`GET3` and `RESP1`/`RESP2`/`RESP3` each come back as separate, unmerged entries (grouped by direction, but not concatenated within it):

```
DatagramID(src=(...,80), ...) payload=b'RESP1'
DatagramID(src=(...,80), ...) payload=b'RESP2'
DatagramID(src=(...,80), ...) payload=b'RESP3'
DatagramID(src=(...,666), ...) payload=b'GET1'
DatagramID(src=(...,666), ...) payload=b'GET2'
DatagramID(src=(...,666), ...) payload=b'GET3'
```

A caller who takes this docstring at its word and reads `datagrams[0].payload` / `datagrams[1].payload` as "the two directions" gets an arbitrary slice of each side's data, not the conversation. This undercuts the stated reason for delegating to this reassembler instead of naive concatenation ("silently wrong on the first retransmission or reorder") — for any realistic bidirectional exchange, the caller still has to stitch multiple `Datagram` entries back together by hand, just along a different (ACK-keyed) axis instead of arrival order, and neither this docstring nor `docs/source/pcapkit/foundation/traceflow/tcp.rst` says so.

pcapkit/foundation/traceflow/tcp.py
```diff
@@ -158,7 +453,10 @@ def submit(self) -> 'tuple[Index, ...]':
         for buf in self._buffer.values():
             ret.append(Index(fpout=f"{self._fproot}/{buf.label}{self._fdpext}" if self._fdpext else None,
                              index=tuple(buf.index),
-                             label=buf.label,))
+                             label=buf.label,
+                             forward=tuple(buf.forward),
+                             reverse=tuple(buf.reverse),
+                             packet=None if buf.reassembly is None else Deferred(buf.reassembly),))
```
JarryShaw, line 521: Reviewing at head `c4934656c`.

`submit()` hands out `Deferred(buf.reassembly)` wrapping the **live, still-open** per-flow reassembler for every flow still in `self._buffer`. `DeferredPacket.__analyse__` (`pcapkit/foundation/traceflow/data/data.py`) caches the result of the first read into `self.__dict__`, so reading `.packet` on one of these mid-capture snapshots freezes whatever the reassembler happened to have produced *at that instant* — permanently, on that particular `Index` object — even though the underlying flow is still open and will go on to accumulate more data and eventually finalise with a fuller picture via `_finalise()`.

Verified against this tree at `c4934656c`:

1. Trace a flow, send one chunk of data (`'first-chunk'`).
2. Call `submit()` and read `.packet` on the returned snapshot → `(Datagram(payload=b'first-chunk'),)`.
3. Keep feeding the *same* flow more data (`'server-reply'`), then call `finish()`.
4. The final, `_finalise()`-produced `Index.packet` for that same flow correctly shows **both** datagrams.
5. Re-reading `.packet` on the **stale snapshot object** from step 2 still returns only `(Datagram(payload=b'first-chunk'),)` — forever, with nothing distinguishing it from a final, complete answer (no `PARTIAL`/`TIMEOUT` marker, since "holes" in the TCP reassembler's sense doesn't cover "more might still come" — see the neighbouring finding on `Completion`).

`submit()`'s own docstring only warns that such a flow's *callback* hasn't fired yet ("reported here but has not fired its callback"); it says nothing about `.packet` on one of these snapshots freezing a partial view the moment it is read, nor does the class-level `analyse=True` docstring above. Worth documenting explicitly: a caller polling `Extractor.trace` (or `follow_tcp_stream` with `trace_analyse=True`) mid-capture and caching the `Index` it gets back will silently keep a stale, partial answer rather than an up-to-date one for any flow that was still open at the time it read `.packet`.

pcapkit/foundation/reassembly/tcp.py
```diff
@@ -253,6 +285,12 @@ def submit(self, buf: 'Buffer', *, bufid: 'BufferID') -> 'list[Datagram]':  # ty
                 holes.append((max(start, 0), min(stop, length)))
             holes.sort()
 
+            # How completely this buffer came out, and why it stopped. Derived
+            # once per buffer, so the two branches cannot disagree about it.
+            completion = Completion.COMPLETE if not holes else (
```
JarryShaw, line 290: Reviewing at head `c4934656c`.

Not new to this PR, but directly relevant to its stated rationale for wiring flow tracing through this reassembler instead of a naive per-packet concatenation ("silently wrong on the first retransmission or reorder... while RFC 815's hole descriptors are not"): a genuinely **conflicting** retransmission — same sequence range, different payload bytes — is resolved by silent last-write-wins in the overlap-handling branch a little above this `submit()` (`RAW[PSN - ISN:PSN - ISN + info.len] = info.payload`), with no signal that two different byte sequences were seen for the same offsets, and the result is still reported `Completion.COMPLETE`.

Reproduced against this tree at `c4934656c`: segment carrying `b'AAAAA'` at seq 2001 arrives, then a second segment at the *same* seq 2001 (same ACK) carrying `b'ZZZZZ'` arrives. The reassembled `Datagram.payload` is `b'ZZZZZ'` — `AAAAA` is gone without a trace — and `completed` reads `Completion.COMPLETE`.

This is an inherent limitation of the underlying reassembler (there's no universally correct resolution for a genuine conflict, and real-world retransmissions normally do carry identical bytes), so I'm not asking for a fix. But the PR body doesn't mention this as a residual gap alongside its retransmission/reorder claim, and it's worth knowing before leaning on this integration for anything where the capture could be adversarial or corrupted (overlapping-segment ambiguity is a known TCP-reassembly evasion technique) — `Completion.COMPLETE` here means "no gaps," not "unambiguous."

JarryShaw:
## Review of PR #435 at head `c4934656c`

Acting as the automated reviewer (GitHub Copilot is out of tokens). This is a full review of the current head, not a delta from the two earlier verdicts (`a939c4f3b`, `a206459f3`) — item 3 below was added after those.

### CI

`gh pr view 435 --json statusCheckRollup` / `gh pr checks 435`, both re-run against head `c4934656c2804cc05b77fa58040eb6a60c8074fe`: every real job is `SUCCESS`/`pass` — `Analyze`, all `Python 3.10`-`3.15`, all `Compat Python 3.10`-`3.15`, all `Integration Python 3.10`-`3.15`, `CodeQL`, `deploy-pages`, `pyup.io/safety-ci`. `Docs test gate` and `Gate (full suite, Python 3.14)` show `SKIPPED`/`skipping`; checked `.github/workflows/unit-tests.yml` — both are gated on `inputs.gate-only == true`, which is a workflow_call input never set for a `pull_request` trigger, so skipping here is expected, not a masked failure. No red anywhere.

### What I checked

- Confirmed `origin/main` (`c28ffc287`) is fully contained in the PR branch (`git rev-list --count pr-435..origin/main` = 0; PR branch is 9 commits ahead) — read all "what main contains" diffs via `git show origin/main:<path>` / `git diff origin/main...pr-435`, never the working tree.
- Read the full diff (45 files, +2561/-240) and the PR body against it. The three claimed capabilities (RFC-1122 60s IPv4/IPv6 reassembly timeout on the capture's own clock, default-on bidirectional TCP flow tracing with teardown-recorded-not-finalizing semantics, and `Deferred` wired into flow tracing via delegation to `reassembly.tcp.TCP`) all match code that is actually present in the diff.
- Verified the `Completion` `StrEnum` contract by hand (`bool(COMPLETE)` True, `bool(PARTIAL)`/`bool(TIMEOUT)` False, `COMPLETE == True` is False, `json.dumps` round-trips, `== 'timeout'` works) — matches the PR body exactly.
- Read the DPKT `attach_timestamp`/`packet2timestamp` change (`pcapkit/foundation/engines/dpkt.py`, `pcapkit/toolkit/dpkt.py`) — raises `UnsupportedCall` rather than defaulting, as claimed, and `tcp_traceflow` now supplies `rst`/`seq`/`ack`/`header`/`payload`, consistent with what `analyse=True` needs.
- Read the IP `strict=False`, TDL-never-received change (`pcapkit/foundation/reassembly/ip.py`): confirmed it reports the contiguous prefix up to the first hole (`RCVBT`-derived), not `b''` and not the old `datagram[:-1]` 65534-octet slice.
- Spot-checked the disclosed pylint suppressions: both `DeferredPacket` mixins (`pcapkit/foundation/traceflow/data/data.py`, `pcapkit/foundation/reassembly/data/data.py`) carry the same `super()`-on-a-mixin rationale for the `E1101`/`no-member` suppressions, as described.
- Built and ran targeted probes against this exact checkout (`PYTHONSAFEPATH=1`, `PYTHONPATH` at this worktree, `pcapkit.__file__` printed and confirmed under it each time, interpreter `.venv/bin/python` 3.14) covering every scenario named for item 3: a flow with no payload at all (clean, empty tuple, no crash), a flow finalised by `finish()` with no FIN/RST ever seen (reports `Completion.COMPLETE` rather than `PARTIAL` — pre-existing reassembler semantics, see inline note), reading `Index.packet` twice (cached, same object, no double work), reading it after the tracer object is garbage-collected (no crash — each flow's `Deferred` owns its own reassembler independently), a retransmission with different payload at the same sequence (silent last-write-wins, still `COMPLETE`), `analyse=True` with `bidirectional=False` (works correctly, two independent unidirectional reassemblers), and a multi-round-trip bidirectional exchange (see finding below — this is the one that doesn't hold up).

### Findings (posted inline)

1. **`pcapkit/foundation/traceflow/tcp.py:84`** — `Index.packet` does not hold "one reassembled datagram per direction" in general; it holds one datagram per distinct ACK value seen in that direction, inherited from `reassembly.tcp.TCP`'s `(BUFID, ACK)`-keyed buffering. A 3-round-trip exchange on one connection comes back as 6 `Datagram` entries, not 2. The PR's own test for this (`test_analyse_gives_a_flow_its_application_layer_per_direction`) only ever uses a constant ACK per side, so it can't see this. This is the substantive one — it means the new capability doesn't deliver "one datagram per direction" for any realistic keep-alive/RPC/chat-style connection, only for a single request/response.
2. **`pcapkit/foundation/traceflow/tcp.py:459`** — `submit()`'s `Deferred` snapshots of a still-open flow freeze permanently on first read (via `DeferredPacket` caching), and can go stale relative to what the same flow's final, `finish()`-produced `Index.packet` later shows, with nothing marking the snapshot as partial. Documentation gap more than a bug.
3. **`pcapkit/foundation/reassembly/tcp.py:290`** — pre-existing reassembler behavior, newly load-bearing here: a conflicting retransmission (different bytes, same sequence range) is silently resolved last-write-wins and still reported `COMPLETE`. Informational, not asking for a fix.

Nothing else I probed broke, and everything in the "claims to test rather than trust" list that I directly checked (CI, `Completion` contract, DPKT timestamp plumbing, IP prefix-to-first-hole) held up. I did not re-run the full 930-test suite or regenerate the 15-capture corpus myself (each is a multi-minute job and CI already ran the equivalent matrix green); I relied on CI for those broad claims and spent the review budget on hands-on probing of item 3, per the brief.

### Coverage

Covered in depth: `pcapkit/foundation/traceflow/tcp.py`, `pcapkit/foundation/traceflow/data/{data,tcp}.py`, `pcapkit/foundation/reassembly/tcp.py`, `pcapkit/foundation/reassembly/reassembly.py`, `pcapkit/foundation/reassembly/ip.py`, `pcapkit/foundation/reassembly/data/data.py`, `pcapkit/foundation/engines/dpkt.py`, `pcapkit/toolkit/dpkt.py`, `pcapkit/interface/misc.py`.

Read but not exercised beyond the diff itself: `pcapkit/foundation/reassembly/ipv4.py`, `ipv6.py`, `pcapkit/foundation/traceflow/traceflow.py`, `pcapkit/interface/core.py`, `pcapkit/toolkit/{pcap,pcapng,pypcapfile,pyshark,scapy}.py`, and the `docs/source/**` and `tests/**` changes (read for consistency with the code, not independently re-run beyond what CI already ran).

### Verdict

Given CI is fully green and findings 2 and 3 are minor/informational, this is not a blocker from me, but finding 1 is a real correctness gap in the flagship new capability's documented contract ("one reassembled datagram per direction") for any connection with more than one exchange per side — worth a decision (fix the buffering, or narrow the documented guarantee) before leaning on `analyse=True` for anything beyond a single request/response. I'd call this **good to merge only if finding 1 is either accepted as a known, documented limitation or addressed** — it's not a crash or data-loss bug, just a capability that's narrower than advertised.

docs/source/pcapkit/foundation/traceflow/tcp.rst
```diff
@@ -105,6 +165,14 @@ Data Structures
    :members:
    :show-inheritance:
 
+.. autoclass:: pcapkit.foundation.traceflow.data.data.Deferred
```
JarryShaw: follow the module and current model declaration in reassembly as well.

pcapkit/foundation/traceflow/tcp.py
JarryShaw: Confirmed, and reproduced independently — the count is **6**, matching yours exactly. Fixed in `3b750a662`, as documentation plus a test rather than a behaviour change; the reasoning for that is below.

Three request/response round trips on one connection, with the acknowledgement number advancing as it does on any real exchange:

```
datagrams: 6  (documented: "one per direction" = 2)
  src=12345  ack=501   payload=b'req1'   frames=(3, 4)
  src=12345  ack=506   payload=b'req2'   frames=(6,)
  src=12345  ack=511   payload=b'req3'   frames=(8,)
  src=  443  ack=105   payload=b'resp1'  frames=(5,)
  src=  443  ack=109   payload=b'resp2'  frames=(7,)
  src=  443  ack=113   payload=b'resp3'  frames=(9,)
```

Mechanism is as you describe: `reassembly/tcp.py` buckets as `self._buffer[BUFID].ack[ACK]` (:168, :183, :207) and `submit()` iterates `for (ack, buffer) in buf.ack.items()` (:269), emitting one `Datagram` per bucket, while `_make_segment` passes `ack=packet.ack` through untouched. And your point about the existing test is exactly right — `_packet` defaults `ack: int = 0` and **every** existing test leaves it there, so the suite could only ever see the collapsed case.

### Why the docs changed rather than the code

The alternative fix is to feed the reassembler a constant `ack`, which would give literally one datagram per direction. I did not do that, because it makes `analyse=True` worse:

the acknowledgement number advances exactly when the peer has spoken, so bucketing on it splits each direction at its **message boundaries**, and every datagram's payload is one application message that `Datagram.packet` can parse on its own. Merge a direction into one stream and the application parser is handed `b'req1req2req3'` and reads only the first message — which is the failure mode for exactly the case this feature exists to serve, HTTP keep-alive. So the behaviour is the useful one and the sentence promising "one per direction" was the defect.

The class docstring now states the rule, shows the six-datagram table above, and says why the bucketing is deliberate. The sibling test's assertion message no longer states the general rule, and says instead that one acknowledgement number each way is *why* it gets one datagram each way, cross-referencing the new test. The new test drives all three round trips and pins all six by `(source port, acknowledgement number)`, plus asserts that the payload-free handshake and bare ACK raise no bucket of their own.

Full suite on 3.14.7 after the change: 932 passed, 17 skipped (930 before, plus the two tests added for this and the snapshot finding).

pcapkit/foundation/traceflow/tcp.py
JarryShaw: Confirmed and documented in `3b750a662`, with a test, since the claim is easy to state wrongly.

Measured — read the flow's `packet` two frames in, then let the third arrive:

```
mid-capture read : [b'first']
same Index again : [b'first']     <- frozen (identical object)
after finish()   : [b'firstsecond']
```

So the snapshot is taken at **first read**, not at `submit()`, and is then fixed in place; a caller holding that `Index` cannot see later segments and nothing on it distinguishes it from a final result.

One correction to the framing, which is why I did not treat this as a caching bug: `submit()`'s own `__cached__['submit']` is *not* what freezes it. Both `trace()` (:195) and `finish()` (:431) clear that key, so a later `submit()` rebuilds the tuple with fresh `Deferred` objects — which is exactly why the `after finish()` line above sees `b'firstsecond'`. What cannot catch up is the *previously returned* `Index`, if a caller kept one. The docstring now says both halves, so the note is not misread as a stale-cache defect later.

The test asserts the frozen object is identical across the intervening `trace()` (`assertIs`), which is what makes it a real pin rather than a coincidence of equal payloads.

pcapkit/foundation/reassembly/tcp.py
JarryShaw: Confirmed as pre-existing and filed as **#443** rather than folded in here, since this PR does not touch the path.

Reproduced against `origin/main` at `c28ffc287` — two segments claiming the same range with different bytes, then the stream continuing:

```
r(seg(1, 100, b'AAAAAAAA'))    # original
r(seg(2, 100, b'BBBBBBBB'))    # same range, different bytes
r(seg(3, 108, b'CCCC'))
-> completed=True  payload=b'BBBBBBBBCCCC'  frames=(1, 2, 3)
```

`b'AAAAAAAA'` is gone without a trace and `completed` is `True`. Mechanism is the unconditional slice assignment at `reassembly/tcp.py:166` on main (`RAW[PSN - ISN:PSN - ISN + info.len] = info.payload`), mirrored at `:176` for a segment reaching back before the ISN; the hole list is updated from the segment's own bounds, so the contested range stays hole-free and `completion` derives as `COMPLETE`.

That this PR does not touch it: `git diff origin/main -- pcapkit/foundation/reassembly/tcp.py` shows no change to either overlap branch, and both lines are byte-identical on `main`.

Worth its own issue rather than a note, because the interesting case is not a buggy sender — it is deliberately overlapping segments, the classic TCP-overlap IDS-evasion technique, whose whole premise is that different reassemblers resolve the conflict differently. Last-write-wins is a defensible choice; making it silently while still reporting completeness is what leaves a caller unable to tell a clean stream from a contested one. #443 lays out three possible directions without picking one, since whether the reassembler should grow a notion of "contested" is a design call.

docs/source/pcapkit/foundation/traceflow/tcp.rst
JarryShaw: Done in `3b750a662` — `Deferred` and `DeferredPacket` moved out of `traceflow/tcp.rst` into `traceflow/index.rst`, directly after `TraceFlowData`, which is where `reassembly/index.rst` declares its equivalents (`ReassemblyData`, then `Deferred`, then `DeferredPacket`, under the same `.. module:: …data` / `.. module:: …data.data` / `.. currentmodule::` trio).

Two things were wrong with the old placement, not just one: these classes live in `pcapkit/foundation/traceflow/data/data.py`, so declaring them in `tcp.rst` put them both in the wrong file *and* under that file's `.. module:: pcapkit.foundation.traceflow.data.tcp` directive — the module the surrounding `Packet`/`Buffer`/`Index` belong to and they do not. `index.rst` already had the matching `data` / `data.data` module declarations in place, so nothing new was needed there beyond the two blocks.

pcapkit/foundation/traceflow/tcp.py
```diff
@@ -19,27 +20,113 @@
...
+        way, one per exchange:
+
+        .. code-block:: text
+
+           src=12345  ack=501  payload=b'req1'    src=443  ack=105  payload=b'resp1'
+           src=12345  ack=506  payload=b'req2'    src=443  ack=109  payload=b'resp2'
+           src=12345  ack=511  payload=b'req3'    src=443  ack=113  payload=b'resp3'
+
+        That is deliberate rather than incidental: the acknowledgement number
+        advances exactly when the peer has spoken, so bucketing on it splits a
```
JarryShaw, line 106: Mechanism and count both check out. Read `pcapkit/foundation/reassembly/tcp.py` and `reassembly/reassembly.py` at `3b750a662` and traced the bucketing by hand: `_buffer[BUFID].ack[ACK]` buckets per (BUFID, ACK), `submit()` there emits one `Datagram` per non-empty bucket, and `_make_segment` here passes `ack=packet.ack` through untouched — nothing renumbers it. Ran `test_each_exchange_of_a_multi_round_trip_flow_is_its_own_datagram` directly (`PYTHONSAFEPATH=1`, `PYTHONPATH` pointed at the worktree) and it passes; walking its nine-packet sequence through the bucketing logic by hand gives the same six (src, ack) pairs the test and the docstring table both claim, so this isn't a test that would pass vacuously.

One phrase overstates the generality, though: "the acknowledgement number advances exactly when the peer has spoken, so bucketing on it splits a conversation at its message boundaries" (lines 105-107) holds for strict lock-step request/response — which is what the new test drives — but not in general. The ACK field records how much of the *peer's* stream has been acknowledged, not how many messages the local side has sent. A direction that sends two application messages back-to-back before the peer replies (HTTP/1.1 pipelining, or any protocol that doesn't wait for a turn) keeps the same ACK across both sends, so both land in one bucket and get concatenated — exactly the failure mode this paragraph says merging directions would cause, just one level down. Not a code defect, and not something this PR needs to solve; just a caveat the "splits at message boundaries" framing doesn't cover, worth a clause if this docstring gets touched again.

JarryShaw:
## Review of PR #435 at head `3b750a662d5e36f617d03184e12371ec36aafe53`

Standing in for Copilot (out of tokens). This is a delta review, not a fresh full-diff pass: the last verdict was at `c4934656c` and raised three findings, all of which are answered by the only new content since then, commit `908913106` ("traceflow: state what analyse=True actually returns, and pin it") — confirmed via `git show 908913106 --no-patch --format='%P'` (parent `c4934656c`) and `git show 3b750a662 --no-patch --format='%P'` (parents `908913106` and `e4b6dd86b`, the latter a pass-through merge of `origin/main` that brought in PR #436 and touches nothing under `traceflow/`, `reassembly/`, or this docstring — confirmed by inspecting its diff separately). `git rev-list --count origin/main..3b750a662` = 12; `git rev-list --count 3b750a662..origin/main` = 0. `origin/main` = `f50436a8a1f4cfd9e2f9dc96080fa9f4d8ca08a4`, matching the brief. Fetched `refs/pull/435/head` and confirmed `FETCH_HEAD` = `3b750a662d5e36f617d03184e12371ec36aafe53` exactly; `gh pr view 435 --json headRefOid` agrees.

### CI

`gh pr view 435 --json statusCheckRollup` — waited for it to settle (dispatch-time snapshot had 11 jobs still `IN_PROGRESS`), then re-polled to completion. Final state, 24 checks total: 21 `SUCCESS` (`Analyze`/CodeQL, `Compat Python 3.10`-`3.15`, `Python 3.10`-`3.15`, `Integration Python 3.10`-`3.15`, `deploy-pages`), 2 `SKIPPED` (`Docs test gate`, `Gate (full suite, Python 3.14)` — both gated on a `workflow_call` input never set for a `pull_request` trigger, confirmed against `.github/workflows/unit-tests.yml` in the prior verdict and unchanged since), plus `pyup.io/safety-ci` `SUCCESS`. No red anywhere.

### What I checked, and what I ran

All three claims in `908913106`'s commit message, against the code at `3b750a662` (never the ambient working tree for reading — used `git show <ref>:<path>`; checked out the exact PR sha in this dedicated worktree, verified `HEAD` == `3b750a662d5e36f617d03184e12371ec36aafe53`, only to *run* tests):

**1. "One datagram per direction per ACK, not per direction."** Read `pcapkit/foundation/reassembly/tcp.py` (`reassembly()` and `submit()`) and `reassembly/reassembly.py` (`datagram`/`fetch()`), plus `traceflow/tcp.py`'s `_make_segment`. Confirmed the mechanism exactly as described: `self._buffer[BUFID].ack[ACK]` is the bucket key, `submit()` in the reassembler emits one `Datagram` per non-empty bucket, and `_make_segment` passes `ack=packet.ack` straight through with no arithmetic (unlike `dsn`, which does get the SYN +1 offset — that adjustment happens on the receiving side in `reassembly()`, not here). Ran the new test `test_each_exchange_of_a_multi_round_trip_flow_is_its_own_datagram` directly:
```
PYTHONSAFEPATH=1 PYTHONPATH=<worktree> .venv/bin/python -m unittest tests.foundation.traceflow.test_tcp.TCPTraceFlowTests.test_each_exchange_of_a_multi_round_trip_flow_is_its_own_datagram -v
```
→ `ok`. Also walked its nine-packet sequence through the bucketing logic by hand (not just trusting the assertion): the SYN, SYN-ACK and the bare ACK at index 3 land in empty-payload buckets that get stripped (`if payload:` in `submit()`), and the six non-empty buckets are exactly `{(12345,501):req1, (12345,506):req2, (12345,511):req3, (443,105):resp1, (443,109):resp2, (443,113):resp3}` — which matches both the test's own `got` dict and the docstring's six-row table verbatim. Six is right, and the table is accurate. The test is not vacuous: it asserts the dict equality and two `assertNotIn`s, not just a length.

Left one inline comment on `traceflow/tcp.py:106` on a real but minor overclaim: "the acknowledgement number advances exactly when the peer has spoken, so bucketing on it splits a conversation at its message boundaries" is true for the lock-step request/response the test drives, but the ACK field tracks how much of the *peer's* data has been acknowledged, not how many messages the local side sent — a direction that pipelines two messages before the peer replies keeps one ACK across both and gets them concatenated into one bucket, same as the "merging would concatenate" failure this paragraph warns about. Not a defect, not blocking, just a caveat the phrasing doesn't carry.

**2. Docs-not-code was the right call.** Given (1) is correct behaviour and not a bug, rewriting the reassembler to merge across ACK buckets would be the change that introduces a real defect: `req1`+`req2`+`req3` would concatenate into `b'req1req2req3'`, and `Datagram.packet`'s on-demand `analyze()` would parse only the head of that blob as a single message, silently dropping the rest — which is precisely the failure mode the commit message says merging causes. The existing per-ACK-bucket granularity is what makes each datagram one parseable unit for the lock-step case, and RFC 815 hole-descriptor buffering is already keyed that way for a reason unrelated to this PR. So: sound call, and it's not a case where I'd want the code changed instead — the mechanism was already right, only the doc claim was wrong.

**3. Open-flow snapshot freezing.** Read `traceflow/tcp.py`'s `submit()`/`trace()`/`finish()`, `traceflow/data/data.py`'s `Deferred`/`DeferredPacket`, and `reassembly/reassembly.py`'s `datagram`/`fetch()` properties. Confirmed both halves: `DeferredPacket.__analyse__` resolves a `Deferred` on first read and overwrites `self.__dict__[key]` with the concrete tuple *on that specific `Index` instance* — a mechanism entirely separate from `TraceFlow.__cached__['submit']`. Both `trace()` (`self.__cached__['submit'] = None` at the top) and `finish()` (same, at the end) do clear that cache, exactly as claimed — but clearing it only changes what a *subsequent* call to `submit()` builds (a fresh `Index` with a fresh `Deferred`); it does nothing to an `Index` object a caller is already holding, since that object's `.packet` was already overwritten with a plain tuple and no longer contains a `Deferred` to re-resolve. Ran the new test:
```
PYTHONSAFEPATH=1 PYTHONPATH=<worktree> .venv/bin/python -m unittest tests.foundation.traceflow.test_tcp.TCPTraceFlowTests.test_an_open_flows_application_layer_is_a_snapshot_of_when_it_was_read -v
```
→ `ok`. The `assertIs(held.packet, resolved, ...)` is meaningful, not coincidental — traced why identity holds: once `__analyse__` replaces the dict entry with the resolved tuple, later reads take the `not isinstance(value, Deferred)` branch and return the same object unchanged. Non-vacuous: the test also proves the *fresh* `Index` from `finish()` sees the new segment (`[b'firstsecond']`), so it isn't merely asserting "nothing ever changes."

**Amended assertion message.** `'one acknowledgement number each way, so one datagram each way'` — true: every segment in that older test carries the default `ack=0`, which is the one shape that collapses to "one per direction." Confirmed by re-running it (below) and by the same by-hand bucket trace.

**Docs move.** `Deferred`/`DeferredPacket` moved from `traceflow/tcp.rst` (where they sat under the `.. module:: pcapkit.foundation.traceflow.data.tcp` directive, though they live in `data/data.py`, not `data/tcp.py`) into `traceflow/index.rst`, immediately after `TraceFlowData`, under `.. module:: ...traceflow.data` / `.. module:: ...traceflow.data.data` / `.. currentmodule:: ...traceflow` — grepped `docs/source/pcapkit/foundation/reassembly/index.rst` and confirmed this is exactly its existing convention for `ReassemblyData`/`Deferred`/`DeferredPacket`. Grepped the whole `docs/` tree for `traceflow.data.data.Deferred(Packet)?` and each name now appears exactly once, only in `index.rst`, so nothing is doubly declared. Cross-references in `tcp.py`'s docstrings use the fully-qualified dotted path (`:class:`~pcapkit.foundation.traceflow.data.data.Deferred``), which Sphinx resolves independent of which `.rst` file carries the `autoclass`, so nothing breaks by the move.

**Test run.** All 14 tests in `tests/foundation/traceflow/test_tcp.py` pass (`Ran 14 tests ... OK`). I also ran the **full suite** (`python -m unittest discover -s tests`), which requires the generated sample corpus (`examples/generators/make_samples.py`) — my first pass, before generating it, threw 130 `FileNotFoundError`s and 3 unrelated `engine_parity` failures, none of which are attributable to this commit (all in `protocols`/`integration` runtime tests that read `examples/captures/*.pcap`). After generating the corpus, the full suite is clean: `Ran 851 tests in 454.6s — OK (skipped=13)`. I could not reproduce the commit message's exact "932 passed, 17 skipped" via bare `unittest discover` (I get 838 passed + 13 skipped = 851 total, not 949) — the difference is almost certainly methodology (the PR's own review history shows earlier passes measured via `pytest` with a subtests-counting plugin, which reports subTest instances as additional discrete counts that `unittest discover`'s tally doesn't split out the same way), not a real regression: zero failures, zero errors, in an environment proven to be this tree (`pcapkit.__file__` printed and asserted to be under the worktree before trusting any of the above).

### Verdict

**GOOD TO MERGE** at `3b750a662d5e36f617d03184e12371ec36aafe53`. All three prior findings are correctly and durably fixed by `908913106` — the mechanism claim, the six-datagram count, the docstring table, and the snapshot-freezing explanation are all accurate and independently verified against the actual code paths, not just against the new tests' own assertions. The docs-vs-code call is sound with evidence either way. The two new tests are non-vacuous and the amended assertion message is true. The docs move follows the established `reassembly/index.rst` convention with no duplicate declarations or broken cross-references. CI is fully green (21 `SUCCESS`, 2 legitimately `SKIPPED`, `pyup.io/safety-ci` `SUCCESS`). One non-blocking nit left inline on `traceflow/tcp.py:106` about the "splits at message boundaries" phrasing overstating its own generality (pipelined sends within one ACK bucket still concatenate) — worth a clause next time this docstring is touched, not worth a revision on its own.

pcapkit/foundation/traceflow/tcp.py
JarryShaw: Right on both counts, and fixed in `1c8d975c4` rather than deferred — the caveat turned out to be cheap to state and cheaper to pin.

Reproduced before rewording, on the PR tree with `pcapkit.__file__` confirmed. Same nine-packet shape as the multi-round-trip test, except the client sends twice before the server replies, so both requests carry `ack=501`:

```
datagrams: 2
  src=12345 ack=501   payload=b'req1req2'
  src=  443 ack=109   payload=b'resp1'
```

So the two requests are one datagram, exactly as you said — and a parser handed it reads only `req1`. Your diagnosis of why is precise: the ACK records how much of the *peer's* stream has been acknowledged, not how many messages the local side has sent, so it moves on the peer's turn and not on a message boundary.

What changed:

- The docstring now says the boundary is **the peer's turn** — one datagram per *exchange* — and then says plainly that this **narrows** the concatenation to within one exchange rather than eliminating it, with the measured `b'req1req2'` quoted and a pointer to the new test. The old phrasing claimed the two coincided in general; they coincide only in lock-step request/response.
- `test_pipelined_sends_share_a_datagram_because_the_ack_never_moved` pins it. A limitation that lives only in prose is one the next reader rediscovers — and this docstring is now the second thing on this PR that was wrong in prose while the code was right, so pinning it seemed the honest response rather than adding another clause.

One thing I want to be explicit about, because it would be easy to read this correction as undermining the choice it justifies: the reason for documenting the behaviour instead of changing it **survives**. Merging a whole direction concatenates every message it ever sent; ACK bucketing concatenates only within an exchange. Strictly better, just not the absolute property I claimed. My original wording overstated the benefit, and I have not weakened the case for which option to take.

`tests/foundation/traceflow`: 18 passed. Resolving — this pushes a new commit, so the PR owes a fresh verdict at `1c8d975c4`; the one you gave at `3b750a662` covered everything before it.

On the suite-count difference you flagged honestly: `unittest discover` giving 851/13 against `pytest`'s 932/17 is a collection difference, not a regression. `pytest` collects test *methods* the bare loader groups differently and additionally reports `subTest` outcomes separately (1268 of them on a sibling branch), so the two counts are not comparable — and both were clean. Good practice to flag it rather than quietly pick the friendlier number.

Issue #438, opened 2026-09-17
A wire-derived length underflows into a struct format string, raising bare struct.error on untrusted input


JarryShaw:
A wire-derived length field is used in arithmetic that can go negative, and the negative value reaches a `struct` format string, so parsing untrusted input raises a bare `struct.error` instead of an in-library exception.

## Reproduction

On `main` (`44aa38ae8`):

```python
from pcapkit.protocols.internet.hopopt import HOPOPT

buf = bytes.fromhex('3b00' '0800' '0000' '00' '00')   # SMF_DPD, I-DPD, Opt Data Len = 0
HOPOPT(buf, len(buf))
```

```
struct.error: bad char in struct format
```

The `-1` is being formatted into a struct template as `'-1s'`.

## Mechanism

`pcapkit/protocols/schema/internet/hopopt.py:426`, in `SMFIdentificationBasedDPDOption`:

```python
id: 'bytes' = BytesField(length=lambda pkt: pkt['len'] - (
    ...
))
```

`pkt['len']` is `Opt Data Len`, straight off the wire, and the subtraction underflows when a peer declares `0`. Nothing clamps it, so the negative width propagates into the field's `struct` template.

## Why it matters

`pcapkit` parses untrusted input, and `struct.error` is not one of the library's own exception types — a caller cannot catch it through `pcapkit.utilities.exceptions`, and it does not carry the field or protocol context an in-library error would. This is the class of defect the project has been closing elsewhere: it is the same shape as the `FieldValueError` guards added in #432 for non-progress, and the same shape as the unbounded `FieldBase.unpack` allocation noted in #432's body.

## Related sites worth checking in the same pass

Three sibling lambdas in `pcapkit/protocols/schema/internet/hip.py` — lines **699**, **717** and **735** — all compute `pkt['len'] - 1` from a wire field with no floor. `hopopt.py:361`, `:614` and `ipv6_opts.py`'s counterparts subtract larger constants (`pkt['len'] - 8 - pkt['cmpt_len'] * 4`, `pkt['len'] - 2 - …`) and so have more headroom to underflow, not less.

There was an earlier unverified report that `_option_padding` could go negative and reach `struct.calcsize('-1s')`. This is very likely that defect, now with a reproduction — worth treating them as one item.

## Provenance

[...]

Issue #439, opened 2026-09-17
Schema subclasses share one ABCMeta cache on Python <=3.10, so isinstance/issubclass return whichever answer was asked first


JarryShaw:
On Python ≤ 3.10, every `Schema` subclass **shares `Schema`'s `_abc_impl`** instead of getting its own. One `ABCMeta` subclass-check cache therefore serves the entire schema family, and because it is keyed only on the class being *tested* and not on the class it is tested *against*, the first answer for a given class is cached and returned for every later question about it.

The result: **`issubclass` and `isinstance` against schema classes return whichever answer was asked for first.**

## Reproduction

Two fresh 3.10 interpreters, same tree, differing only in the order of the two questions:

```python
import sys; sys.path.insert(0, '<tree>')
from pcapkit.protocols.schema.schema import Schema
from pcapkit.protocols.schema.internet.mh import (
    ANIGeoLocationSuboption as G, ANINetworkIdentifierSuboption as N)
```

| order asked | first answer | second answer |
|---|---|---|
| `issubclass(G, Schema)` then `issubclass(G, N)` | `True` ✔ | **`True`** ✘ (siblings) |
| `issubclass(G, N)` then `issubclass(G, Schema)` | `False` ✔ | **`False`** ✘ (it is a `Schema`) |

```
Python 3.10.20:  G._abc_impl is Schema._abc_impl  ->  True
Python 3.14.7 :  G._abc_impl is Schema._abc_impl  ->  False   (and both answers correct)
```

The first question is always right; the second always inherits the first's answer. No instances are involved, so this is not about `__eq__`, `__hash__` or value equality — it is a shared cache *object*. `SchemaMeta` derives from `abc.ABCMeta` (`pcapkit/protocols/schema/schema.py:136`) and every schema descends from `collections.abc.Mapping`.

Verified on 3.10.20 and 3.14.7. CPython reworked the `_abc` caches across the 3.11/3.12 line, which is why 3.14 is unaffected — but **3.10 is in the supported matrix** (`pyproject.toml` claims `>=3.6`, and CI runs 3.10 through 3.15).

## Why this is worth fixing rather than avoiding

[...]

PR #440, opened 2026-09-17
tests: round-trip 258 option codes from the registries, recording the 86 that cannot close the cycle

JarryShaw:
Closes the **Test Cases** ask on the Help Wanted page, which says what remains wanted is "**coverage rather than infrastructure**: the protocols and options the suite does not touch".

This adds that coverage for the option/chunk/parameter/block space, in both directions, and the result is the point of the PR: **of 258 registry codes exercised, 172 close a construct → parse → construct cycle and 86 do not.** Every one of the 86 is recorded in `tests/protocols/test_option_roundtrip_unit.py` against the `file:line` that stops it, and asserted to **still fail in the recorded way** — so fixing one turns the tier red rather than leaving a stale entry behind.

`pcapkit/` is untouched: `git diff origin/main -- pcapkit/` is empty. This is tests and generators only.

## How the codes are enumerated

`examples/generators/options.py` reads the **registries themselves** rather than a hand-written list: TCP options 22, MPTCP subtypes 8, IPv4 options 14, HOPOPT 15, IPv6-Opts 15, IPv6-Route 3, SCTP chunks 13 / parameters 8 / causes 13, MH messages 14 / options 20 / extensions 1, HIP parameters 49, HTTP/2 frames 10, PCAP-NG blocks 11 / options 40 / records 3 / secrets 4.

Registries are only ever *iterated*, which mutates nothing, and single-code resolution goes through `ProtocolBase._lookup_registry`; a test asserts registry sizes are unchanged after reading. IPv4 and HIP have no registry yet, so those enumerate enum × handler presence, and prefer a real `__option__`/`__parameter__` if one appears — so #434 landing is a no-op here.

166 cases are written into five `options-*.pcap` fixtures via `make_samples.py`, regenerating byte-identically across three passes. Fixtures deliberately carry only cases whose **parse terminates** — a capture containing the SMF_DPD non-progress frame would wedge every reader of it.

## The seven known construction defects: all reached, and two are worse than recorded

Each was expected and is asserted, not fixed:

1. **`ipv4.py` `data=` where the field is `ts_data`** — the call opens at :1480 and the bad keyword is at **:1488**. `Schema.__init__` only *warns*, so the value is dropped and `post_process` then iterates a `ListField`.
2. **`_make_mptcp_*` — all 8 subtypes fail**, from three distinct causes: `KeyError: 'length'` (`schema/transport/tcp.py:658/790/807/827`), `no attribute 'kind'` (`tcp.py:668`, three subtypes), and `_make_mptcp_join` reading `self._flags`, which exists only while parsing (`tcp.py:2675`).
3. **`hip.py:3445`** — `cipher=` is not a field of `EncryptedParameter`, so the cipher id is silently dropped.
4. **`CGAParameter.extensions` `KeyError: 'length'`** (`schema/internet/mh.py:516`) — it needs `pkt['__packet__']['length']` on the unpack path. The `MPTCPUnknown` half of that item is the same `KeyError` at the four `tcp.py` lines above.
5. **The Quick-Start `SchemaField(length=5)` is silent corruption, not a failure.** A well-formed 8-octet option `1908002adeadbee0` parses "successfully" with `SchemaWarning: packet length < 0: -3`, decoding `nonce` as **55** instead of **933982136**, and the three unconsumed octets are then read as a fabricated extra option, making the enclosing packet unparseable. Verified independently: parsing `QuickStartRequestOption` *directly* with all 8 octets gives the correct `nonce=933982136` with no warning, so the schema is right and the **selector's hardcoded length** is the defect. Identical at `hopopt.py:224` and `ipv6_opts.py:224`. **Caveat**: the QS cases never reach it, because they fail earlier at CONSTRUCT on a separate defect — `func` is set only by `post_process` — so fixing `length=5` alone will not make them pass.
6. **`_MPTCP.test`'s `'length': (1, 8)`** — window `1e0c01` decodes `length=60`; with the intended `(8, 8)` it decodes `12`. It feeds `SchemaField(length=pkt['test']['length'])`, so it asks for 60 octets of a 12-octet option, and **a genuine MP_CAPABLE option cannot be parsed at all**.
7. **Negative `_option_padding`** — mechanism verified (`while length > 0`, then `length -= len(data)`, then `_option_padding = length`, reaching `struct.calcsize('-1s')` → `error: bad char in struct format`) but **not reachable from the existing corpus**: an instrumented sweep of all 15 captures saw only `{0: 2651, 1: 3, 52: 1}`. It is latent. This is the same family as #438. The related non-progress *hang* is reached — two SMF_DPD cases TIMEOUT, which is why the sweep is deadlined.

## Larger finds beyond that list

- **PCAP-NG `__post_init__` packs and re-parses on one instance sharing `self._opt`**, making **27 of 40** block options unconstructible.
- **HIP cannot be parsed without `extension=True`** — `alias` reads `_info` before it is assigned — and its `len = total // 8 + 4` cannot represent a single parameter, so the harness uses two copies and pins the single-parameter defect in its own test.
- **16 HIP parameters return tuples where `_make_*` needs lists** — reconstruct-only failures, exactly the class the third leg of the cycle exists to catch.
- **6 of 10 HTTP/2 frames** raise `KeyError: 'flags'`, and PRIORITY checks `length != 9` for a frame `make` always builds as 14.

## Scope deliberately narrowed

**PCAP-NG writes no capture.** All 58 codes are exercised and recorded, but the construction API cannot produce a reproducible file: `_make_block_shb` takes byte order from `sys.byteorder`, and the TLS/WireGuard key-log writers stamp `datetime.now()` with no override. `examples/generators/pcapng.py` already covers that space by hand-rolling octets, which is presumably why.

## Verification

Suite at the current head `0e6a9e15c`: **926 passed / 17 skipped / 1268 subtests** (567 s) on 3.14.7, zero failures. That is 943 collected, confirmed by counting collection directly.

This figure was **886 / 17** when the PR was opened and is corrected here rather than silently: 886 + 17 = 903 collected, which matches no sha now on the branch, because two pass-through merges of `main` have landed since. Collection is 926 at both `d7240dc0c` and `4535f72f2` and 943 at `651665214` onwards, the +17 being #436's tests arriving — `tests/protocols/test_dispatch_bindings_unit.py` 0 → 11 methods and `tests/protocols/link/test_link_unit.py` 19 → 25. None of it is this PR's work.

Two things about baselines worth keeping, since both have caught agents on this repo. First, `git archive origin/main` measures 859 passed / 35 skipped, but **18 of those skips are an artifact** of the archive not being a git tree — all `tests/test_tier_guard.py`, "git could not be run" — and they pass in a real checkout, so use `git clone`. Second, CI's own numbers are a different scope again and are not comparable to a local run: **Unit Tests** runs `pytest -q --ignore=tests/integration` and never regenerates the sample captures, while **Integration** runs the full suite after a deliberately partial regeneration on runners lacking the optional engines (`pcap`, `pcapfile`).

All 19 pre-existing capture files byte-identical to an untouched `origin/main` tree, and all 30 `tree`/`json` dumps under `ip=True, tcp=True, reassembly=True` byte-identical. mypy 9 errors, all pre-existing classes — 5 in `_support.py` identical to `main`, 4 `scapy.all` stub errors of the class `pcap.py` already has 15 of. pylint **10.00/10**. Tree under test proven on every run via `pcapkit.__file__` under `PYTHONSAFEPATH=1`.

## One coupling to know about before merging

`tests/_support.py` on `main` has no `time_limit`; it lives on the unmerged #432. It is vendored here **byte-identically to that branch**, so whichever lands first, the second merges cleanly. If #432's version changes before merge, this copy must be re-synced.


tests/protocols/test_option_roundtrip_unit.py
```diff
@@ -0,0 +1,890 @@
...
+    #
+    #     HOPOPT(...)    -> options=[SMF_DPD, PadN]
+    #     IPv6_Opts(...) -> ProtocolError: IPv6-Opts: invalid format
+    #
+    # Identical on 3.10.20 and 3.14.7, so this one is not interpreter-dependent.
+    #
+    # The deadline in the sweep stays regardless. It is protection against the
+    # *next* non-progress defect, not against this one.
+    'ipv6-opts-option/SMF_DPD': Gap(
+        'PARSE', 'invalid format',
```
JarryShaw: **`fragment='invalid format'` is too generic to pin this entry (and five siblings) to the recorded defect**

The module docstring (lines 46-53) says the point of checking the fragment in both directions is that fixing the defect "turns this module red... which is the reminder to delete the entry." That guarantee only holds if the fragment can't be satisfied by a *different* failure. Here the fragment is the bare string `'invalid format'`, which is the tail of many textually similar but semantically distinct `ProtocolError` messages in the same file.

Measured on `d7240dc0c`, `PYTHONSAFEPATH=1`, `.venv` 3.14.7 (`pcapkit.__file__` confirmed resolving into the PR tree), calling `options.roundtrip()` directly on each case:

```
tcp-option/User_Timeout_Option        -> CONSTRUCT | ProtocolError: TCP: [OptNo 28] invalid format
ipv6-route-type/Source_Route          -> CONSTRUCT | ProtocolError: IPv6-Route [TypeNo <class 'type'>]: invalid format
ipv6-route-type/Type_2_Routing_Header -> CONSTRUCT | ProtocolError: IPv6-Route: [TypeNo <class 'type'>] invalid format
hip-parameter/HOST_ID                 -> CONSTRUCT | ProtocolError: HIPv2: invalid format
httpv2-frame/PRIORITY                 -> CONSTRUCT | ProtocolError: HTTP/2: [Type 2] invalid format
ipv6-opts-option/SMF_DPD (this entry) -> PARSE     | ProtocolError: IPv6-Opts: invalid format
```

For *this* entry specifically: `pcapkit/protocols/internet/ipv6_opts.py:497` raises the literal string `'IPv6-Opts: invalid format'` (the threshold check this entry names, exactly one site) -- but the same read path can also raise four *other*, differently-worded "...invalid format" errors at lines 575, 577, 579 and 619 (each `f'{self.alias}: [OptNo {code}] invalid format'`). `assertIn('invalid format', outcome.detail)` cannot tell the recorded #441 threshold-mismatch apart from any of those. If a future change fixed the line-497 over-count but broke one of those other checks for the same option, this entry would still read as "failing in the recorded way" -- the exact rot `test_round_trip_is_identity_or_a_recorded_gap`'s docstring says the reverse-check exists to catch, sailing through undetected.

The same genericness affects the other five cases above: `tcp.py`, `hip.py`, `ipv6_route.py` and `httpv2.py` each raise `ProtocolError(...'invalid format')` from many distinct sites in the surrounding code (26, 31, 3 and 12 occurrences respectively, by grep), differing only in a bracketed code/type identifier the bare fragment never checks.

Not a functional defect -- all six cases genuinely fail today, and the round-trip cycle itself is sound. It's a gap in the test's own specificity: tightening the fragment to include the distinguishing prefix (e.g. `'IPv6-Opts: invalid format'` instead of `'invalid format'` here, `'TCP: [OptNo 28]'` for the TCP entry, etc.) would restore the guarantee the docstring claims. Minor, and not something I'd block the PR on, but worth a follow-up since it's exactly the "loose fragment" failure mode this table is designed to prevent.

JarryShaw:
## Review of PR #440 at head `d7240dc0c21a1c618a6c88741ba00325e12fa5c9`

Copilot is out of tokens, so I reviewed this by hand as an adversarial pass over the recorded expectations, since the PR itself changes no library code (`git diff origin/main pr-440 -- pcapkit/` is empty, confirmed) and its entire value is in whether `tests/protocols/test_option_roundtrip_unit.py` and `examples/generators/options.py` say true things.

### CI

`gh pr view 440 --json statusCheckRollup` at the time of this review: every check green, including `Python 3.10` and `Integration Python 3.10` (the two that were red for hours per the task brief). `Gate (full suite, Python 3.14)` and `Docs test gate` show `SKIPPED`, which is a gating condition, not a failure. `git rev-list --count HEAD..origin/main` is `0`, so this repo checkout is at `main` tip -- no stale-checkout risk.

### What I ran, and against what

Cloned fresh (`git clone` + `git fetch origin pr-440:pr-440` + `git checkout pr-440`) into `/tmp/pr440review/pr440-tree`, HEAD confirmed `d7240dc0c21a1c618a6c88741ba00325e12fa5c9`. A second clean clone of `main` (`c28ffc287...`) went to `/tmp/pr440review/main-tree` as a baseline. Two interpreters, both with `PYTHONSAFEPATH=1`, `PYTHONPATH` pointed at the tree under test, and `pcapkit.__file__` printed and confirmed resolving inside that tree before trusting any result:

- `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` -- 3.14.7, has `scapy` 2.7.0.
- `/tmp/mh310-a3a0ca27/venv/bin/python` -- 3.10.20, **no** `scapy` (re-pointed at the PR tree via `PYTHONPATH`; verified it wasn't silently resolving `pcapkit` from the venv's own main-checkout install).

Ran:
- `python -m unittest tests.protocols.test_option_roundtrip_unit -v` on both interpreters -- **7/7 pass on 3.14.7, 7/7 pass on 3.10.20**.
- Loaded `examples/generators/options.py` directly and cross-checked every one of the 258 cases' `roundtrip()` outcome against `_gap_for()` on both interpreters: **258 total, 173 OK / 85 recorded-gap / 0 mismatches on 3.14.7; 169 OK / 89 recorded-gap / 0 mismatches on 3.10.20 (`SCHEMA_ABC_IS_PER_CLASS=False` there, as expected)**. Matches the PR's claimed 173/169 exactly.
- `examples/generators/make_samples.py` (3.14.7, which has `scapy`) to actually generate the five `options-*.pcap` fixtures, then `python -m unittest tests.protocols.test_option_coverage_runtime -v` on **both** interpreters against the generated captures -- **3/3 pass on each**.
- Determinism: regenerated the five captures twice more (a fresh process, and again under `PYTHONHASHSEED=12345`) -- **byte-identical** to the first run in both cases (`cmp` clean on all five).
- Cross-contamination check on the "PCAP-NG writes no fixture" reasoning: generated captures from the `main` baseline (generators `pcap`, `pcapng`, `legacy` only) and from the PR tree (which adds `options` as a fourth generator, run *last*), and compared the 13 pre-existing generated files -- **all 13 byte-identical** between `main` and the PR. So adding the registry-driven sweep after the others has no observable side effect on their output, which is also indirect evidence the registries aren't being mutated by enumeration.
- Counted `pcapng-block` (11) + `pcapng-option` (40) + `pcapng-record` (3) + `pcapng-secrets` (4) cases = **58**, matching the "58 codes exercised without a fixture" claim exactly; and the 27-entry `pcapng-option` dict-comprehension in `EXPECTED_FAILURES` (14 `if_*` + 4 `epb_*` + 3 `ns_*` + 5 `isb_*` + 1 `pack_*`) matches "27 of 40 block options unconstructible" from the shared-`self._opt` cause.
- Spot-verified ~15 `file:line` citations in `EXPECTED_FAILURES`/`INTERPRETER_GAPS` directly against `git show origin/main:<path>`, including the two the task flagged as previously mis-attributed: `hopopt.py` and `ipv6_opts.py` both show `'len': (8, 8)` (not `(1, 8)`) at their `_SMFDPDOption` definitions, confirming #432 landed symmetrically; and `ipv6_opts.py:434` does carry the extra `test: 'SMFDPDTestFlag' = ForwardMatchField(...)` field that `hopopt.py`'s equivalent class lacks, confirming the #441 re-attribution in this branch is the correct one (not the "missing #432 fix" explanation the task says was wrong). Also confirmed `IPv4.__option__` (ipv4.py:196) and `HIP.__parameter__` (hip.py:377, not `__param__`) are real registries post-#434, `pcapng.py:1248`'s `isinstance(record, (IPv4Record, IPv6Record))` matches the #439 attribution verbatim, and `ProtocolBase._lookup_registry` (protocol.py:1242) reads via `if code in registry` before ever subscripting, so enumeration genuinely cannot grow a registry.

### Finding

One inline comment posted (below), on `tests/protocols/test_option_roundtrip_unit.py` line 276:

**`fragment='invalid format'` is too generic for at least 6 of the ~85 `EXPECTED_FAILURES` entries** (`tcp-option/User_Timeout_Option`, `ipv6-route-type/Source_Route`, `ipv6-route-type/Type_2_Routing_Header`, `hip-parameter/HOST_ID`, `httpv2-frame/PRIORITY`, `ipv6-opts-option/SMF_DPD`). Each of these sits in a source file that raises dozens of textually-different-but-substring-identical `"...invalid format"` `ProtocolError`s from other, unrelated checks (measured: 26 sites in `tcp.py`, 31 in `hip.py`, 3 in `ipv6_route.py`, 12 in `httpv2.py`, 5 in `ipv6_opts.py`). `assertIn(fragment, outcome.detail)` can't distinguish the recorded defect from a different one at a nearby line in the same file, which is exactly the failure mode `test_round_trip_is_identity_or_a_recorded_gap`'s own docstring says the two-directional check exists to prevent. Demonstrated with actual captured `outcome.detail` strings for all six cases in the comment. Not a functional defect in the library, and not something I'd block merge on -- it's a specificity gap in the test's own guard-rail, worth a tightening follow-up.

### What I did not cover

- Did not exhaustively re-derive all ~85 `EXPECTED_FAILURES`/`INTERPRETER_GAPS` entries' root causes from scratch -- spot-checked roughly 15-20 of the more load-bearing ones (the two re-attributed SMF_DPD entries, the #439 isinstance mechanism, the #434 registries, the pcapng packet_data / `self._opt` / `_isb_interface_id` claims, the IPv6 Quick-Start `SchemaField(length=5)` claim) plus the six flagged above for fragment looseness; did not check all 27 `pcapng-option/self._opt` entries' individual line numbers or the 16 HIP tuple-vs-list entries' individual attributions line-by-line.
- Did not independently reproduce the full CI matrix (909 passed/17 skipped on 3.14, 838 passed/88 skipped on 3.10) or the "30 tree/json dumps byte-identical" claim -- ran the two new test modules specifically (10/10 passing on both interpreters) rather than the whole suite, and trusted the now-green CI rollup for the rest.
- Did not audit pylint output or anything style-level, per the review brief.

### Verdict

**Good to merge.** The enumeration is genuinely registry-driven (confirmed non-mutating, confirmed IPv4/HIP now use real registries post-#434), the determinism story for the five new fixtures holds under repeated regeneration and a different hash seed, the numeric claims (258/173/169/85/58/27) all check out exactly against a live run on both interpreters, the two SMF_DPD re-attributions are correct, and CI is green including the previously-red 3.10 jobs. The one finding above is a minor test-quality gap in a handful of expected-failure fragments, not a defect in the library or in the round-trip mechanism itself.

tests/protocols/test_option_roundtrip_unit.py
JarryShaw: Fixed in 4535f72f2 — the finding is right, and measuring it made it worse than "six entries": there were **seven**, because `hip-parameter/HIP_TRANSFORM` recorded the bare `invalid parameter` on the same reasoning.

Counted on `c28ffc287`: `invalid format` occurs **205** times across 13 modules (31 `internet/hip.py`, 26 `transport/tcp.py`, 22 `internet/ipv4.py`, 19 each in `internet/ipv6_opts.py` and `internet/hopopt.py`, 12 `application/httpv2.py`, …), and `invalid parameter` 3 times, all in `internet/hip.py`. So the check was satisfiable by a regression at any of them, exactly as you said.

Each fragment now carries the alias and whatever bracketed code the message actually prints, captured from the real `outcome.detail` rather than guessed:

| case | fragment | sites it can match |
|---|---|---|
| `tcp-option/User_Timeout_Option` | `TCP: [OptNo 28] invalid format` | 1 of 26 |
| `ipv6-opts-option/SMF_DPD` | `IPv6-Opts: invalid format` | 1 of 19 |
| `hip-parameter/HIP_TRANSFORM` | `HIPv2: [ParamNo 577] invalid parameter` | 1 of 3 |
| `hip-parameter/HOST_ID` | `HIPv2: invalid format` | 1 of 31 |
| `httpv2-frame/PRIORITY` | `HTTP/2: [Type 2] invalid format` | 1 of 12 |

The 25 `[OptNo …]` sites in `tcp.py` share one literal (`f'{self.alias}: [OptNo {schema.kind}] invalid format'`), but only the User Timeout handler can print `28`, so the code is what discriminates. The two bare-alias forms are the only sites in their module that omit the bracket at all — `ipv6_opts.py:497` and `hip.py:761` — so those pin a single line each.

### The two IPv6-Route entries needed a different answer, and turned up a defect

Their real details are:

```
ipv6-route-type/Source_Route           "ProtocolError: IPv6-Route [TypeNo <class 'type'>]: invalid format"
ipv6-route-type/Type_2_Routing_Header  "ProtocolError: IPv6-Route: [TypeNo <class 'type'>] invalid format"
```

`<class 'type'>` is the **builtin** `type`: `_read_data_type_src`/`_2`/`_rpl` are `(self, schema, *, header)` and bind no `type`, so the f-string picks up the builtin. Filed as #442 — three sites, `ipv6_route.py:462`, `:506`, `:546`, and the read dispatch (`meth(schema.data, header=schema)`) passes no type, so `header.type` is the fix. I swept every `{type}` interpolation in `pcapkit/`: the ten in `internet/mh.py` are all fine, each sitting in a `_make_opt_*` whose first parameter really is `type: 'Enum_Option'`. These three are the only broken ones.

Pinning the literal rendering would therefore bake #442 into this table and turn it red when #442 is fixed — a false round-trip regression. So `fragment` now also accepts a tuple, all of whose members must be present, and those two record `('IPv6-Route', '[TypeNo', 'invalid format')`: stable across #442, and still narrowing 205 candidates to the three `[TypeNo` lines of that one module. The `defect` field carries the exact line, as before.

### Verification

7 passed / 299 subtests on both **3.14.7** and **3.10.20** (`pcapkit.__file__` confirmed pointing at the PR tree on each). Negative control, since a fragment assertion that is never reached would pass just as quietly: setting the TCP fragment to `[OptNo 99]` and the tuple's last member to `nonsense` fails exactly 3 subtests —

```
AssertionError: 'TCP: [OptNo 99] invalid format' not found in 'ProtocolError: TCP: [OptNo 28] invalid format'
AssertionError: 'nonsense' not found in "ProtocolError: IPv6-Route [TypeNo <class 'type'>]: invalid format"
AssertionError: 'nonsense' not found in "ProtocolError: IPv6-Route: [TypeNo <class 'type'>] invalid format"
```

— so both the string and the tuple form are live. The failure message now also prints the detail it received, which is what made diagnosing these quick.

JarryShaw:
## Review of `4535f72f2` at sha `651665214`

Standing in for Copilot review. Per the dispatch instructions, the last verdict was given at `d7240dc0c` (good to merge, with one inline finding about over-broad `EXPECTED_FAILURES` fragments); the only new reviewable content since then is `4535f72f2` ("tests: pin the round-trip gaps to messages that identify their site"), plus a pass-through merge of `main` (PR #436, not reviewed here). This comment attacks that commit's claims directly.

### Scope confirmed

- Fetched `refs/pull/440/head` → `65166521476efdb5c3a4a5c484c899292bdbb486`, matching the requested prefix `651665214`.
- `origin/main` = `f50436a8a1f4cfd9e2f9dc96080fa9f4d8ca08a4`, matching the stated base.
- `git rev-list --count pr440-review..origin/main` = `0` — the branch is fully caught up with `main`.
- `4535f72f2 --stat` touches exactly one file: `tests/protocols/test_option_roundtrip_unit.py` (39 insertions, 14 deletions), as claimed.

### CI status (checked, and waited for completion)

At dispatch time several checks were `IN_PROGRESS` (Python 3.10/3.13/3.15 unit tests, several Integration jobs). Re-polled `gh pr view 440 --json statusCheckRollup` until settled. Final state: **every check is `SUCCESS`** except two legitimately-conditional `SKIPPED` gates (`Docs test gate`, `Gate (full suite, Python 3.14)`) and `deploy-pages`/`CodeQL` which are also green. No red checks anywhere.

### Claim 1 — the fragment counts (verified exact)

Ran `git grep -c "invalid format" origin/main -- pcapkit/` and `git grep -c "invalid parameter" origin/main -- pcapkit/`:

```
httpv1.py 1, httpv2.py 12, hip.py 31, hopopt.py 19, ipv4.py 22, ipv6_opts.py 19,
ipv6_route.py 3, mh.py 26, pcapng.py 2, schema/hip.py 2, schema/ipv4.py 4,
sctp.py 38, tcp.py 26   → sum = 205, across 13 modules
"invalid parameter": hip.py 3, nowhere else
```

Both numbers (205 across 13 modules, 3 all in `hip.py`) match the commit message exactly, as do the per-module call-outs (`hip.py:31`, `tcp.py:26`).

### Claim 2 — each fragment against the *real* `outcome.detail` (verified exact)

Checked out `651665214` into an isolated worktree and ran the harness directly (not through the test's assertions) to print the raw detail for all seven cases. Interpreter: `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` 3.14.7, `PYTHONSAFEPATH=1`, `PYTHONPATH` pointed at the checked-out tree, `pcapkit.__file__` printed and asserted to be under that tree before trusting anything:

```
pcapkit.__file__ = .../agent-a4d4d937fcc41ff27/pcapkit/__init__.py
tcp-option/User_Timeout_Option        -> 'ProtocolError: TCP: [OptNo 28] invalid format'
ipv6-opts-option/SMF_DPD              -> 'ProtocolError: IPv6-Opts: invalid format'
hip-parameter/HIP_TRANSFORM           -> 'ProtocolError: HIPv2: [ParamNo 577] invalid parameter'
hip-parameter/HOST_ID                 -> 'ProtocolError: HIPv2: invalid format'
httpv2-frame/PRIORITY                 -> 'ProtocolError: HTTP/2: [Type 2] invalid format'
ipv6-route-type/Source_Route          -> "ProtocolError: IPv6-Route [TypeNo <class 'type'>]: invalid format"
ipv6-route-type/Type_2_Routing_Header -> "ProtocolError: IPv6-Route: [TypeNo <class 'type'>] invalid format"
```

Every recorded fragment/tuple is a genuine substring/subset of the real detail. No broken entry.

Spot-checked the "narrows to one site" sub-claims:
- `Enum_Option.User_Timeout_Option = 28` (`pcapkit/const/tcp/option.py:106`) against the 25 other `[OptNo {schema.kind}] invalid format` raises in `tcp.py` — the bracketed *value* is what narrows it, since the raise template itself is shared code across ~25 methods.
- `Enum_Parameter.HIP_TRANSFORM = 577` (`pcapkit/const/hip/parameter.py:55`) — matches.
- `Frame.PRIORITY = 0x02` (`pcapkit/const/http/frame.py:28`) against the site at `httpv2.py:572`, which matches the defect's cited line exactly.
- `git grep -n "invalid format" pcapkit/protocols/internet/ipv6_opts.py` on `main`: line 497 (`raise ProtocolError('IPv6-Opts: invalid format')`) is the **only** one of 19 hits with no `[OptNo ...]` bracket — matches the claim.
- Same for `hip.py`: line 761 (`f'HIPv{version}: invalid format'`) is the only bracket-less raise that can render `"HIPv2: invalid format"` (the other bracket-less pair at 539/541 renders `"HIP: invalid format"`, a different string that wouldn't satisfy this fragment) — matches.

### Claim 3 — the assertion loop, and the empty-string form (verified)

```python
fragments = ((gap.fragment,) if isinstance(gap.fragment, str) else gap.fragment)
for fragment in fragments:
    if not fragment:
        continue
    self.assertIn(fragment, outcome.detail, ...)
```

- String form → wrapped in a 1-tuple, asserted once. Correct.
- Tuple form → iterated as-is, every non-empty member asserted. Correct.
- Empty string `''` → wrapped as `('',)`, `if not fragment: continue` skips it, so only `status` is checked. Six existing entries (`hip-parameter/ENCRYPTED`, three `pcapng-block/*`, two `pcapng-secrets/*`) rely on exactly this path and all pass cleanly in the full run below — the empty case is not regressed.
- `test_expected_failures_name_real_cases` compares `(status, fragment)` tuples between `EXPECTED_FAILURES` and `INTERPRETER_GAPS` for overlapping labels. `INTERPRETER_GAPS` only holds `pcapng-*` labels, which never intersect the seven touched here, so this is a non-issue in practice, but the comparison is plain tuple equality and works fine whether `fragment` is a `str` or a `tuple[str, ...]` regardless.

### Claim 4 — the tuple form's reason to exist (verified against the actual issue)

Read GitHub issue #442 (open): it names exactly the three sites `ipv6_route.py:462,506,546`, confirms the bare `{type}` resolves to the builtin because none of the three enclosing methods bind a `type` parameter, and proposes the fix `header.type` (plus a colon-placement cleanup at :462 to match :506/:546's convention). Neither part of that proposed fix touches the substrings `'IPv6-Route'`, `'[TypeNo'`, or `'invalid format'` — only the interpolated value between `[TypeNo ` and `]`, and the colon's position relative to the bracket. So the tuple fragment is provably stable across the fix #442 proposes, which is exactly the claim. Matching the *literal* rendering (`<class 'type'>`) instead would indeed have gone red the moment #442 lands, for a reason unrelated to the round-trip defect these entries actually pin (`ipv6_route.py:276`, unit-conversion bug).

### Claim 5 — negative control (reproduced)

Copied the test file, changed the TCP entry's fragment to `'TCP: [OptNo 99] invalid format'` and both `ipv6-route-type` tuples' last member to `'nonsense'`, ran it in place (then deleted the copy — tree confirmed clean after, `git status` = "nothing to commit"):

```
FAILED (failures=3)
  case='tcp-option/User_Timeout_Option'        -- 'TCP: [OptNo 99] ...' not found
  case='ipv6-route-type/Source_Route'          -- 'nonsense' not found
  case='ipv6-route-type/Type_2_Routing_Header' -- 'nonsense' not found
```

Exactly 3 subtest failures, matching the commit's negative-control claim — the assertion is reached, not vacuous.

### The 7/299 figure (reproduced exactly, both interpreters)

Using a `CountingResult` that hooks `addSubTest`, ran the whole `OptionRoundTripTests` class:

| Interpreter | tree confirmed via `pcapkit.__file__` | tests run | subtests | failures |
|---|---|---|---|---|
| 3.14.7 (`.venv`) | under the checked-out worktree | 7 | **299** | 0 |
| 3.10.20 (`/tmp/mh310-a3a0ca27/venv`, tree forced via `sys.path[0]`) | under the checked-out worktree | 7 | **299** | 0 |

Matches "7 passed / 299 subtests on both 3.14.7 and 3.10.20" exactly.

### The "full suite 909 passed / 17 skipped" figure — could not reproduce; flagging, not blocking

This one doesn't check out. CI's own `Unit Tests` jobs (the `make test` selection, i.e. `pytest -q --ignore=tests/integration --ignore-glob='*_runtime.py' --ignore-glob='*_regression.py'`) report **721 passed, 69 skipped, 1040 subtests passed**, identically across Python 3.10–3.15. CI's `Integration Python *` jobs report **855 passed, 88 skipped, 1188 subtests passed**, also identically across versions. I reproduced the 3.10 unit-job number exactly locally (721/69/1040, tree confirmed via `pcapkit.__file__`). My local 3.14.7 run came back as 782 passed/8 skipped/1082 subtests — 61 more passes than CI's own 3.14 job, which is a dev-venv-has-more-optional-deps artifact (`RUNTIME_DEPS`-gated tests running instead of skipping), not a code issue. None of these four numbers is 909/17, and I did not run `make samples && pytest -q` (the truly-unfiltered selection, needing generated fixtures) to check whether that configuration hits it — that's the one combination I didn't try. What I can say confidently: **every configuration I ran was 100% clean, zero failures, zero errors**, matching the green CI. The specific "909/17" figure in the commit message appears to be inaccurate or drawn from an environment/selection I couldn't reproduce; it doesn't affect the correctness of the diff itself.

### Verdict

**GOOD TO MERGE** at sha `651665214`. Commit `4535f72f2` does exactly what it claims for the seven `EXPECTED_FAILURES` entries: every new fragment was measured against the real `outcome.detail` (not guessed), the counts backing the "too generic" argument check out exactly against `origin/main`, the tuple form is implemented correctly for the string/tuple/empty-string cases and is genuinely robust to the fix issue #442 proposes, and the negative control is reproducibly non-vacuous. CI is fully green on this sha. The one soft spot is the commit message's "full suite 909 passed / 17 skipped" line, which I could not reproduce under CI's own selection or locally on either interpreter (real numbers: 721/69 unit, 855/88 integration, consistently, with zero failures throughout) — worth a correction in the message, but not a reason to hold the CR.

JarryShaw:
Thanks — the numbers do disagree, but the stale one is the **PR body's**, not the `909 / 17` figure. Settled by counting collection at each sha rather than comparing run summaries:

| sha | collected |
|---|---|
| `d7240dc0c` (previous verdict) | **926** |
| `4535f72f2` (my fragment commit) | **926** |
| `651665214` (current head, after the `main` merge) | **943** |

So:

- `909 passed / 17 skipped` = **926**, which is exactly the collection at `4535f72f2` — the tree I ran it on. That figure is right for that sha. My commit added no test methods, which is why `d7240dc0c` and `4535f72f2` collect identically.
- The body's `886 passed / 17 skipped` = **903**, which matches no sha currently on this branch. It was measured before two rounds of `main` merges landed here, so it is simply out of date rather than wrong at the time.
- `943 − 926 = 17` is precisely what the pass-through merge of `main` brought in: `tests/protocols/test_dispatch_bindings_unit.py` 0 → 11 test methods and `tests/protocols/link/test_link_unit.py` 19 → 25, i.e. +11 and +6. Those are #436's tests, not this PR's.

**On CI's own figures** (721 / 69 for Unit Tests, 855 / 88 for Integration) — those are not comparable to a local `pytest tests -q`, because the two jobs run different scopes:

- **Unit Tests** runs `python -m pytest -q --ignore=tests/integration` and does *not* regenerate the sample captures, so everything fixture-dependent skips. Hence the far larger skip count.
- **Integration** runs `python -m pytest -q` after a "Regenerate sample captures" step, but that step is partial by design, and the runners lack the optional engine dependencies (`pcap`, `pcapfile`, and friends) that are installed here — so tests that pass locally skip there.

That difference is expected rather than a discrepancy: neither job is trying to reproduce a local full-suite number. What matters for the verdict is that every configuration came back with zero failures, which both of us observed.

I am updating the body's suite line to the current post-merge measurement so it stops misleading the next reader. Good catch on the mismatch even though the cause turned out to be elsewhere — the body had been carrying a figure from two merges ago and nobody had noticed.

Issue #446, opened 2026-09-17
ForwardMatchField's non-consuming bytes count toward Schema.__len__, so correct input fails a declared-length check


JarryShaw:
`ForwardMatchField` exists to look ahead without consuming — it reads bytes so a later field's length can depend on them, and leaves the buffer position alone. But the bytes it matched still occupy a slot in the schema's `__buffer__`, and `Schema.__len__` is defined as the length of the concatenated buffer:

```python
# pcapkit/protocols/schema/schema.py:443
def __len__(self) -> 'int':
    return len(self.__bytes__())
```

So any schema containing a `ForwardMatchField` reports itself longer than the octets it actually read, by exactly the width of the forward match. Where that length is then checked against a declared area, the check fails on correct input.

### Observed

The CGA Parameters option is the case that exposed it. `CGAParameter` (`pcapkit/protocols/schema/internet/mh.py:499-521`) carries

```python
public_key_test: 'ANSIKeyLengthTest' = ForwardMatchField(BitField(length=2, namespace={
    'len': (8, 8),
}))
```

and is nested by `CGAParametersOption` through `ListField(length=lambda pkt: pkt['length'], item_type=SchemaField(schema=CGAParameter))`. With the enclosing-length lookup corrected (that is issue #445, a separate fault that hits first), a well-formed 40-octet option then fails:

```
FieldValueError: Field parameters has invalid length.
```

because the `ListField` compares its declared area against item lengths that are each 2 octets too large.

### The same root is already filed once

#441 is the same mechanism seen from the other end: `ipv6_opts.py`'s SMF-DPD schema carries a `test` `ForwardMatchField` that `hopopt.py`'s equivalent does not, so identical octets parse under HOPOPT and fail under IPv6-Opts, the nested schema reporting 4 octets where it read 3. That issue can be closed by deleting the stray field, which is worth doing on its own — but it treats the symptom. As long as a non-consuming field counts toward `len(schema)`, the next schema that legitimately needs a forward match hits the same wall, and `CGAParameter` is a case where the field is *not* stray and cannot simply be deleted: the public key's length genuinely has to be read before the key can be sized.

### Fix

The question is what `len(schema)` is for. It is used as "how many octets did this schema account for", which is the consumed count — so a non-consuming field should not contribute. Either:

- exclude forward-match slots from `__bytes__`/`__len__`, keeping them in `__buffer__` for value access; or
[...]

Issue #445, opened 2026-09-17
A nested schema cannot reach the enclosing packet's fields by name, so CGA Parameters raises KeyError: 'length'


JarryShaw:
A nested schema's field callbacks receive a packet dict that does **not** contain the enclosing schema's fields by name. The enclosing packet is reachable, but only under a `__packet__` key, and that convention is documented in exactly one comment in the whole tree. So the natural thing to write — `length=lambda pkt: pkt['length']`, exactly as every top-level schema writes it — raises `KeyError` when the schema is nested.

### The live casualty

`pcapkit/protocols/schema/internet/mh.py:515-521`, the CGA Parameters option:

```python
class CGAParameter(Schema):
    ...
    extensions: 'list[CGAExtension]' = OptionField(
        length=lambda pkt: pkt['length'] - 25 - len(pkt['public_key']),
        ...
    )
```

`CGAParameter` has no `length` field. The `length` this wants belongs to the enclosing `CGAParametersOption` (`:529-536`), which reaches it via `ListField(length=..., item_type=SchemaField(schema=CGAParameter))`.

A well-formed 40-octet option therefore cannot be parsed on `f50436a8a`:

```python
raw = bytes.fromhex('11040000123400000c1e'
                    '0000000000000000000000000000086f'
                    '0000000020010db8'
                    '00'
                    '3003010203')          # len(raw) == 40
MH(io.BytesIO(raw), len(raw), extension=True)
-> KeyError: 'length'
```

Four `SchemaWarning: packet length < 0` warnings (-17, -25, -26, -31) come out on the way, which is the same arithmetic going negative during the option scan.

### Mechanism

`SchemaField.pack` and `SchemaField.unpack` (`pcapkit/corekit/fields/misc.py`) hand the nested schema a dict with the parent nested *under a key* rather than merged:

```python
return value.pack({
    '__packet__': packet,
})
```

So `pkt['length']` is absent and `pkt['__packet__']['length']` is what exists. Confirmed by fixing just that lookup in place, which turns the `KeyError` into a different error (see below) rather than another `KeyError`.

### Why this is a design problem and not one bad lambda

The `__packet__` contract is real but undiscoverable:

- It is explained in **one** place, `pcapkit/protocols/schema/misc/pcapng.py:155-157`, in a docstring belonging to an unrelated helper.
[...]

PR #449, opened 2026-09-17
protocols: drop ipv6_opts' stray SMF_DPD test field, fix two length underflows

### LINKED ISSUE #438: A wire-derived length underflows into a struct format string, raising bare struct.error on untrusted input

### LINKED ISSUE #441: ipv6_opts' SMFIdentificationBasedDPDOption has a stray test field HOPOPT lacks, so identical octets fail there
`SMFIdentificationBasedDPDOption` in `pcapkit/protocols/schema/internet/ipv6_opts.py` declares a `test` `ForwardMatchField` that its HOPOPT twin does not. The two modules are otherwise line-for-line equivalents, and this one extra field makes the same octets parse under HOPOPT and fail under IPv6-Opts.

## The difference

Enumerating the class body from `origin/main` (`c28ffc287`):

```
hopopt.py     SMFIdentificationBasedDPDOption fields = ['info', 'tid', 'id']
ipv6_opts.py  SMFIdentificationBasedDPDOption fields = ['test', 'info', 'tid', 'id']
                                                        ^^^^^^ line 434
```

```python
# pcapkit/protocols/schema/internet/ipv6_opts.py:434
    test: 'SMFDPDTestFlag' = ForwardMatchField(BitField(length=1, namespace={
        'mode': (0, 1),
    }))
```

Normalising both modules for their protocol names leaves ten differing lines: a module docstring, a five-line explanatory comment, and these three.

## Why it is a stray rather than intentional

- The **enclosing** `_SMFDPDOption` already declares a `test` `ForwardMatchField` (`length=3`), in **both** modules.
- **Nothing reads the nested one.** `smf_dpd_data_selector` reads `pkt['test']['mode']` and `pkt['test']['len']`, and that `pkt` is the *enclosing* schema's.
- It is the only field in the class carrying no `#:` comment, unlike `info`, `tid` and `id` around it.

It predates #432, which neither added nor removed it.

## Effect

`ForwardMatchField` consumes nothing from the stream but still occupies a slot in `__buffer__`, so it counts toward `len(schema)`. The nested schema therefore over-reports its length by one octet, and `OptionField` accumulates that over-report against the declared option area until the area check fails.

Same wire layout, same octets, different outcomes — built through HOPOPT as `1100080100010100`, option area `080100010100`:

```
                 len(schema)
hopopt                 3        <- correct
ipv6_opts              4        <- one too many

HOPOPT(built, 8, extension=True)     -> options=[SMF_DPD, PadN]
IPv6_Opts(built, 8, extension=True)  -> ProtocolError: IPv6-Opts: invalid format
```

The error is raised at `pcapkit/protocols/internet/ipv6_opts.py:497`, the option-area threshold check in `_read_ipv6_opts`, reached from `read` at :315 via `options=self._read_ipv6_opts(schema.len * 8 + 6)` and `unpack` → `__post_init__` at `protocols/protocol.py:678`.

**Byte-for-byte identical on Python 3.10.20 and 3.14.7**, so unlike the PCAP-NG failures in #439 this is not interpreter-dependent.

## Provenance

Surfaced by the option round-trip harness in #440, which records `ipv6-opts-option/SMF_DPD` as a `PARSE` gap. Two corrections happened on the way to it and are worth recording, since both were wrong in ways that looked right:

- The first explanation offered was that #432's SMF_DPD fix had been "applied once where it was needed twice". **That is false** — #432 (`84233da3a`) touched both schema modules symmetrically, 26 changed lines each, and both now carry `'len': (8, 8)` with zero `(1, 8)` remaining.
- My own attempt to check that claim reported both files still carrying `'len': (1, 8)` at line 374. **Also false** — I was reading a working tree two commits behind `origin/main`. On `origin/main` line 374 is unrelated and the namespace reads `(8, 8)` in both.

Not fixed in #440 deliberately: that PR keeps `git diff origin/main -- pcapkit/` empty, which is what makes a 258-case harness reviewable.

JarryShaw:
## Summary

Two related schema defects in `pcapkit/protocols/schema/internet/`, both surfaced while investigating option/parameter length arithmetic in that directory.

### #441 — stray `test` field makes `len(schema)` over-report by one

**Defect.** `pcapkit/protocols/schema/internet/ipv6_opts.py`'s `SMFIdentificationBasedDPDOption` declared:

```python
test: 'SMFDPDTestFlag' = ForwardMatchField(BitField(length=1, namespace={
    'mode': (0, 1),
}))
```

right before `info`. Its HOPOPT twin (`pcapkit/protocols/schema/internet/hopopt.py`) has no such field — the two classes are otherwise line-for-line equivalent.

**Mechanism.** `ForwardMatchField` consumes no bytes from the stream but still occupies a `Schema.__buffer__` slot (`pcapkit/protocols/schema/schema.py`), so it counts toward `len(schema)`. The nested schema therefore over-reported its length by one octet, and `OptionField` (`pcapkit/corekit/fields/collections.py`) accumulated that over-report against the declared option area until the threshold check in `pcapkit/protocols/internet/ipv6_opts.py` (`_read_ipv6_opts`) raised `ProtocolError: IPv6-Opts: invalid format` on octets that parse cleanly under HOPOPT.

**Why it was a stray field, not a missing one — verified against RFC 6621 §6.1.1.** The I-DPD option header layout is: octet 0 Option Type, octet 1 Opt Data Len, octet 2 bit 0 the H-bit/mode (0 = I-DPD), bits 1–3 TidTy, bits 4–7 TidLen, then the TaggerID and Identifier. There is no separate "test" octet in the wire format — the mode bit is part of the same octet as TidTy/TidLen, which `SMFIdentificationBasedDPDOption.info` (`BitField(length=1, namespace={'mode': (0,1), 'type': (1,3), 'len': (4,4)})`) already reads. Confirmed by usage: `pkt['test']` is read only by `smf_dpd_data_selector`, which operates on the *enclosing* `_SMFDPDOption`'s packet dict (both modules already declare `test` there, at `length=3`), not on `SMFIdentificationBasedDPDOption`'s own fields — `smf_i_dpd_id_len` and `post_process` on this class never reference `pkt['test']`. So the field was dead weight left over from before #432, not a field HOPOPT is missing.

**Fix.** Removed the stray field from `ipv6_opts.py`. HOPOPT was already correct; ipv6_opts.py now matches it exactly.

**Independent corroboration from #440.** After this branch was up, #440 merged to `main` and added `tests/protocols/test_option_roundtrip_unit.py`, a harness that round-trips 258 option codes and records the ones that cannot close the cycle. It had an `EXPECTED_FAILURES` entry for exactly this defect (`'ipv6-opts-option/SMF_DPD'`, recorded as a `PARSE` gap at `ipv6_opts.py:434`, discovered independently of this PR). Merging `origin/main` into this branch and running that file turns the entry's assertion red — `AssertionError: 'OK' != 'PARSE'` — because the round trip now closes. That is the harness telling us, from a direction neither this PR nor #440 anticipated together, that the #441 diagnosis and fix are correct. Removed the now-stale `EXPECTED_FAILURES` entry in this PR (see Tests below); left the surrounding historical comment in place, updated to past tense.

### #438 — a wire-derived length underflows into a negative struct format

**Defect.** `SMFIdentificationBasedDPDOption.id` (both `hopopt.py` and `ipv6_opts.py`) sized itself as:

```python
id: 'bytes' = BytesField(length=lambda pkt: pkt['len'] - (
    1 if pkt['info']['type'] == 0 else (pkt['info']['len'] + 2)
))
```

with nothing floored at zero.

**Mechanism.** A peer declaring an `Opt Data Len` too small for the TaggerID it claims (e.g. `Opt Data Len = 0` for a null TaggerID, which needs at least 1) drives this negative. `_TextField.__call__` (`pcapkit/corekit/fields/strings.py`) then sets `self._template = f'{length}s'`, i.e. `'-1s'`, and the base `FieldBase.length` property (`pcapkit/corekit/fields/field.py`) computes `struct.calcsize('-1s')`, raising a bare `struct.error: bad char in struct format` — not one of pcapkit's own exception types, and uncatchable through `pcapkit.utilities.exceptions`.

**Fix.** Replaced the lambda with a named function, `smf_i_dpd_id_len`, in each module (matching the existing `mpl_opt_seed_id_len` convention already in both files), which raises `FieldValueError` when the computed length would be negative instead of letting it reach `struct`. `FieldValueError` is the exception this file already raises for every other "computed field parameter is invalid" case (`smf_i_dpd_tid_selector`, `smf_dpd_data_selector`, `mpl_opt_seed_id_len`), so no new exception type was needed.

**HIP siblings — same guard, needed.** `pcapkit/protocols/schema/internet/hip.py`'s `RegRequestParameter`, `RegResponseParameter` and `RegFailedParameter` all sized their registration-type list the same unfloored way: `ListField(length=lambda pkt: pkt['len'] - 1, ...)`. These *do* need the guard, though the failure mode differs from the `BytesField` case: `ListField.length` (`pcapkit/corekit/fields/collections.py`) returns its raw `int` directly rather than going through `struct.calcsize`, and its `unpack()` loop is `while length > 0`, so a negative length never crashes — it silently returns an empty list. Verified end-to-end with a crafted `Length = 0` `REG_REQUEST` parameter: pre-fix, this parses to `reg_type=()` with no exception and no diagnostic, exactly the "silently produce a wrong parse" failure the fix is meant to avoid, not a `struct.error`. Fixed with a shared helper, `registration_type_list_len`, reused by all three parameter schemas.

Checked, post-merge, that this change does not disturb any of #440's *other* HIP-related `EXPECTED_FAILURES` entries — `hip-parameter/R1_Counter`, `ENCRYPTED`, `HIP_TRANSFORM`, `HOST_ID`, and the sixteen-parameter `RECONSTRUCT`/tuple-vs-list group that includes `REG_REQUEST`/`REG_RESPONSE`/`REG_FAILED` themselves. The `RECONSTRUCT` group's gap is a different defect entirely (`_read_param_*` returning a tuple where `_make_param_*` needs a list) on the well-formed construct/parse/reconstruct path; this fix only changes behavior for a malformed `Length` that well-formed round-trip data never produces. Confirmed by running the full round-trip file after merging `origin/main`: 7 passed, 299 subtests, 0 failures — the one flip is exactly the SMF_DPD entry above, nothing else moved.

Two other siblings the issue named for the same check — `RegInfoParameter.reg_info` (`pkt['len'] - 2`) in `hip.py`, and the `CALIPSOOption`/`MPLOption` padding fields in `hopopt.py`/`ipv6_opts.py` (`pkt['len'] - 8 - pkt['cmpt_len']*4`, `pkt['len'] - 2 - ...`) — share the identical unfloored shape but are **not** touched by this PR; they were outside the lines the issue asked to check, and are being filed separately.

## Tests

- `tests/protocols/internet/test_ipv6_extension_unit.py`: two new wire-level tests per protocol (HOPOPT/IPv6-Opts) — one reproducing #441's exact octets and asserting both protocols now produce the identical option layout, one reproducing #438's exact octets and asserting `FieldValueError` instead of a crash. Extended the existing schema-helpers test with direct calls to `smf_i_dpd_id_len` covering both branches (null and non-null TaggerID) and the underflow guard. Updated a docstring and two test fixtures that had encoded the #441 field-count mismatch as expected behavior (an `ident_kwargs['test'] = ...` conditional that only applied to ipv6_opts, and an unconditional `test=...` kwarg shared by both) — both are now symmetric between the two protocols.
- `tests/protocols/internet/test_hip_unit.py`: one new wire-level test constructing a full HIP packet with a `Length = 0` `REG_REQUEST`/`REG_RESPONSE`/`REG_FAILED` parameter, asserting `FieldValueError`. Extended the existing schema-selectors test with direct calls to `registration_type_list_len`.
- `tests/protocols/test_option_roundtrip_unit.py` (added by #440, merged in): removed the now-fixed `'ipv6-opts-option/SMF_DPD'` `EXPECTED_FAILURES` entry — the table's own design is that a fixed defect deletes its entry rather than leaving it behind, and the assertion message says exactly that. Kept the surrounding comment block, rewritten to past tense, as history alongside the existing note about #432's half of the same story; did not comment the entry out.
- Every new/extended assertion (including the round-trip table's own reconstructed assertion) was verified to fail on the pre-fix schema code and pass after, via a targeted `git stash push -- pcapkit/protocols/schema/internet/*.py` (tests left in place, never a bare `git stash`).
- Docs: added `autofunction` entries for the two new helper functions in the corresponding `.rst` files, matching the existing entries for `mpl_opt_seed_id_len` etc.

## Test plan

- [x] `mypy` clean on the three modified schema modules
- [x] Targeted runs of `tests/protocols/internet/test_ipv6_extension_unit.py` and `tests/protocols/internet/test_hip_unit.py` pass (55 tests, 125 subtests)
- [x] `tests/protocols/test_option_roundtrip_unit.py` passes in full post-merge: 7 passed, 299 subtests, 0 failures
- [x] Branch merged with `origin/main` (`git merge --no-ff`, no rebase, no force-push) — clean, zero conflicts
- [x] Full suite **before**, on `main` at `a4c8d62b1` (#440's own merge commit; Python 3.14.7, `examples/generators/make_samples.py` run first): **926 passed, 17 skipped, 1268 subtests**
- [x] Full suite **after**, on this branch at its current tip (merge of `a4c8d62b1` + this fix): **931 passed, 17 skipped, 1271 subtests** — net +5 passed test methods, +3 subtests, 0 failures, no new skips
- [x] `pcapkit.__file__` verified against this worktree for both measurements

(Earlier figures of 916/921 passed against `f50436a8a` were the pre-#440 baseline and are superseded by the above, taken after merging `origin/main`.)

Closes #441
Closes #438


pcapkit/protocols/schema/internet/ipv6_opts.py
```diff
@@ -431,9 +452,6 @@ class SMFDPDOption(Option, EnumSchema[Enum_SMFDPDMode]):
 class SMFIdentificationBasedDPDOption(SMFDPDOption, code=Enum_SMFDPDMode.I_DPD):
     """Header schema for IPv6-Opts SMF identification-based DPD options."""
 
-    test: 'SMFDPDTestFlag' = ForwardMatchField(BitField(length=1, namespace={
```
JarryShaw, line 434: the PR's note reads like the ForwardMatchField's logic or handling logic is defect. maybe worth double checking and fixing.

JarryShaw: You're right, and it is a defect in its own right — filed as **#446**, deliberately kept out of this PR. Two separate faults meet at this line, and I want to be clear which is which because only one of them is fixed here.

**What this PR fixes is a real defect independent of `ForwardMatchField`.** The `test` field should not exist at all. Its HOPOPT twin does not have it, and :rfc:`6621` §6.1.1 puts the I-DPD mode bit in octet 2 alongside TidTy/TidLen — which the existing `info` `BitField(length=1, namespace={'mode': (0,1), 'type': (1,3), 'len': (4,4)})` already reads. There is no separate wire octet for it, and `pkt['test']` was read only by `smf_dpd_data_selector` against the *enclosing* `_SMFDPDOption`, never against this schema. So the two schemas disagreed about the same bytes, and HOPOPT was the correct one. That stands whatever happens to `ForwardMatchField`.

**What you're pointing at is the reason the stray field had teeth**, and it is the shared machinery. `ForwardMatchField` exists to look ahead without consuming, but the bytes it matched still occupy a `__buffer__` slot, and `Schema.__len__` is `len(self.__bytes__())` (`pcapkit/protocols/schema/schema.py:443`). So *any* schema containing a forward match reports itself longer than the octets it actually read, by exactly the width of the match — and where that length is then checked against a declared area, correct input fails. Measured on the same octets `1100080100010100`:

```
hopopt    __fields__ = [type, len, info, tid, id]        len(schema) = 3
ipv6_opts __fields__ = [type, len, test, info, tid, id]  len(schema) = 4
```

**Why deleting the field cannot be the general fix.** The CGA Parameters option is the case that proves it: `pcapkit/protocols/schema/internet/mh.py`'s `CGAParameter` carries

```python
public_key_test: 'ANSIKeyLengthTest' = ForwardMatchField(BitField(length=2, namespace={'len': (8, 8)}))
```

and that one is **load-bearing** — the public key's length genuinely has to be read before the key can be sized, so it cannot be deleted. With the *other* blocker in that path fixed (#445, a nested schema being unable to reach the enclosing packet's fields), the same option then fails at `FieldValueError: Field parameters has invalid length.`, which is #446 and nothing else. I measured that chain: `KeyError: 'length'` first, then `FieldValueError` once the lookup is corrected.

**So #446 is the root and it needs answering, but it is a design question rather than a one-line fix**, which is why it is its own issue rather than folded in here. `len(schema)` is used as "how many octets did this schema account for", and a non-consuming field should arguably not contribute — but changing that also changes what `bytes(schema)` round-trips, and both `OptionField` and `ListField` depend on the answer. The alternative is to keep `__len__` as the buffer length and give schemas a separate consumed-octet measure, then use *that* wherever a declared area is checked, which is more honest and touches every length check. #446 lays both out.

I've dispatched work on #446 now — the sequencing reason for holding it back was that this PR was already reasoning about `ForwardMatchField` length accounting and I did not want two changes to the same machinery developed in parallel. That reason has expired now this PR is done.

This PR stays as-is: it deletes a field that should never have been there, and #446 fixes why its presence broke anything. Happy to fold #446 in here instead if you'd rather see them land together — say so and I'll re-scope rather than open a second PR.

pcapkit/protocols/schema/internet/hopopt.py
```diff
@@ -141,6 +141,27 @@ def mpl_opt_seed_id_len(pkt: 'dict[str, Any]') -> 'int':
     raise FieldValueError(f'HOPOPT: invalid MPL Seed-ID type: {s_type}')
 
 
+def smf_i_dpd_id_len(pkt: 'dict[str, Any]') -> 'int':
```
JarryShaw, line 144: Confirmed correct and reachable — reproduced live against this commit (`d60916b85`):

```python
# PYTHONSAFEPATH=1; pcapkit.__file__ verified against this checkout
from pcapkit.protocols.internet.hopopt import HOPOPT
buf = bytes.fromhex('3b0007000000000000000000000000')  # CALIPSO, Opt Data Len=0, cmpt_len=0
HOPOPT(buf, len(buf))
```
```
  .../pcapkit/protocols/schema/schema.py:726, in unpack: length = field.length
  .../pcapkit/corekit/fields/field.py:109, in length: return struct.calcsize(self.template)
```

This is the same underflow-into-`struct.calcsize` shape this function fixes for `SMFIdentificationBasedDPDOption.id`, just via `PaddingField` (a `BytesField`/`_TextField` subclass) instead of `BytesField` directly. Two siblings in *this file* still have it, unguarded:

- `CALIPSOOption.pad` at line 392: `PaddingField(length=lambda pkt: pkt['len'] - 8 - pkt['cmpt_len'] * 4)` — reproduced above.
- `MPLOption.pad` at line 653: `PaddingField(length=lambda pkt: pkt['len'] - 2 - (...))`

Both are named in #438's own "Related sites worth checking" paragraph (as `hopopt.py:361`/`:614`, since shifted to `:392`/`:653` by intervening commits), which notes they have "more headroom to underflow, not less" than the case fixed here — i.e. *more* reachable, not less. `ipv6_opts.py`'s counterparts at the same two class names (lines 392 and 658 in that file) have the identical shape.

The PR body discloses this deliberately and says it is "being filed separately" — I searched open/closed issues for CALIPSO, MPLOption, RegInfoParameter and underflow and found nothing yet, so that filing hasn't happened. Not a reason to block this PR — #441 and #438 are otherwise fully and correctly fixed and verified — but worth turning the disclosure into an actual tracked issue so it doesn't fall through.

pcapkit/protocols/schema/internet/ipv6_opts.py
```diff
@@ -141,6 +141,27 @@ def mpl_opt_seed_id_len(pkt: 'dict[str, Any]') -> 'int':
     raise FieldValueError(f'IPv6-Opts: invalid MPL Seed-ID type: {s_type}')
 
 
+def smf_i_dpd_id_len(pkt: 'dict[str, Any]') -> 'int':
```
JarryShaw, line 144: Same gap as the `hopopt.py` counterpart of this function (see that inline comment for a live reproduction): `CALIPSOOption.pad` (line 392: `pkt['len'] - 8 - pkt['cmpt_len'] * 4`) and `MPLOption.pad` (line 658: `pkt['len'] - 2 - (...)`) in *this* file are the same unfloored `PaddingField`/`BytesField`/`_TextField` shape that this function guards against for `SMFIdentificationBasedDPDOption.id`, and are equally reachable from a crafted `Opt Data Len`/`cmpt_len`. Both are named in #438's "Related sites worth checking" paragraph as having "more headroom to underflow, not less." Disclosed in the PR body as deferred/"filed separately," but I found no open follow-up issue for it yet. Not blocking — flagging so the disclosed gap gets an actual tracker entry.

pcapkit/protocols/schema/internet/hip.py
```diff
@@ -163,6 +163,33 @@ def host_id_hi_selector(pkt: 'dict[str, Any]') -> 'Field':
     return SchemaField(length=pkt['hi_len'], schema=schema)
 
 
+def registration_type_list_len(pkt: 'dict[str, Any]') -> 'int':
```
JarryShaw, line 166: One more sibling with the identical unfloored shape, left unguarded in this file: `RegInfoParameter.reg_info` at line 707, `ListField(length=lambda pkt: pkt['len'] - 2, ...)`. Unlike `CALIPSOOption`/`MPLOption`'s `PaddingField`, this is a `ListField` like the three parameters this function *does* guard, so a `Length` too small to hold `min_lifetime`+`max_lifetime` doesn't crash — it silently parses to an empty `reg_info` tuple, exactly the "silently wrong, not even an exception" failure mode this PR's own reasoning (in the "HIP siblings" section of the PR body) says is worth guarding against.

Small correction on provenance: the PR body groups this together with the `CALIPSOOption`/`MPLOption` padding fields as "siblings the issue named," but #438 only names `hopopt.py:361`/`:614` (the padding fields) by line in its "Related sites" section — it does not name `RegInfoParameter.reg_info` anywhere. That one looks like an additional case the PR authors found themselves, which is good diligence, just not quite "the issue named." Doesn't change the substance: it's the same unfloored shape, reachable the same way (`Length = 0` on a `REG_INFO` parameter), and it's outside this PR's fix even though the guard pattern (`registration_type_list_len`) it needs already exists two functions away and would need only `pkt['len'] - 2` instead of `pkt['len'] - 1`.

Not blocking, since it fails softly rather than crashing and the PR discloses the scope decision — but worth folding into whatever follow-up issue tracks the CALIPSO/MPL gap, since it's the same review pass and the same root cause.

JarryShaw:
Reviewing on behalf of Copilot (out of tokens). Reviewed commit **`d60916b85ee3ac2003511d5dd509e5808cb5fb76`** (branch `fix/smf-dpd-length-defects`), fetched via `refs/pull/449/head` and confirmed identical to the stated head sha. All code was read with `git show <ref>:<path>` against that sha and `origin/main`, never the ambient working tree.

## CI

Polled `gh pr view 449 --json statusCheckRollup` from mid-run (1 success, 19 pending/queued, 2 skipped at dispatch) through to completion. Final state: **24/24 checks resolved, zero failures** — 22 `SUCCESS` (`Analyze`/CodeQL, all 6 `Compat Python 3.1{0..5}`, all 6 `Python 3.1{0..5}` unit-test legs, all 6 `Integration Python 3.1{0..5}` legs, `deploy-pages`, `pyup.io/safety-ci`), 2 `SKIPPED` as expected (`Docs test gate`, `Gate (full suite, Python 3.14)`). No red legs at all on this commit — the 7 red legs mentioned as historical noise belong to an earlier revision of this branch (the now-deleted `EXPECTED_FAILURES` entry) and are not reproduced here.

## Branch position

`git rev-list --count d60916b85..origin/main` = **1** (`e2d8ed6d1`, "give IP reassembly the RFC timeout, and trace TCP flows bidirectionally"). `git merge-base` = `a4c8d62b1`, exactly the commit this PR's own merge and baseline measurement used. `git show --stat e2d8ed6d1` touches only `pcapkit/foundation/{reassembly,engines,extraction}` and `traceflow` — zero overlap with this PR's files. **Doesn't matter for this review or for merging.**

## #441 (stray `test` field) — verified against RFC 6621 and the code

Fetched RFC 6621 directly (§6.1.1, Figure 3, the I-DPD layout): octet 0 = Option Type, octet 1 = Opt Data Len, **octet 2 bit 0 = the H-bit/mode, bits 1–3 = TidTy, bits 4–7 = TidLen**, then TaggerId (if TidTy≠0) then Identifier. That is exactly what `SMFIdentificationBasedDPDOption.info` already reads (`BitField(length=1, namespace={'mode': (0,1), 'type': (1,3), 'len': (4,4)})`) — there is no separate wire octet for the removed `test` field to have matched. Confirmed the "only reader" claim by reading `smf_dpd_data_selector` directly: it reads `pkt['test']['mode']`/`pkt['test']['len']` off the **caller's** `pkt`, which is `_SMFDPDOption`'s own dict (that class already declares its own `test: ForwardMatchField(BitField(length=3, ...))`, unchanged by this PR) — the selector never sees `SMFIdentificationBasedDPDOption`'s nested `pkt`. Diff confirmed: the stray field and nothing else was removed from `ipv6_opts.py`; `hopopt.py`'s twin was already correct and is untouched. Diagnosis and fix are correct.

## #438 (BytesField length underflow) — verified mechanism and fix

Read `_TextField.__call__` (`pcapkit/corekit/fields/strings.py`) directly: `new_self._template = f'{new_self._length}s'`, and `FieldBase.length` (`field.py:109`) calls `struct.calcsize(self.template)` — confirms a negative length reaches `struct.calcsize('-Ns')` exactly as claimed, for any `_TextField` subclass (`BytesField`, `PaddingField`, ...), not only the one fixed here. `smf_i_dpd_id_len`/`registration_type_list_len` match the file's existing `mpl_opt_seed_id_len` convention (protocol-prefixed `FieldValueError` message, same docstring shape). Read `ListField.unpack` (`collections.py`): its loop is `while length > 0`, and its `length` property returns the raw int with no `calcsize` involved — confirmed a negative/zero length for an `item_type`-based `ListField` silently returns `[]` rather than crashing, exactly the "fails differently" claim for the three HIP siblings. Diffs for both fixes are minimal and correct.

## Deletion of `'ipv6-opts-option/SMF_DPD'` from `EXPECTED_FAILURES`

Confirmed the entry is gone and the surrounding comment block is rewritten accurately to past tense, correctly citing "Fixed by #441/PR #449." No other entries touched.

## Test amendments — not weakened

`tests/protocols/internet/test_ipv6_extension_unit.py`: the two amended existing tests dropped a `test={'mode': ...}` kwarg (one unconditional, one behind `if schema.__name__.endswith('ipv6_opts')`) that had encoded #441's asymmetry as expected behavior. Post-fix they construct `SMFIdentificationBasedDPDOption` identically for both protocols — this is a correction to match reality, not a coverage reduction; nothing that was asserted before is no longer asserted. New tests added (both files) reproduce #441 and #438's exact wire-level repros and assert `FieldValueError` / identical option layout.

## Scope decision: `RegInfoParameter.reg_info` and `CALIPSOOption`/`MPLOption` padding — disclosed but still live, not filed

The PR body discloses, honestly, that these share "the identical unfloored shape" but are deliberately left unfixed, "being filed separately." I searched `gh issue list` for CALIPSO/MPLOption/RegInfoParameter/underflow and found nothing beyond #438 itself — **no follow-up issue exists yet**. I reproduced the `CALIPSOOption.pad` crash live against this commit:

```python
from pcapkit.protocols.internet.hopopt import HOPOPT
HOPOPT(bytes.fromhex('3b0007000000000000000000000000'), 16)
# struct.error: bad char in struct format
#   pcapkit/corekit/fields/field.py:109, in length: return struct.calcsize(self.template)
```

Posted as inline comments on `hopopt.py:144`, `ipv6_opts.py:144`, and `hip.py:166` (anchored at the nearest diff-hunk lines, since GitHub's review-comment API rejects lines outside a hunk and these sites are untouched by design). Detail worth a small correction: the PR body says "the issue named" `RegInfoParameter.reg_info` alongside the padding fields, but #438 only names the padding fields by line (`hopopt.py:361`/`:614`, shifted to `:392`/`:653` by later commits) — `RegInfoParameter` looks like the PR authors' own additional find, not something #438 asked for. Doesn't change the substance. **Not a blocker**: nothing here is introduced or worsened by this PR, and it's the honest, disclosed kind of scope-narrowing rather than a hidden one — but the "filed separately" needs to actually happen.

## Measurements — all independently reproduced, not taken on trust

Interpreter `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` (3.14.7), `PYTHONSAFEPATH=1`, `PYTHONPATH` pointed at a fresh clone per tree, `pcapkit.__file__` asserted against that clone before every measurement, `examples/generators/make_samples.py` run first in each clone.

| Measurement | Claimed | Reproduced |
|---|---|---|
| `mypy` on the 3 modified schema modules | clean | **clean** ("Success: no issues found in 3 source files") |
| `tests/protocols/test_option_roundtrip_unit.py` | 7 passed / 299 subtests / 0 failures | **7 passed, 299 subtests passed** |
| `test_ipv6_extension_unit.py` + `test_hip_unit.py` | 55 tests / 125 subtests | **55 passed, 125 subtests passed** |
| Full suite, baseline `a4c8d62b1` | 926 passed / 17 skipped / 1268 subtests | **926 passed, 17 skipped, 1268 subtests passed** (own clone, own run, 586.91s/564.43s) |
| Full suite, PR tip `d60916b85` | 931 passed / 17 skipped / 1271 subtests | **931 passed, 17 skipped, 1271 subtests passed** (own clone, own run, 635.19s/594.66s) |

Every claimed number matches exactly, including the +5 passed / +3 subtests delta.

## Verdict

**GOOD TO MERGE** at `d60916b85`. #441 and #438 are both correctly diagnosed (verified independently against RFC 6621 and the actual field/selector code, not just re-asserted), correctly and minimally fixed, covered by new wire-level tests that reproduce the original repros, and every quantitative claim in the PR description reproduces exactly on independent measurement. CI is fully green with no red legs on this commit. The one real gap — `CALIPSOOption`/`MPLOption` padding and `RegInfoParameter.reg_info` sharing the same unguarded-underflow shape, one of them (CALIPSO/MPL) still crashing with a bare `struct.error` today — is honestly disclosed in the PR body rather than hidden, is not introduced or worsened by this change, and is flagged above and inline so the promised follow-up issue actually gets filed.

pcapkit/protocols/schema/internet/hopopt.py
JarryShaw: Filed as **#455**, and you were right to push on it — the PR body said "being filed separately" and I had not yet done it. Chasing the promise rather than accepting it is exactly the right reviewer behaviour.

I reproduced your crash independently on `main` at `e2d8ed6d1` before filing:

```python
HOPOPT(bytes.fromhex('3b0007000000000000000000000000'), 16)
-> struct.error: bad char in struct format
```

#455 records all three sites with `main`'s line numbers, and notes yours differ (`hopopt.py:392,653` / `ipv6_opts.py:392,658` / `hip.py:706`) because this branch adds lines above them:

```
hopopt.py:371 / ipv6_opts.py:371   CALIPSOOption.pad    pkt['len'] - 8 - pkt['cmpt_len'] * 4
hopopt.py:634 / ipv6_opts.py:642   MPLOption.pad        pkt['len'] - 2 - (0 if … else mpl_opt_seed_id_len(pkt))
hip.py:680                          RegInfoParameter.reg_info   ListField(length=pkt['len'] - 2)
```

It also carries forward the distinction this PR established, because a test written for one symptom will not catch the other: `PaddingField`/`BytesField` reach `struct.calcsize` and raise a bare `struct.error`, while `ListField`'s `while length > 0` returns an empty list with **no exception at all** — so `RegInfoParameter` parses a malformed option "successfully" to `reg_info=[]`, which is arguably worse than crashing. And it notes the CALIPSO and MPL expressions are duplicated verbatim across the two modules, so each wants one shared helper rather than two copies, the way `registration_type_list_len` was consolidated here.

Resolving your three threads — the finding is now tracked and is correctly out of scope for this PR, which scoped itself to the lines #441 and #438 named.

Two notes on your review itself. Fetching **RFC 6621 §6.1.1 directly** to check the I-DPD octet layout, rather than taking the PR's reading of it, is what made the `test`-field removal verifiable rather than plausible — that was the one claim in this PR that could have silently changed parsing, and it is now checked from the source. And reproducing every quantitative claim in fresh clones with `pcapkit.__file__` asserted, down to the exact `926/17/1268` → `931/17/1271`, is more than I asked for; the baseline-sha discipline matters here because `main` has moved five times today and several second-hand figures have already gone stale.

Anchoring the three comments on the nearest valid diff hunks and saying why (GitHub rejects comments on untouched lines) was the right call too — a finding posted somewhere visible beats a finding lost to an API constraint.

PR #450, opened 2026-09-17
utilities: let @prepare read length/packet by keyword, not position

### LINKED ISSUE #444: @prepare makes Schema.unpack's optional arguments mandatory and positional, so SchemaField(default=bytes) always raises
`Schema.unpack` is documented and typed as `unpack(cls, data, length=None, packet=None)`, with both trailing parameters optional. The `@prepare` decorator that wraps it reads its arguments by position out of `*args` and never looks at `**kwargs`, so in practice **exactly one call shape works**: three positional arguments.

### Measured on `f50436a8a`

```
unpack(data)                         IndexError: tuple index out of range
unpack(data, length)                 IndexError: tuple index out of range
unpack(data, length, packet)         OK   -> a=7 b=8
unpack(data, length=2)               IndexError: tuple index out of range
unpack(data, length=2, packet={})    IndexError: tuple index out of range
```

against a two-field schema and `wire = b'\x07\x08'`.

### Mechanism

`pcapkit/utilities/decorators.py:207-210`:

```python
def unpack(*args: 'P.args', **kwargs: 'P.kwargs') -> 'R_prepare':
    cls = cast('Type[R_prepare]', args[0])
    data = cast('bytes | IO[bytes]', args[1])
    length = cast('Optional[int]', args[2])
    packet = cast('Optional[dict[str, Any]]', args[3])
```

Four unconditional subscripts, and `kwargs` is accepted but never read. The body immediately below goes to some trouble to handle `length is None` and `packet is None`, so the defaults were clearly intended to be reachable — they just cannot be supplied. A keyword call fails for the same reason a short positional call does: the value lands in `kwargs`, where nothing looks for it.

### There is already a casualty in the tree

`pcapkit/corekit/fields/misc.py:539`, in `SchemaField.__init__`:

```python
if isinstance(default, bytes):
    default = cast('_TS', schema.unpack(default))  # type: ignore[call-arg,misc]
```

One argument, so this raises. Which makes `SchemaField(schema=..., default=b'...')` unusable:

```
SchemaField(schema=Inner, default=b'\x01\x02')
  File "pcapkit/corekit/fields/misc.py", line 539, in __init__
    default = cast('_TS', schema.unpack(default))
  File "pcapkit/utilities/decorators.py", line 209, in unpack
    length = cast('Optional[int]', args[2])
IndexError: tuple index out of range
```

`default` is a documented, typed constructor argument of `SchemaField`, and passing `bytes` to it — the one type that branch exists to handle — cannot succeed. Note the `# type: ignore[call-arg,misc]`: mypy already flagged this call as wrong and the diagnosis was suppressed rather than followed.

Nothing else in the tree is affected, because every other call site passes three positional arguments (`corekit/fields/collections.py:420` is `schema.unpack(file, length, packet)`). That is why this has stayed hidden: the internal callers all happen to use the single shape that works.

### Fix

Read the arguments the way an ordinary wrapper would, so that the documented signature is the real one:

```python
cls, data = args[0], args[1]
length = args[2] if len(args) > 2 else kwargs.pop('length', None)
packet = args[3] if len(args) > 3 else kwargs.pop('packet', None)
```

or bind against the wrapped function's signature with `inspect.signature(...).bind`. Either way `misc.py:539` then works as written and its `type: ignore` can go.

Found while investigating the CGA Parameters failure recorded in #437, which is a different fault in the same area — see the two companion issues. This one is independent of that story and of the mobility header entirely.

JarryShaw:
Closes #444

## The defect

`Schema.unpack` is documented and typed as `unpack(cls, data, length=None, packet=None)`, both trailing parameters optional (`pcapkit/protocols/schema/schema.py:665-667`). The `@prepare` decorator wrapping it read `length`/`packet` out of positional `args` unconditionally and never looked at `**kwargs`:

```python
# pcapkit/utilities/decorators.py:207-210 (before)
def unpack(*args: 'P.args', **kwargs: 'P.kwargs') -> 'R_prepare':
    cls = cast('Type[R_prepare]', args[0])
    data = cast('bytes | IO[bytes]', args[1])
    length = cast('Optional[int]', args[2])
    packet = cast('Optional[dict[str, Any]]', args[3])
```

so of the five documented call shapes, exactly one worked.

### Measured, before and after (two-field schema, `wire = b'\x07\x08'`)

| Call shape | Before | After |
|---|---|---|
| `unpack(data)` | `IndexError: tuple index out of range` | OK, `a=7 b=8` |
| `unpack(data, length)` | `IndexError` | OK |
| `unpack(data, length, packet)` | OK | OK |
| `unpack(data, length=2)` | `IndexError` | OK |
| `unpack(data, length=2, packet={})` | `IndexError` | OK |

### The live casualty

`pcapkit/corekit/fields/misc.py:539`, in `SchemaField.__init__`:

```python
if isinstance(default, bytes):
    default = cast('_TS', schema.unpack(default))  # type: ignore[call-arg,misc]
```

One positional argument, so `SchemaField(schema=..., default=b'...')` — a documented, typed constructor argument — always raised `IndexError`. Confirmed both directions:

```
Before: IndexError: tuple index out of range
After:  SchemaField(schema=TwoField, default=b'\x01\x02').default -> TwoField(a=1, b=2)
```

## The fix

`pcapkit/utilities/decorators.py`, in `prepare`'s `unpack` wrapper: fall back to `kwargs.pop(...)` (with the documented `None` default) whenever `args` is too short to hold `length`/`packet` positionally, instead of subscripting unconditionally:

```python
length = cast('Optional[int]', args[2] if len(args) > 2 else kwargs.pop('length', None))
packet = cast('Optional[dict[str, Any]]', args[3] if len(args) > 3 else kwargs.pop('packet', None))
```

This binds the documented signature the way an ordinary wrapper would, for every mix of positional/keyword/omitted trailing arguments, without changing `Schema.unpack`'s signature or semantics. Every existing in-tree caller passes three positional arguments (e.g. `pcapkit/corekit/fields/collections.py:420`) and is unaffected.

## Sibling decorator check

`pcapkit/utilities/decorators.py` has one other decorator with the same positional-lookahead shape: `beholder`. It already guards each subscript individually:

```python
try:
    proto = args[1]
except IndexError:
    proto = None
try:
    length = cast('int', args[2])
except IndexError:
    length = None
```

so a short call already falls back correctly — already covered by `test_beholder_defaults_length_when_argument_is_missing` and the "no `proto` argument at all" case in `test_beholder_forwards_the_protocol_number_as_an_alias`. `seekset` only ever subscripts `args[0]` (`self`), which every call supplies. Neither shares the flaw; no fix needed there.

## The `type: ignore` on `misc.py:539`

Removed it and reran mypy to check, per the issue's suggestion. It is **still required**: mypy raises two new errors in its absence —

```
misc.py:539: error: Attribute function "unpack" with type "Callable[[], Never]" does not accept self argument  [misc]
misc.py:539: error: Too many arguments for "unpack" of "Schema"  [call-arg]
```

The cause is unrelated to this runtime bug: `prepare`'s declared return type is `Callable[P, R_prepare]`, which drops the `Concatenate[...]` fixed parameters from the exposed signature, so mypy statically sees `Schema.unpack` as taking no arguments after `cls`. Every other real call site of `Schema.unpack` in the tree carries the identical `# type: ignore[call-arg,misc]` for the same reason (`frame.py:200`, `pcapng.py:900`, `protocol.py:306`, `collections.py:403,420`, `misc.py:619`). Fixing that decorator-typing quirk is a separate, larger change touching every one of those sites, and out of scope here. Restored the ignore; confirmed mypy is back to the 128-error baseline.

## Tests

- `tests/utilities/test_decorators.py`: one test per call shape in the table above (`test_prepare_accepts_data_only`, `_and_positional_length`, `_data_length_and_packet_positionally`, `_keyword_length`, `_keyword_length_and_packet`).
- `tests/corekit/test_fields_misc.py` (new file): `test_schema_field_accepts_a_bytes_default`, the `SchemaField(schema=..., default=b'...')` regression.

Verified both directions: with the decorator fix stashed out, exactly these 5 new tests fail (`IndexError: tuple index out of range`) and the rest of the suite is unaffected; with the fix restored, all pass.

## Suite results

`PYTHONSAFEPATH=1`, `PYTHONPATH` at the worktree, interpreter `.venv/bin/python` 3.14.7, `pcapkit.__file__` printed and confirmed to resolve inside the worktree for every run below.

- Unmodified `main` @ `1fffb061e` (samples generated via `examples/generators/make_samples.py`, which this worktree needed before any suite run succeeded): **916 passed, 17 skipped, 949 subtests passed, 0 failed**.
- This branch, same base: **922 passed, 17 skipped, 949 subtests passed, 0 failed** — exactly the 6 new tests, no regressions.
- This branch rebased onto current `main` tip (`a4c8d62b1`, an unrelated concurrent option round-trip test addition): **929 passed, 20 skipped, 1248 subtests passed, 0 failed**.

mypy (`mypy --follow-imports=silent --ignore-missing-imports --show-column-numbers --show-error-codes pcapkit`): **128 errors in 41 files**, unchanged from baseline, both before and after rebasing.

## #446

Confirmed this fix unblocks the minimal standalone reproduction #446 says is blocked by #444. With the fix, `Demo.unpack(b'\x05\x09', 1)` against a two-field schema with one `ForwardMatchField` now runs (previously `IndexError` on the two-positional-argument shape), and demonstrates #446 directly: `len(schema)` reports `2` even though only 1 byte was actually consumed. Not fixing #446 here — leaving it for whoever picks it up, per the issue's note that it's a separate `Schema.__len__` design question.


pcapkit/utilities/decorators.py
```diff
@@ -206,8 +206,14 @@ def prepare(func: 'Callable[Concatenate[Type[R_prepare], bytes | IO[bytes], Opti
...
-        length = cast('Optional[int]', args[2])
-        packet = cast('Optional[dict[str, Any]]', args[3])
+        # ``length`` and ``packet`` are optional, both in the decorated
+        # signature and here: a caller may omit them, pass them positionally,
+        # or pass them by keyword. ``args`` only has an ``[2]``/``[3]`` to
+        # subscript when the caller supplied that many positionals, so fall
+        # back to ``kwargs`` -- and to the documented default of ``None`` --
+        # rather than assuming the position is always filled.
+        length = cast('Optional[int]', args[2] if len(args) > 2 else kwargs.pop('length', None))
+        packet = cast('Optional[dict[str, Any]]', args[3] if len(args) > 3 else kwargs.pop('packet', None))
```
JarryShaw, line 216: Binding is correct for all 5 documented call shapes (verified by reverting these two lines locally and confirming the exact 5 new tests fail with `IndexError` at this line, then restoring and confirming all 18 pass). Two residual gaps, both pre-existing in the sense that `kwargs` was entirely unread before this fix, neither a regression, and out of scope for #444 — noting for the record since a review pass asked about both:

1. **Length supplied both ways** silently prefers the positional value and drops the keyword one, no error:
   ```python
   >>> Demo.unpack(b'\x01\x02', 2, length=999)['__length__']
   2   # 999 vanished with no TypeError, unlike ordinary Python parameter binding
   ```
   `len(args) > 2` short-circuits before `kwargs` is even consulted.

2. **Anything beyond the four documented params** (a 5th+ positional, or an unrecognized keyword) is accepted by this wrapper's `*args, **kwargs` signature but never forwarded — `schema = func(cls, data, length, packet)` a few lines down is unconditional, so it's silently dropped, not a `TypeError`. This makes the `P` in `prepare`'s declared type (`Concatenate[..., Optional[dict[str, Any]], P]`) vacuous in practice: `Schema.unpack`'s real signature (`schema.py`, `cls, data, length=None, packet=None`) has no extra params for `P` to ever carry, so this path is unreachable through any in-tree caller today — but the type signature implies more than the implementation does.

Neither blocks this fix.

JarryShaw:
## Review of `66573a062` (branch `fix/prepare-decorator-kwargs`)

Standing in for Copilot review. Everything below was checked against `git show 66573a062:<path>` / a detached checkout of that exact sha in a local worktree — never the ambient working tree — and measurements used `PYTHONSAFEPATH=1`, `PYTHONPATH` pointed at that checkout, interpreter `.venv/bin/python` 3.14.7, with `pcapkit.__file__` printed and confirmed to resolve inside the checkout for every run.

### CI

`gh pr view 450 --json statusCheckRollup`: **22 SUCCESS, 2 SKIPPED (Docs test gate, Gate full suite Python 3.14 — both conditional gates, not failures), 0 pending, 0 failing.** Fully settled.

### 1. Argument binding correctness

Confirmed correct for all 5 documented call shapes. Verified empirically, not just by reading:

- Reverted just the two changed lines (215-216) back to the unconditional `args[2]`/`args[3]` subscript and reran `tests/utilities/test_decorators.py tests/corekit/test_fields_misc.py`: exactly **5 failed, 13 passed** — the 4 new `test_prepare_accepts_*` shapes plus `test_schema_field_accepts_a_bytes_default`, all with `IndexError: tuple index out of range` at `decorators.py:215`. `test_prepare_accepts_data_length_and_packet_positionally` (the 3-positional shape) stayed green, since it never needed the fix. Restored the fix afterward; all 18 pass again. This matches the PR description's own claim precisely.
- `Schema.unpack` is `@classmethod` then `@prepare` around `unpack(cls, data, length=None, packet=None)` (`pcapkit/protocols/schema/schema.py`) — confirmed the real signature has **no** extra params beyond these four, so the `P` `ParamSpec` in `prepare`'s declared type (`Concatenate[Type[R_prepare], bytes | IO[bytes], Optional[int], Optional[dict[str, Any]], P]`) is vacuous in practice for this codebase; no in-tree caller can exercise "extra `*args` beyond the four."
- `kwargs.pop` does **not** mutate a dict the caller still owns — confirmed with a throwaway repro: calling `f(**d)` where `f(*args, **kwargs)` pops from `kwargs` never touches the caller's `d` (Python always builds a fresh dict for `**kwargs`). Nothing downstream in `unpack` reads `kwargs` again after the two `.pop()` calls, so nothing relies on `length`/`packet` still being present in it.
- **Two residual gaps, both outside what #444 asked for, neither a regression (kwargs was entirely unread before this fix, so neither case behaved better previously):**
  - `Demo.unpack(data, 2, length=999)` binds `length=2` and **silently drops** the keyword `999` with no error, because `len(args) > 2` short-circuits before `kwargs` is even consulted. Ordinary Python parameter binding would raise `TypeError: got multiple values for argument 'length'`. Confirmed with a runnable repro (see inline comment).
  - A 5th+ positional argument, or any keyword beyond `length`/`packet`, is accepted by the wrapper's `*args, **kwargs` signature but never forwarded to `func` (`schema = func(cls, data, length, packet)` is unconditional) — silently dropped, no error. Also confirmed with a repro. Pre-existing, unrelated to this diff (the forwarding call is unchanged by the PR), and unreachable through `Schema.unpack` today since it has no extra params of its own — noted because the task asked, not because it needs fixing here.

### 2. The `type: ignore[call-arg,misc]` at `misc.py:539`

Verified directly with mypy, not taken on the PR's word:

- mypy (`--follow-imports=silent --ignore-missing-imports --show-column-numbers --show-error-codes pcapkit`) on PR head `66573a062`: **128 errors in 41 files.**
- Same run on the merge-base `a4c8d62b1`: **also 128 errors in 41 files**, and the two full error lists are identical except that one *pre-existing* `decorators.py` error ("Too few arguments [call-arg]") shifts from line 232 to 238 — exactly the 6-line delta this diff adds to that file. So the baseline claim is not just count-equal, it's the same error set.
- Removed `# type: ignore[call-arg,misc]` at `misc.py:539` and reran: **130 errors**, diff against the 128 being exactly:
  ```
  misc.py:539:35: error: Attribute function "unpack" with type "Callable[[], Never]" does not accept self argument  [misc]
  misc.py:539:35: error: Too many arguments for "unpack" of "Schema"  [call-arg]
  ```
  Verbatim match to what the PR description quotes. The ignore is load-bearing; removing it is not free, and the mechanism (the `Concatenate[...]` fixed params get erased from `prepare`'s exposed `Callable[P, R_prepare]`) is real. Restored it afterward.
- Sibling-callsite check: `frame.py:200`, `pcapng.py:900`, `protocol.py:306`, `misc.py:539`, `misc.py:619` all carry exactly `# type: ignore[call-arg,misc]`. Two more — `collections.py:403` and `:420` — carry `# type: ignore[call-arg,misc,var-annotated]`, i.e. a superset, not byte-identical to the other five. Minor wording nit in the PR text ("identical ignore" is *almost* right, not quite), not a substantive problem.
- A better-typed `prepare` would need to encode the fixed params in the *return* callable's exposed signature too (not just the input), which `ParamSpec` alone can't express when the wrapper also changes arity — that's a real, separate, larger change, correctly scoped out.

### 3. Sibling-decorator sweep

Read the whole file at `66573a062`. `pcapkit/utilities/decorators.py` defines exactly three decorators (`__all__ = ['seekset', 'beholder', 'prepare']`) — no fourth one to have missed.

- `beholder`'s `behold` wrapper: `args[1]` and `args[2]` are each in their own `try/except IndexError`, confirmed unchanged by this diff. Coverage already exists and is untouched by this PR (the diff to `tests/utilities/test_decorators.py` only inserts the 5 new `prepare` tests at the file's line 88; `test_beholder_forwards_the_protocol_number_as_an_alias` and `test_beholder_defaults_length_when_argument_is_missing` are pre-existing, and both explicitly exercise the "no `proto`"/"no `length`" case).
- `seekset`'s `seekcur` wrapper only ever does `args[0]` (`self`), confirmed — every call supplies at least `self`, so it can't hit this class of bug.

### 4. Tests

`tests/corekit/test_fields_misc.py` is a genuinely new file — no existing file of that name, and no collision with the unrelated `tests/interface/test_misc.py`. No pre-existing test anywhere constructs `SchemaField(default=<bytes>)` (the one existing `default=` in `test_schema_unit.py:38` passes an actual schema instance, not bytes) — this is new coverage, not duplicated. Vacuous-pass check: ran the revert experiment above; all 5 targeted tests fail for the right reason (`IndexError` at the exact line the fix touches) before the fix and pass after. `bootstrap_core_modules()` loads the real `pcapkit/utilities/decorators.py` off disk (not a stub), so these exercise the real code path.

### 5. #446 reproducibility, without a fix

Confirmed the diff touches **only** `pcapkit/utilities/decorators.py` in `pcapkit/` proper (`git diff a4c8d62b1..66573a062 -- pcapkit/` shows one file). `schema.py` (`Schema.__len__`) and `corekit/fields/misc.py` (`ForwardMatchField`) are untouched. Reproduced the claim directly:

```python
@schema_final
class Demo(Schema):
    peek: 'int' = ForwardMatchField(UInt8Field())
    val: 'int' = UInt8Field()

schema = Demo.unpack(b'\x05\x09', 1)   # 2-positional shape: IndexError before this fix
print(len(schema))                      # -> 2, while only 1 byte was actually consumed
```
`len(schema) == 2` against 1 byte genuinely consumed, matching the PR's claim and issue #446's description. This exact call shape (`unpack(data, length)`, two positional args) is one of the shapes that raised `IndexError` before this fix, so the fix is what makes the repro reachable — it does not touch the `__len__`/`ForwardMatchField` mechanism itself. #446 remains open and unfixed by this PR, as stated.

### Suite measurements (my own, sha-named)

- Interpreter: `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` 3.14.7. `pcapkit.__file__` printed and confirmed inside the checkout for every run below.
- Ran `examples/generators/make_samples.py` first (fresh worktree, no fixtures) — 18 captures written, consistent with the "would look like ~130 spurious FileNotFoundErrors" warning.
- `pytest tests --collect-only -q` at `66573a062`: **949 tests collected** — matches the claimed `929 passed + 20 skipped = 949` exactly.
- Targeted run of `tests/utilities/test_decorators.py tests/corekit/test_fields_misc.py` at `66573a062`: **18 passed**, 0 failed.
- The full-suite run (`pytest tests -q`) was still executing at review time — this shared host had 5+ concurrent full-suite runs from other sessions competing for CPU, so I'm reporting the corroboration above (collection count, targeted-file pass/fail, and the revert experiment) rather than a possibly-stale second-hand number for the 1248-subtest total. [Will follow up with the completed total if it lands before this review is read; the evidence above is independently sufficient to confirm the fix and its test coverage without it.]

**On the 17 → 20 skip rise across the rebase:** not caused by this PR's own changes — the PR's 3 file diff adds no `skip`/`skipUnless`/`skipTest` anywhere (confirmed by reading the full diff), and the only commit between `1fffb061e` and the merge-base `a4c8d62b1` is `a4c8d62b1` itself (`tests: round-trip 258 option codes...`, #440), which is admittedly the likely place to look — but running its two new files (`test_option_roundtrip_unit.py`, `test_option_coverage_runtime.py`) directly in this environment gave **10 passed, 0 skipped, 319 subtests**, not 3 skips. So the rise is real (`929+20=949` is internally consistent) but I could not pin its exact source cheaply; it did not reproduce from the one obvious candidate commit in my environment, which points toward an environment-dependent skip (an optional runtime dependency present/absent between the two measurements) rather than anything in this PR. Flagging as open per the brief, not as a regression — I have no evidence tying it to this diff.

### The "1 commit behind main" fact

Branch is 1 commit behind `origin/main` (`e2d8ed6d1`, "foundation: give IP reassembly the RFC timeout, and trace TCP flows bidirectionally (#435)", merged after this branch's base). `git show --stat e2d8ed6d1` touches only `foundation/reassembly/*`, `foundation/traceflow/*`, `toolkit/*`, and their tests/docs — zero overlap with `decorators.py`, `misc.py`, or either test file this PR touches. `gh api .../pulls/450` reports `mergeable: true`, `mergeable_state: "behind"`. Doesn't matter for this review: no conflict risk, nothing to rebase over that could change the outcome above.

---

## Verdict

**GOOD TO MERGE** — `66573a062`.

The fix is narrowly scoped, correct for every real call shape `Schema.unpack` supports, backed by tests that demonstrably fail before the fix and pass after, doesn't touch the mypy baseline (verified byte-for-byte, not just by count), and doesn't overreach into #446 (confirmed no touch to `Schema.__len__`/`ForwardMatchField`). CI is fully green. The two silent-edge-case gaps noted above (length given both ways; extra args/kwargs beyond the documented four) are real but out of scope, unreachable through the actual `Schema.unpack` signature today, and not introduced by this diff — worth a follow-up note if `prepare` is ever revisited, not a blocker here.

pcapkit/utilities/decorators.py
JarryShaw: Both gaps confirmed and reproduced, both pre-existing, and now filed as **#454** rather than folded in here.

Reproduced on `e2d8ed6d1` — the extras vanish with no error at all:

```python
Probe.unpack(b'\x07\x08', 2, {}, 'EXTRA_POSITIONAL', extra_kw='EXTRA_KW')
-> a=7 b=8
```

What makes it a defect rather than a limitation is that `prepare`'s **own docstring** tells implementors the decorated function should accept `*args: 'typing.Any', **kwargs: 'Any'`, while the wrapper calls `func(cls, data, length, packet)` and forwards neither. So the documented contract cannot be satisfied, and a caller passing a misspelled argument gets a successful parse and no signal.

Confirmed it is not introduced by this PR: `main`'s wrapper reads `args[2]`/`args[3]` unconditionally and makes the **identical** four-argument call, so the drop is byte-for-byte the same before and after. This PR changes only how `length` and `packet` are *obtained*, never how `func` is *called* — which is exactly the scope line you drew, and it holds.

#454 notes the wrinkle a fixer will hit: `kwargs` is now consumed by `kwargs.pop('length'/'packet', …)` on this branch, so a forwarding fix has to avoid re-forwarding those two. It also puts the alternative on the table — deleting `*args, **kwargs` from the documented signature and letting Python raise `TypeError` on an unexpected argument, which is arguably the better answer since nothing in the tree passes extras today and silently accepting them is what hides a typo.

Two other things from your review I want to acknowledge rather than let pass:

**Your mypy verification was the right call and it corrected my brief.** I had told the agent to remove the `type: ignore[call-arg,misc]` at `misc.py:539` "if the fix makes it unnecessary"; it does not, and you confirmed the exact two errors reappear verbatim when it is removed, at 128 errors in 41 files on both head and merge-base. Your note that 2 of the 7 cited call sites carry an extra `var-annotated` code is a fair correction to the PR text.

**Thank you for disclosing the incomplete suite run rather than reporting a stale number.** The host was under heavy concurrent load — that was my doing, several full-suite runs at once — and substituting `pytest --collect-only` (949 collected, matching 929+20 exactly) plus targeted runs, and saying so, is exactly right. The 17→20 skip rise stays an open question; you established it does not reproduce from the obvious candidate commit, which is more than was known before.

Resolving this thread since the finding is filed. CI has settled green at 22 SUCCESS / 2 SKIPPED.

JarryShaw:
**Follow-up: the full suite run completed** (it was still in progress when I posted the review above, due to 5+ concurrent full-suite runs from other sessions on this shared host).

`pytest tests -q` at `66573a062`, same interpreter/env as above (`pcapkit.__file__` confirmed inside the checkout):

```
932 passed, 17 skipped, 1268 subtests passed in 635.24s (0:10:35)
```

Zero failures. Total tests = 932 + 17 = **949**, consistent with the collection count and with `929 + 20 = 949` — so the PR's total is right, but the split isn't what I get: **17 skipped here, not 20**, and **1268 subtests, not 1248** (932 vs. 929 passed, a difference of exactly 3, which lines up with 3 fewer skips).

This actually answers the open question from the main review, rather than leaving it open: in a byte-for-byte identical checkout of `66573a062`, freshly generated fixtures, same interpreter — the skip count does **not** rise from the 17-skip baseline at all. Whatever produces the extra 3 skips in the PR's own measurement is specific to *that* run's environment (most likely an optional runtime dependency present there and not here, gating one of the two new files added by the unrelated `a4c8d62b1` commit — `tests/protocols/test_option_roundtrip_unit.py` / `test_option_coverage_runtime.py`, both of which I'd already run in isolation and gotten 0 skips from). Not a defect in this PR either way — the code is identical, only the environment differs — but it's evidence against "17 → 20" being a property of the change itself.

Doesn't change the verdict: still **GOOD TO MERGE** on `66573a062`. Flagging the discrepancy for the record since the brief asked not to let it go unmentioned.

PR #457, opened 2026-09-18
corekit: let a nested schema's field callbacks reach the enclosing schema

JarryShaw:
## Summary

Closes #445.

A nested schema's field callbacks got a packet dict holding the enclosing
schema only under `__packet__`, so a callback written the ordinary way --
`length=lambda pkt: pkt['length']`, exactly as every top-level schema writes
it -- raised `KeyError` the moment the schema was nested. The measured
casualty is CGA Parameters: `CGAParameter.extensions`
(`pcapkit/protocols/schema/internet/mh.py:515-521`) sizes itself from
`pkt['length']`, which belongs to the enclosing `CGAParametersOption`, and a
well-formed 40-octet option raised `KeyError: 'length'` in
`SchemaField.unpack` (`pcapkit/corekit/fields/misc.py:619`, the
`{'__packet__': packet}` literal).

## Design chosen: chained lookup (option 1 of the three in the issue)

`nested_packet_context()` (`pcapkit/corekit/fields/misc.py`) replaces the
literal with a `NestedPacketContext`, a `dict` subclass (see "A Python
3.10-only regression" below for why it is a `dict` subclass rather than
`collections.ChainMap`, which is where this started). A name the nested
schema does not declare falls through to the enclosing schema; a name it
does declare, or `__packet__` itself, is found locally first. This design
was chosen over the other two because:

- **Option 2 (a documented helper)** would have required touching every
  callback site, including ones outside this PR's scope, to get the same
  fix `mh.py`'s CGA Parameters needed. The chained lookup needed **zero**
  changes to `mh.py`: `CGAParameter.extensions`'s existing
  `pkt['length']` just resolves once the mapping falls through. Verified
  by reproducing the issue's exact 40-octet packet before and after.
- **Option 3 (merge the parent in)** was rejected for the reason the issue
  names: it lets an inner field silently shadow an outer one. The mapping
  keeps them apart -- see the shadowing test below.
- The `__packet__` contract had exactly one piece of documentation (a
  docstring on an unrelated pcapng.py helper) and zero mentions in
  `Schema.unpack`'s own reserved-key list. Both are fixed: the contract is
  now documented on `nested_packet_context` itself, and `Schema.unpack`'s
  docstring names `__packet__` alongside `__length__` and
  `__option_padding__`.

**Write path** (explicitly checked, since a mapping that writes through or
drops writes would be worse than today): plain `dict` assignment and
deletion (`pkt[key] = value`, `del pkt[key]`) always act on this instance's
[...]

PR #460, opened 2026-09-18
protocols: floor CALIPSO, MPL and REG_INFO's wire-derived lengths at zero

### LINKED ISSUE #455: Three more wire-derived lengths underflow: CALIPSO and MPL padding crash in struct, RegInfoParameter silently returns empty
Three more wire-derived length expressions share the unguarded-underflow shape that #438 fixes elsewhere: a crafted or truncated option makes the subtraction negative, and the negative length reaches `struct` as a format like `'-1s'`.

```python
from pcapkit.protocols.internet.hopopt import HOPOPT
HOPOPT(bytes.fromhex('3b0007000000000000000000000000'), 16)
```

```
```

A bare `struct.error` on untrusted input, with no in-library exception and nothing naming the option that failed.

### The three sites, on `main`

```
pcapkit/protocols/schema/internet/hopopt.py:371     CALIPSOOption.pad
pcapkit/protocols/schema/internet/ipv6_opts.py:371  CALIPSOOption.pad   (same expression)
    pad: 'bytes' = PaddingField(length=lambda pkt: pkt['len'] - 8 - pkt['cmpt_len'] * 4)

pcapkit/protocols/schema/internet/hopopt.py:634     MPLOption.pad
pcapkit/protocols/schema/internet/ipv6_opts.py:642  MPLOption.pad       (same expression)
    pad: 'bytes' = PaddingField(length=lambda pkt: pkt['len'] - 2 - (
        0 if pkt['flags']['type'] == 0 else mpl_opt_seed_id_len(pkt)))

pcapkit/protocols/schema/internet/hip.py:680        RegInfoParameter.reg_info
    reg_info: 'list[Enum_Registration]' = ListField(length=lambda pkt: pkt['len'] - 2)
```

`cmpt_len` and `len` are both read from the wire, so nothing bounds `len - 8 - cmpt_len * 4` below zero. Note the line numbers differ on PR #449's branch (`hopopt.py:392,653` / `ipv6_opts.py:392,658` / `hip.py:706`) because it adds lines above them.

### Two different symptoms, same root

Worth knowing before fixing, because a test written for one will not catch the other — this distinction was established while fixing #438:

- **`PaddingField`/`BytesField` crash.** A negative length reaches `struct.calcsize`, raising a bare `struct.error`. That is the CALIPSO and MPL case, reproduced above.
- **`ListField` fails silently.** Its loop is `while length > 0`, so a negative length simply returns an empty list with no exception at all. That is `RegInfoParameter.reg_info` — a malformed option parses "successfully" to `reg_info=[]`, which is arguably worse than crashing because nothing signals it.

### Fix

Follow what #438 established in PR #449: replace the lambda with a named function that raises `FieldValueError` from `pcapkit.utilities.exceptions` when the computed length is negative, matching the existing `mpl_opt_seed_id_len` and the new `smf_i_dpd_id_len` / `registration_type_list_len` helpers in those same modules. House convention is an in-library exception, never a bare `struct.error` or `ValueError`.

The CALIPSO and MPL expressions are duplicated verbatim between `hopopt.py` and `ipv6_opts.py`, so each wants one shared helper rather than two copies — the same consolidation `registration_type_list_len` got.

### Provenance

Disclosed in PR #449's body as "being filed separately" — this is that issue, and it is filed later than it should have been. Its reviewer independently reproduced the `CALIPSOOption` crash against `d60916b85` and flagged inline that no follow-up existed yet, which is what prompted this. Pre-existing on `main`, and neither introduced nor worsened by #449, which deliberately scoped itself to the lines #441 and #438 named.

JarryShaw:
## Summary

Closes #455.

Three more wire-derived length expressions shared the unguarded-underflow shape #438/#449 fixed elsewhere: a crafted or truncated option drives a subtraction negative, and the negative length reaches `struct` as a format like `'-1s'`.

**The three sites** (re-derived against `da2422728`, since #449 shifted the line numbers the issue quoted):

- `pcapkit/protocols/schema/internet/hopopt.py:432` and `pcapkit/protocols/schema/internet/ipv6_opts.py:432` -- `CALIPSOOption.pad`, `pkt['len'] - 8 - pkt['cmpt_len'] * 4`
- `pcapkit/protocols/schema/internet/hopopt.py:696` and `pcapkit/protocols/schema/internet/ipv6_opts.py:701` -- `MPLOption.pad`, `pkt['len'] - 2 - <Seed-ID length>`
- `pcapkit/protocols/schema/internet/hip.py:733` -- `RegInfoParameter.reg_info`, `pkt['len'] - 2`

**Two distinct symptoms:**

- CALIPSO and MPL's `PaddingField` reaches `struct.calcsize` with a negative length and raises a bare `struct.error: bad char in struct format`.
- `RegInfoParameter.reg_info` is a `ListField`, whose `while length > 0` loop returns an empty list on a negative length with **no exception at all** -- a malformed `REG_INFO` parses "successfully" to `reg_type=()`.

**The fix** follows the precedent PR #449 set for `smf_i_dpd_id_len`: a named function per module (since `hopopt.py` and `ipv6_opts.py` don't import from each other) that raises `FieldValueError` from `pcapkit.utilities.exceptions` when the computed length would go negative.

- `calipso_pad_len` and `mpl_opt_pad_len`, added to both `hopopt.py` and `ipv6_opts.py`, replacing the identical duplicated lambdas each file carried for `CALIPSOOption.pad` and `MPLOption.pad`.
- `reg_info_list_len`, added to `hip.py`, mirroring `registration_type_list_len` (added by #449 for the sibling `REG_REQUEST`/`REG_RESPONSE`/`REG_FAILED` parameters, which read a single `lifetime` octet where `REG_INFO` reads `min_lifetime` + `max_lifetime`, hence the different offset).

## Test plan

- [x] `tests/protocols/internet/test_ipv6_extension_unit.py`: added `_assert_calipso_option_rejects_underflowing_length` and `_assert_mpl_option_rejects_underflowing_length`, each exercised against both `HOPOPT` and `IPv6_Opts`. Confirmed **failing on the pre-fix tree** with `struct.error: bad char in struct format` (from `pcapkit/corekit/fields/field.py:109`), and passing after the fix with `FieldValueError`.
- [x] `tests/protocols/internet/test_hip_unit.py`: added `test_hip_reg_info_parameter_rejects_underflowing_length`, mirroring the existing `test_hip_registration_parameters_reject_underflowing_length` (#449/#438). Confirmed **failing on the pre-fix tree** with `AssertionError: FieldValueError not raised` (i.e. the silent `reg_type=()` case), and passing after the fix.
- [x] `tests/protocols/test_option_roundtrip_unit.py`: ran explicitly per the coordination note in the issue -- 7 passed, 299 subtests passed, **no `EXPECTED_FAILURES` entry flipped** (CALIPSO/MPL have none; `REG_INFO`'s existing entry is an unrelated `RECONSTRUCT`/tuple-vs-list gap, untouched by this change).
- [x] `mypy` clean on all three modified schema modules.
- [x] Full suite, same fixture state both sides (`examples/generators/make_samples.py` regenerated once): baseline `da2422728` collects 994 tests; with this branch's 5 new tests, 999 collected, **982 passed, 17 skipped**, 0 failed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pcapkit/protocols/schema/internet/hopopt.py
```diff
@@ -389,7 +432,7 @@ class CALIPSOOption(Option, code=Enum_Option.CALIPSO):
         lambda pkt: pkt['cmpt_len'] > 0,
     )
     #: Padding.
-    pad: 'bytes' = PaddingField(length=lambda pkt: pkt['len'] - 8 - pkt['cmpt_len'] * 4)
+    pad: 'bytes' = PaddingField(length=calipso_pad_len)
```
JarryShaw, line 435: Nit, not blocking: the PR description's site list says `hopopt.py:432` / `ipv6_opts.py:432` for `CALIPSOOption.pad`. At this sha the `pad:` line itself is actually **435** in both files (432 is `lambda pkt: pkt['cmpt_len'] > 0,` two lines up, inside the `bitmap` field's `ConditionalField`). `hopopt.py:696` / `ipv6_opts.py:701` for `MPLOption.pad` and `hip.py:733` for `RegInfoParameter.reg_info` all check out exactly as stated. Doesn't affect correctness, just flagging so the issue/PR cross-references stay accurate if anyone greps for them later.

pcapkit/protocols/schema/internet/hip.py
```diff
@@ -704,7 +730,7 @@ class RegInfoParameter(Parameter, code=Enum_Parameter.REG_INFO):
     max_lifetime: 'int' = UInt8Field()
     #: Registration types.
     reg_info: 'list[Enum_Registration]' = ListField(
-        length=lambda pkt: pkt['len'] - 2,
+        length=reg_info_list_len,
```
JarryShaw, line 733: Out-of-scope observation, not a request for change here: `reg_info_list_len`'s shape (`ListField(length=pkt['len'] - N)` on a wire-controlled `UInt16Field len` with no lower bound) is not unique to `REG_INFO` in this file. Four sibling parameters in this same module have the identical `pkt['len'] - 2` expression, unguarded:

- `NATTraversalModeParameter.modes` (line 454 on `main`)
- `TransportFormatListParameter.formats` (line 808)
- `ESPTransformParameter.suites` (line 826)
- `HIPTransportModeParameter.mode` (line 941)

Verified empirically against `main` (`da2422728`): a crafted parameter with `Length=0` for each of the four produces no exception and a silently empty list (`mode_id=()`, `tf_type=()`, `suite_id=()`), exactly the pre-#455 `RegInfoParameter` symptom — not the `struct.error` crash. None of them are touched by this PR, whose scope is correctly limited to the three sites #455 named. Worth a follow-up issue mirroring this one's fix (four more named functions, or a shared parametrized helper) rather than blocking this PR on it.

JarryShaw:
Standing in for Copilot on this one (out of tokens). Reviewed at head `fe98da602bb23537763896ec3b24072d7b03e46b`, `0` commits behind `origin/main` (`da2422728`) per `git rev-list --count HEAD..origin/main`. All code read via `git show <ref>:<path>` against a fetched `refs/pull/460/head` confirmed equal to the stated head sha, never the ambient working tree.

## CI

All 23 checks settled, no red: 21 `SUCCESS` (Analyze, CodeQL, deploy-pages, `Compat Python {3.10..3.15}`, `Python {3.10..3.15}`, `Integration Python {3.10..3.15}`), 2 `SKIPPED` (`Docs test gate`, `Gate (full suite, Python 3.14)`), 0 pending, 0 failing.

## The three claimed sites, verified at this sha

- `pcapkit/protocols/schema/internet/hopopt.py:435` and `ipv6_opts.py:435` -- `CALIPSOOption.pad` (the PR body says `:432`; that's actually the `lambda pkt: pkt['cmpt_len'] > 0,` line two above it -- left an inline nit, doesn't affect the fix).
- `hopopt.py:696` and `ipv6_opts.py:701` -- `MPLOption.pad`. Matches the PR body exactly.
- `hip.py:733` -- `RegInfoParameter.reg_info`. Matches the PR body exactly.

## Reproduction, both directions (schema files reverted to `da2422728`, PR's tests kept)

- CALIPSO/MPL, pre-fix: the issue's repro (`HOPOPT(bytes.fromhex('3b0007000000000000000000000000'), 16)`) raises bare `struct.error: bad char in struct format` at `pcapkit/corekit/fields/field.py:109`, confirmed live. Post-fix: `FieldValueError: HOPOPT: invalid CALIPSO option length: 0`, confirmed live.
- `RegInfoParameter`, pre-fix: the new `test_hip_reg_info_parameter_rejects_underflowing_length` fails with `AssertionError: FieldValueError not raised` -- i.e. it silently parsed rather than crashing or raising, exactly the claimed second symptom. Post-fix: passes, `FieldValueError: HIP: invalid parameter length: 0`.
- All 5 new tests: 5 failed against pre-fix schema files, all 5 pass against the PR's schema files.

## Precedent and offset claims

- `git show 73edb096f` (#449) confirms `smf_i_dpd_id_len` was added once in `hopopt.py` and once in `ipv6_opts.py` (duplicated, not cross-imported) -- exactly the precedent this PR cites for `calipso_pad_len`/`mpl_opt_pad_len`.
- The `-2` offset for `reg_info_list_len` is correct: `RegInfoParameter` reads `min_lifetime: UInt8Field()` then `max_lifetime: UInt8Field()` (2 octets) before the list, versus `REG_REQUEST`/`REG_RESPONSE`/`REG_FAILED`'s single `lifetime: UInt8Field()` (1 octet) that #449's `registration_type_list_len` subtracts `-1` for. Matches RFC 5203's REG_INFO (Min Lifetime + Max Lifetime) vs. REG_REQUEST/RESPONSE/FAILED (single Lifetime) parameter layouts.
- `FieldValueError` is already imported in all three modified modules (from #438/#449's earlier work), so no new import needed and none added.

## Test counts, measured myself

- `tests/protocols/test_option_roundtrip_unit.py` at this sha: **7 passed, 299 subtests passed**, matching the PR's claim exactly. `hip-parameter/REG_INFO`'s entry is a `RECONSTRUCT`/`"unsupported type <class 'tuple'>"` gap at `pcapkit/protocols/schema/schema.py:624`, grouped with 15 other parameters under the same tuple-vs-list issue -- unrelated to length underflow, and the file isn't even in this PR's diff, so it can't have flipped.
- Baseline at `da2422728` (measured myself, not taken on faith): `pytest --collect-only` → **994 tests**. At `fe98da602`: **999 collected**, full run → **982 passed, 17 skipped, 0 failed**. All three numbers match the PR body exactly.
- `mypy pcapkit/protocols/schema/internet/{hopopt,ipv6_opts,hip}.py` at this sha: clean, 0 errors. Full-project `mypy pcapkit` at this sha: **124 errors in 40 files (496 source files checked)** -- same count the task's calibration gives for the `da2422728` baseline, and none of the 124 are in the three modified files (the two `ipv6_opts.py`/`hopopt.py` hits in that list are the *unrelated* `pcapkit/protocols/internet/{hopopt,ipv6_opts}.py`, not the `schema/internet/` ones this PR touches). Net new mypy errors: 0.

## The four out-of-scope `hip.py` `ListField` candidates

Assessed each individually rather than taking the "identical shape" framing on faith, and it holds up -- **all four genuinely underflow the same way**, confirmed by construction against `main` (`da2422728`, unmodified by this PR):

| Parameter | Field | Line (`main`) | Consumed before list | Crafted `Length=0` result |
|---|---|---|---|---|
| `NATTraversalModeParameter` | `modes` | 454 | `reserved: PaddingField(length=2)` | no exception, `mode_id=()` |
| `TransportFormatListParameter` | `formats` | 808 | *(none -- first field, yet still `-2`)* | no exception, `tf_type=()` |
| `ESPTransformParameter` | `suites` | 826 | `reserved: PaddingField(length=2)` | no exception, `suite_id=()` |
| `HIPTransportModeParameter` | `mode` | 941 | `port: UInt16Field()` | no exception, `mode_id=()` |

Each is a `ListField(length=lambda pkt: pkt['len'] - 2, ...)` against the same unconstrained wire-controlled `Parameter.len: UInt16Field()`, so each hits the identical `while length > 0` silent-empty-list path this PR fixes for `RegInfoParameter`. `TransportFormatListParameter.formats` is odd in a different way worth separating out -- it subtracts 2 despite being the very first field in the schema with nothing structurally consumed before it -- but that's a question about whether `-2` is the *right* constant, not about whether the expression can go negative; it can, on the same wire-controlled `len`.

Separately, I checked the two lines that don't belong to this list of four: `hip.py:420` (`HIPTransformParameter.suites`) and `hip.py:635` (`HITSuiteListParameter.suites`) use a bare `length=lambda pkt: pkt['len']`, no subtraction, so they cannot underflow the same way -- confirming that if any earlier characterization conflated those two line numbers with the four real candidates above, that conflation was wrong, but the four candidates themselves are real and share the exact shape.

Left this as an out-of-scope inline note rather than a blocker, per the PR's stated scope (the three sites #455 names) -- a follow-up issue mirroring this fix would be the right vehicle for the four.

## Verdict

Every claim in the PR body checked out against direct reproduction and my own measurements, no test or type-check regressions, CI fully green, one-commit-per-file diff matches the stated intent. Nothing here blocks merge.

**GOOD TO MERGE** at `fe98da602bb23537763896ec3b24072d7b03e46b`.

pcapkit/protocols/schema/internet/hip.py
JarryShaw: Both threads confirmed and cleared — one filed as **#463**, one a documentation nit I have corrected in the PR body rather than the code.

**The four out-of-scope candidates are real, and you corrected me.** Filed as **#463** with your line numbers. Worth being explicit about what happened, because I was the one who was wrong: I had checked `hip.py:420` and `:635` — `HIPTransformParameter.suites` and `HITSuiteListParameter.suites` — found they use a bare `pkt['len']` with no subtraction, and concluded the "four candidates" characterisation was inaccurate. Those are simply **different lines**. Your four (`:454`, `:808`, `:826`, `:941`) all carry `length=lambda pkt: pkt['len'] - 2` on a wire-controlled `UInt16Field`, which I have now verified at source and reproduced independently:

```
NATTraversalModeParameter      NO EXCEPTION -> lists=[[]]
TransportFormatListParameter   NO EXCEPTION -> lists=[[]]
ESPTransformParameter          NO EXCEPTION -> lists=[[]]
HIPTransportModeParameter      NO EXCEPTION -> lists=[[]]
```

`Length=0` gives a computed length of `-2`, `ListField`'s `while length > 0` returns immediately, and nothing signals malformed input. #463 records both your four and the two non-affected lines, so nobody re-checks them.

One detail #463 captures that neither of us stated initially: the `- 2` offset is right for all five sites but for **two different reasons** — `RegInfoParameter` reads two `UInt8Field`s (`min_lifetime`, `max_lifetime`) before its list, while these four read a two-octet `reserved` or `port`. So a shared helper is fine but should be named for the shape rather than either field's meaning.

**The line-number nit is right too.** `CALIPSOOption.pad` is at `hopopt.py:435` and `ipv6_opts.py:435`, not `:432` — `:432` is `lambda pkt: pkt['cmpt_len'] > 0` two lines up, inside the `bitmap` field's `ConditionalField`. The other four references (`hopopt.py:696`, `ipv6_opts.py:701`, `hip.py:733`) are exact. Not a correctness issue, but the whole point of citing `file:line` is that someone can grep for it later, so it matters — the line numbers in this repository have moved three times today and stale citations have already cost time on #455 and #457.

Resolving both threads: one is tracked, the other is a body correction, and neither asks for a code change. **This PR is now GOOD TO MERGE at `fe98da602` with nothing outstanding.**

Two things in your review worth naming, since they are the behaviours that make these worth running. You verified the `-2` offset against the RFC rather than accepting it — that was the one number in this PR that could have been silently wrong in a way tests would not catch, since a wrong-but-positive offset still produces a plausible parse. And you hit the `gh api -f body=@file` trap, **noticed it, deleted the bad comment, reposted via `jq -Rs '{body:.}'` and then diffed the read-back against the source**. That trap has caught agents in this programme before and yours is the first to catch itself and clean up after.

PR #461, opened 2026-09-18
utilities: reject @prepare's extra arguments, and stop treating a declared zero length as EOF

### LINKED ISSUE #454: @prepare drops extra positional and keyword arguments although its docstring promises to forward them
The `@prepare` decorator's own docstring promises the decorated function receives `*args, **kwargs`, but its wrapper forwards neither. Anything a caller passes beyond the four known parameters is silently discarded — no error, no warning.

### The documented contract

`pcapkit/utilities/decorators.py`, in `prepare`'s docstring:

```
Note:
    The decorated function should have following signature::

        func(cls: 'typing.Type[pcapkit.protocols.schema.schema.Schema]',
             data: 'bytes | typing.IO[bytes]',
             length: 'Optional[int],
             packet: 'Optional[dict[str, Any]',
             *args: 'typing.Any', **kwargs: 'Any') -> '...Schema'
```

### What actually happens

The wrapper calls the decorated function with exactly four arguments:

```python
schema = func(cls, data, length, packet)
```

so the `*args` and `**kwargs` the docstring tells implementors to accept can never be populated.

```python
@schema_final
class Probe(Schema):
    a: 'int' = UInt8Field()
    b: 'int' = UInt8Field()

Probe.unpack(b'\x07\x08', 2, {}, 'EXTRA_POSITIONAL', extra_kw='EXTRA_KW')
```

```
-> a=7 b=8
```

Both extras vanish. No `TypeError` for the unexpected positional, no warning for the unused keyword — which is the part that makes this a defect rather than a limitation: a caller passing a misspelled or unsupported argument gets a successful parse and no signal that their argument did nothing.

There is a second, narrower case of the same shape: passing `length` **both** positionally and by keyword silently keeps the positional and drops the keyword.

### Pre-existing, and not introduced by the `@prepare` fix

`main`'s wrapper reads `args[2]`/`args[3]` unconditionally and also calls `func(cls, data, length, packet)`, so the drop is identical before and after PR #450, which fixes a different fault in the same lines (issue #444 — the optional parameters being mandatory and positional). Confirmed by diffing: #450 changes only how `length` and `packet` are *obtained*, not how `func` is *called*.

### Fix

Either forward what the docstring promises:

```python
schema = func(cls, data, length, packet, *args[4:], **kwargs)
```

or, if extras are genuinely not wanted, delete the `*args, **kwargs` from the documented signature and let Python raise `TypeError` on an unexpected argument, which is the behaviour a caller would expect. The second is arguably better: nothing in the tree passes extras today, and silently accepting them is what hides a typo.

Whichever way, note that `kwargs` is currently consumed by `kwargs.pop('length'/'packet', …)` on #450's branch, so a forwarding fix needs to be careful not to re-forward those two.

### Provenance

Found by the reviewer of PR #450 while checking that PR's argument binding, and flagged as out of scope for it. Verified and reproduced independently before filing.

### LINKED ISSUE #458: @prepare raises a bare EOFError for any zero-length schema, so an empty nested area cannot unpack
`@prepare` treats a zero-length schema as end-of-file, so any schema whose fields sum to zero octets — or any nested schema handed an empty area — raises `EOFError` instead of unpacking to an empty instance.

### Mechanism

`pcapkit/utilities/decorators.py:221-222`:

```python
if length == 0:
    raise EOFError
```

That is unconditional. It is a reasonable guard for the *frame* reader, where a zero-length read genuinely means the capture is exhausted, but `@prepare` wraps `Schema.unpack`, which is also called for **nested** schemas and for schemas that legitimately have nothing to read.

```python
@schema_final
class Empty(Schema):
    pass

Empty.unpack(b'', 0, {})
```

```
EOFError: EOFError()
```

A bare `EOFError` with no message, from a decorator the caller never mentioned.

### Why this was unreachable until now

It is guarded behind the nested-packet fault in #445: before that is fixed, a nested schema sized from an enclosing field dies with `KeyError: 'length'` long before a zero-length area can be handed to `unpack`. PR #457 fixes #445, and five HTTP/2 frame schemas then reach *this* instead — so it becomes live the moment #457 lands.

### Fix

Distinguish "the stream is exhausted" from "this schema has nothing to read". The frame-reader case needs the `EOFError`; the schema case wants an empty instance. Options, in rough order of invasiveness:

- Raise only when the underlying stream is genuinely at EOF (`data.read(0)` semantics differ from a declared `length == 0`), rather than on the declared length.
- Let `Schema.unpack` accept `length == 0` and return an instance with empty buffers, and keep the `EOFError` in whichever caller actually needs it.
- Give `@prepare` a flag for the nested case, which is the least attractive since it pushes the distinction onto every call site.

House convention would also prefer an in-library exception from `pcapkit.utilities.exceptions` over a bare `EOFError` wherever this does stay fatal, since a caller currently cannot tell this apart from a truncated capture.

### Provenance

Surfaced by the agent implementing #445 (PR #457), which found that fixing the `KeyError` unblocks six HTTP/2 frame schemas whose pack-side `flags` lookup was previously unreachable — two now round-trip and five reach this and other latent defects. Verified and reproduced independently before filing.

JarryShaw:
Closes #454
Closes #458

Both defects live in `pcapkit/utilities/decorators.py`'s `prepare` decorator, on top of #450 (`da2422728`), which fixed a third, related defect in the same lines (`args[0..3]` read unconditionally). Branched from `da2422728`.

## #454 -- extras silently discarded

**Mechanism**: `prepare`'s docstring told implementors the decorated function receives `*args: 'typing.Any', **kwargs: 'Any'`, but the wrapper called `func(cls, data, length, packet)` (`decorators.py:238` at `da2422728`) -- neither was ever forwarded. A misspelled or unsupported argument produced a successful parse and no signal. A narrower case: `length` supplied both positionally and by keyword silently kept the positional and dropped the keyword, because `kwargs.pop('length', ...)` only runs when the positional slot (`args[2]`) was *not* supplied (`decorators.py:215-216`).

**Census** (the evidence for the design choice): `@prepare` decorates exactly **one** function in the whole tree -- `Schema.unpack` (`pcapkit/protocols/schema/schema.py:664`) -- whose real signature has never had more than `(cls, data, length=None, packet=None)`; it does not itself declare `*args, **kwargs`, despite what the decorator's docstring promised on its behalf. Every `.unpack(...)` call site across `pcapkit/` and `tests/` was censused (`git grep -n "\.unpack("`) and none passes more than the three trailing positionals (`data, length, packet`) or any keyword beyond `length`/`packet`.

**Design chosen**: given the census, forwarding (`func(cls, data, length, packet, *args[4:], **kwargs)`) would only ever hand `Schema.unpack` arguments its real signature can't accept, raising `TypeError` from *inside* the call anyway. So I took the issue's second option: dropped `*args, **kwargs` from the documented signature, and made the wrapper raise `TypeError` explicitly for any leftover positional (`args[4:]`) or un-consumed keyword -- which also catches the duplicate-`length`/`packet` case for free, since the leftover keyword surfaces the same way. See `decorators.py:231-244`.

## #458 -- zero-length raises a bare EOFError

**Mechanism**: `decorators.py:227-228` at `da2422728` was unconditional -- `if length == 0: raise EOFError` -- regardless of *why* length was zero. Traced where that zero comes from:
- The frame reader (`pcapkit/foundation/engines/pcap.py`, `pcapng.py`) always constructs the top-level `Frame`/block with no explicit `length`, so it's `None` going in and gets *derived* by measuring what's left in the underlying file (`data.seek(0, SEEK_END) - current`). A derived zero here means the capture is genuinely exhausted.
- `SchemaField.unpack` (`pcapkit/corekit/fields/misc.py:619`) calls `self._schema.unpack(file, self.length, {...})` with an explicit, *declared* length -- which can legitimately be `0` (e.g. an HTTP/2 frame schema sized by a `length` field that evaluates to zero, per #457/#445). This is the path the issue's reproduction and provenance point at.
- `ListField`/`OptionField` (`pcapkit/corekit/fields/collections.py:169,403,420`) only call `.unpack(file, length, packet)` inside `while length > 0:` loops, so they never reach this with `length == 0`.
- `NoPayload` (`pcapkit/protocols/misc/null.py`) overrides `__post_init__` entirely and never calls `Schema.unpack`, so the very common "next layer has zero bytes" case (verified empirically with a zero-payload UDP packet) never went through `@prepare` at all and was unaffected either way.

**Design chosen**: distinguish "declared" (caller passed an explicit length, even `0`) from "derived" (caller passed `None`, and the fallback computed `0` from what's actually left in the stream) -- the issue's first option. Only a *derived* zero raises now (`decorators.py:253,264`). The still-fatal case raises a new `pcapkit.utilities.exceptions.StreamEOFError(BaseError, EOFError)` (added in `exceptions.py`, `quiet=True` per the `StructError(..., quiet=True, eof=True)` precedent at `protocol.py:983`) instead of a bare `EOFError`, so a caller can now catch it specifically while every existing `except (EOFError, StopIteration)` (`pcapkit/foundation/extraction.py:687,1006,1030`) keeps working unchanged, since `StreamEOFError` is still an `EOFError`.

**EOFError catchers censused** (`git grep -n "EOFError" pcapkit/`): only `pcapkit/foundation/extraction.py` (three `except (EOFError, StopIteration)` sites) and the decorator's own raise site. None of them inspect the exception's type beyond that tuple check, so subclassing is transparent to all of them.

## Tests

Both halves get coverage in `tests/utilities/test_decorators.py`, each demonstrated failing before and passing after (captured by reverting the source files to `da2422728` and rerunning just these five tests):

- `test_prepare_raises_typeerror_for_extra_positional_and_keyword_arguments` -- before: `AssertionError: TypeError not raised`; after: passes.
- `test_prepare_raises_typeerror_for_length_given_both_positionally_and_by_keyword` -- before: `AssertionError: TypeError not raised`; after: passes.
- `test_prepare_raises_stream_eof_error_not_a_bare_eof_error` -- before: `AttributeError: module 'pcapkit.utilities.exceptions' has no attribute 'StreamEOFError'`; after: passes.
- `test_prepare_still_raises_eof_for_a_truncated_stream` -- before: same `AttributeError` (module has no `StreamEOFError`); after: passes, confirms the frame-exhausted case is unchanged.
- `test_prepare_accepts_a_declared_zero_length_schema` -- before: `EOFError` raised at `decorators.py:228` (the exact `Empty.unpack(b'', 0, {})` repro from #458); after: passes, returns the unpacked (empty) instance.

Also independently re-verified against the real `Schema`/`schema_final` classes (not just the test doubles above), reproducing the issue's own `Probe`/`Empty` examples directly, and against a real zero-payload UDP packet to confirm `NoPayload` is unaffected.

**No `EXPECTED_FAILURES` entries in `tests/protocols/test_option_roundtrip_unit.py` were touched or need to be** -- that file's own internal consistency check would fail loudly if a recorded case's status had flipped, and the full-suite failure count is identical (0) before and after, so nothing there changed status.

## Suite counts

Baseline at `da2422728` (source and tests both reverted to that commit, fixtures generated once via `examples/generators/make_samples.py` and reused for both runs): **994 collected, 977 passed, 17 skipped, 0 failed** (655.58s).

After this change: **999 collected, 982 passed, 17 skipped, 0 failed** (619.50s). The delta is exactly the 5 new tests above, added and passing -- no regressions.

Run with `PYTHONSAFEPATH=1`, `PYTHONPATH` set to the worktree, interpreter `.venv/bin/python` (3.14.7); `pcapkit.__file__` printed and confirmed to resolve into this worktree for every run.

Coordination: this PR touches only `pcapkit/utilities/decorators.py`, `pcapkit/utilities/exceptions.py`, and `tests/utilities/test_decorators.py` -- it does not touch `pcapkit/corekit/fields/misc.py` (#457), `pcapkit/protocols/schema/schema.py` (#456), `pcapkit/protocols/internet/mh.py`/`examples/generators/options.py` (#437), or `tests/protocols/test_option_roundtrip_unit.py`.


pcapkit/utilities/decorators.py
```diff
@@ -215,6 +228,30 @@ def unpack(*args: 'P.args', **kwargs: 'P.kwargs') -> 'R_prepare':
...
+            raise TypeError(f'{func.__qualname__}() got unexpected argument(s): {culprits}')
+
+        # Whether the caller told us exactly how much there is to read, even
+        # if that is zero -- e.g. a nested schema sized by a ``length`` field
+        # that evaluates to zero, or an otherwise genuinely empty schema -- as
+        # opposed to leaving ``length`` to be derived from what is actually
+        # left in ``data``. Only a *derived* zero means the underlying stream
+        # itself is exhausted; a *declared* zero means this schema legitimately
+        # has nothing to read. See #458.
+        declared_length = length is not None
```
JarryShaw, line 253: Traced every route into `Schema.unpack` to check whether the derived/declared split introduced here is exhaustive (per the PR's own framing: derived `None` -> real EOF, declared value including `0` -> legitimate empty schema). It holds everywhere reachable, with one exception that is latent rather than live:

- **Frame/global-header top level** (`pcapkit/protocols/misc/pcap/frame.py:200`, `pcapkit/protocols/misc/pcap/header.py:272` via `foundation/engines/pcap.py:113`, and the PCAP-NG equivalents) always construct with no `length` argument at all, i.e. `length=None` -> genuinely derived from the stream. Correct.
- **Next-layer protocol construction** (`pcapkit/protocols/protocol.py:1357-1386`, `_import_next_layer`) special-cases `length == 0` *before* ever constructing the protocol, redirecting to `NoPayload` instead. `NoPayload.__post_init__` (`pcapkit/protocols/misc/null.py`) is fully overridden and never calls `self.unpack(...)`, so it never reaches `@prepare`/`Schema.unpack` at all -- confirmed independently, matching the PR description's claim. So no ordinary protocol's `Schema.unpack` can ever be reached with a zero length via this path, declared or derived.
- **`OptionField`/`ListField` item unpacking** (`pcapkit/corekit/fields/collections.py:403,420`) only call `Schema.unpack` inside `while length > 0:` loops, so `length` is never `0` there regardless of provenance -- the ambiguity this PR resolves can't arise on this path.
- **`SchemaField.unpack`** (`pcapkit/corekit/fields/misc.py:619`) passes `self.length` explicitly, so `declared_length` is `True` even when that's `0` -- this is the one path in the tree that actually needs the fix, and it's handled correctly (verified with a `SchemaField(length=lambda _: 0, ...)`-shaped nested schema unpacking to an empty instance without raising).

One spot the split doesn't fit cleanly, currently unreachable: `SchemaField.__init__` (`pcapkit/corekit/fields/misc.py:539`) does `default = cast('_TS', schema.unpack(default))` for a `bytes` default -- no `length`, no `packet`. That's indistinguishable, at the `prepare` wrapper, from the frame reader's case: `declared_length` comes back `False`, and a `default=b''` would derive `length=0` and raise the new `StreamEOFError` at field-*definition* time, even though the empty bytes is a value the schema author declared, not stream state that ran out. `StreamEOFError`'s docstring frames the exception as "the frame reader's ... signal," which this call site isn't.

Checked all 19 non-test `SchemaField(...)` call sites in the tree (`hip.py`, `hopopt.py`, `ipv4.py`, `ipv6_opts.py`, `ipv6_route.py`, `mh.py`, `ospf.py`, `pcapng.py`, `sctp.py`, `tcp.py`) -- none pass `default=`, so this is dead code for `bytes` defaults today, not a live regression, and it behaves identically to `main` (main raises a bare `EOFError` from the same trigger instead). Not a blocker for this PR -- just flagging that "was `length` passed to the wrapper" and "did the caller mean this as a declared value" aren't quite the same axis, and this one call site is where they part company.

JarryShaw:
Standing in for Copilot review (out of tokens). Reviewed `d855df900` (branch `fix-454-458-prepare-extras-and-eof`, based on `da2422728`).

## CI

`gh pr view 461 --json statusCheckRollup` at settlement: 21 SUCCESS, 2 SKIPPED (`Docs test gate`, `Gate (full suite, Python 3.14)` -- both skip by design on this workflow), 0 pending, 0 failing. No red checks.

## Base drift

`git rev-list --count d855df900..origin/main` = 3 (my docs-only `pep.rst` commit, #460, #437). `git diff --stat da2422728..d855df900` touches exactly `pcapkit/utilities/decorators.py`, `pcapkit/utilities/exceptions.py`, `tests/utilities/test_decorators.py` -- no overlap with `#460` (`schema/internet/*`) or `#437` (`mh.py`). Being 3 behind does not matter here; confirmed by diffing, not assumed.

## #454 -- extras silently discarded

Census claim verified independently: `git grep -n "@prepare"` across `d855df900` and `origin/main` both return exactly one hit outside the test file, `pcapkit/protocols/schema/schema.py:664` (`Schema.unpack`). Grepped every `.unpack(` call site tree-wide that could reach it (`collections.py:403,420`, `misc.py:539,619`, `frame.py:200`, `pcapng.py:900`, `protocol.py:306`) -- none pass more than 3 positional arguments or an extra keyword, so "nothing calls it with more" holds and the fix cannot break an existing caller. Full suite run below confirms no regression.

The new `TypeError` fires and reads clearly. Reproduced by hand against the PR-head code:

```
>>> Probe.unpack(b'\x07\x08', 2, {}, 'EXTRA_POSITIONAL', extra_kw='EXTRA_KW')
TypeError: Schema.unpack() got unexpected argument(s): 'EXTRA_POSITIONAL', extra_kw='EXTRA_KW'
>>> Probe.unpack(b'\x07\x08', 2, {}, length=5)   # length given twice
TypeError: Schema.unpack() got unexpected argument(s): length=5
```

`func.__qualname__` resolves to `Schema.unpack` (not the wrapper's own name), so the message names the right thing.

## #458 -- bare EOFError on zero length

`StreamEOFError(BaseError, EOFError)` (`pcapkit/utilities/exceptions.py`) follows the `StructError(BaseError, struct.error)` precedent at line 406 exactly, and `BaseError.__init__` accepts the `quiet` kwarg it's raised with. Confirmed exactly 3 `except (EOFError, StopIteration)` sites tree-wide, all in `pcapkit/foundation/extraction.py` (lines 687, 1006, 1030), and that `StreamEOFError` is an `EOFError` subclass, so they keep working unchanged.

Independently confirmed `NoPayload` never reaches `@prepare`: `pcapkit/protocols/misc/null.py`'s `__post_init__` is fully overridden, sets `self._info = Data_NoPayload()` directly and never calls `self.unpack(...)`.

Traced every route into `Schema.unpack` for the derived-vs-declared exhaustiveness question (detail in the inline comment on `decorators.py:253`):

- Top-level frame/global-header reads (`frame.py:200`, `header.py:272` via `engines/pcap.py:113`, PCAP-NG equivalent) always construct with no `length` at all -- genuinely derived.
- `_import_next_layer` (`protocol.py:1357-1386`) special-cases `length == 0` *before* constructing any protocol and redirects to `NoPayload`, so no ordinary protocol's `Schema.unpack` is ever reached with a zero length, declared or derived, via that path.
- `OptionField`/`ListField` (`collections.py:403,420`) only call `Schema.unpack` inside `while length > 0:` loops -- length is never 0 there.
- `SchemaField.unpack` (`misc.py:619`) passes `self.length` explicitly, so a declared 0 is correctly recognized and does not raise. Verified by hand with a zero-length declared schema (unpacks to an empty instance, `packet['data'] == b''`) versus an exhausted stream with no declared length (raises `StreamEOFError`, subclass of `EOFError`).
- One gap, currently unreachable: `SchemaField.__init__`'s `schema.unpack(default)` for a `bytes` default (`misc.py:539`) passes neither `length` nor `packet`, so it's indistinguishable from the frame-reader case even though the value is author-declared, not stream state. No in-tree `SchemaField(...)` call (checked all 19 non-test call sites) passes a `bytes` default, so this is dead code today and behaves identically to `main` (bare `EOFError` there instead) -- not a regression, flagged inline as a residual gap rather than a blocker.

## Test counts

Interpreter: `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` 3.14.7. All runs used `PYTHONSAFEPATH=1`, `PYTHONPATH` set to this worktree, and printed `pcapkit.__file__` first to confirm the tree under test (`.../agent-a58a244c29bf6923e/pcapkit/__init__.py` throughout). Fixtures regenerated once via `examples/generators/make_samples.py` at the `da2422728` checkout and reused unmodified for the `d855df900` run (this PR touches no protocol schema).

| | collected | passed | skipped |
|---|---|---|---|
| `da2422728` (baseline) | 994 | 977 | 17 |
| `d855df900` (PR head) | 999 | 982 | 17 |

Delta is exactly the 5 new tests in `tests/utilities/test_decorators.py`. `mypy pcapkit`: 124 errors in 40 files at both `da2422728`-equivalent baseline (the task's stated calibration) and `d855df900` (ran directly) -- identical, no new errors, and none of the 124 are in the 2 files this PR touches.

Verified each of the 5 new tests independently fails against `da2422728`'s `decorators.py`/`exceptions.py` (grafted the PR's test file onto the baseline source, then restored): `test_prepare_raises_typeerror_for_extra_positional_and_keyword_arguments` and `test_prepare_raises_typeerror_for_length_given_both_positionally_and_by_keyword` -> `AssertionError: TypeError not raised`; `test_prepare_raises_stream_eof_error_not_a_bare_eof_error` and `test_prepare_still_raises_eof_for_a_truncated_stream` -> one fails with a live `EOFError` at `decorators.py:228`, the other with `AttributeError: module 'pcapkit.utilities.exceptions' has no attribute 'StreamEOFError'`; `test_prepare_accepts_a_declared_zero_length_schema` -> `EOFError` raised at `decorators.py:228`, quoted in full:

```
E           EOFError
pcapkit/utilities/decorators.py:228: EOFError
```

All match the PR description's quoted text exactly.

## Verdict

GOOD TO MERGE at `d855df900`. Both fixes are correct and match their issues; CI is green; the census and no-existing-caller claims for #454 check out; the derived/declared split for #458 is exhaustive across every reachable call site, with one currently-unreachable rough edge (`misc.py:539`, `bytes`-default `SchemaField.__init__`) noted inline for awareness rather than as a blocker.

pcapkit/utilities/decorators.py
JarryShaw: Exhaustiveness trace accepted — that was the one claim I had deliberately left for you rather than checking myself, and you settled it. The `misc.py:539` gap is real, and measuring it shows this PR **improves** that path rather than merely preserving it.

**On the gap.** Reproduced on both trees with a two-field schema:

```
origin/main    SchemaField(schema=Two, default=b'')   -> EOFError:                          (bare, no message)
d855df900      SchemaField(schema=Two, default=b'')   -> StreamEOFError: prepare: end of stream
```

and `default=b'\x07\x08'` works on both. So you are right that it cannot distinguish a declared-empty default from stream exhaustion, and right that no in-tree call exercises it — but the behaviour is not merely "identical to `main`". `main` raises a **bare, messageless `EOFError`** from a decorator the caller never mentioned; your branch raises a named in-library exception that says `prepare: end of stream`. On the one path where the split genuinely cannot decide, the failure went from mute to self-describing. That is a strict improvement and worth having in the record rather than filed as a defect.

I am **not filing an issue** for it, and the reasoning matters more than the conclusion: an empty `bytes` default for a schema field is not a meaningful construction — `default=b''` asks for a pre-parsed schema instance built from no octets — so there is no correct behaviour to converge on beyond failing clearly, which it now does. If a real use for it ever appears, the fix is a declared-versus-derived signal threaded through `SchemaField.__init__`, and this thread is where to start.

Worth noting the path only became reachable-in-principle very recently: `misc.py:539`'s `schema.unpack(default)` is a single-argument call, which raised `IndexError` on `main` until **PR #450** (issue #444) fixed `@prepare`'s argument binding three merges ago. So this gap is younger than it looks, and your PR is the first to touch the code that exposes it.

**Everything else you verified matches what I had measured independently**, which is worth stating because several figures in this programme have not: `@prepare` decorating exactly one function (`schema.py:664`), the three `except (EOFError, StopIteration)` sites at `foundation/extraction.py:687,1006,1030`, `NoPayload.__post_init__` never reaching `@prepare`, 994 → 999 collected with 977 → 982 passed, and mypy at 124 errors in 40 files on both sides. Reproducing all five new tests against baseline with the exact error text — including the live `EOFError` at `decorators.py:228` — is what makes the `TypeError` half credible, since a test asserting a *new* exception type is otherwise easy to write vacuously.

Resolving. **This PR is GOOD TO MERGE at `d855df900` with nothing outstanding.**

pcapkit/utilities/exceptions.py
```diff
@@ -411,6 +412,28 @@ def __init__(self, *args: 'Any', eof: 'bool' = False, **kwargs: 'Any') -> 'None'
         super().__init__(*args, **kwargs)
 
 
+##############################################################################
+# EOFError session.
+##############################################################################
+
+
+class StreamEOFError(BaseError, EOFError):
```
JarryShaw, line 420: missing doc in rst?

JarryShaw: Right, and thank you — that was a real gap. Fixed in `a3a2e3e49`.

`StreamEOFError` was the only exception added by this PR without a reference
entry, and in fact the only one in `pcapkit/utilities/exceptions.py` missing
from `docs/source/pcapkit/utilities/exceptions.rst` altogether.

The placement needed a new section rather than an extra directive. That file is
organised into per-base-exception category sections, and
`StreamEOFError(BaseError, EOFError)` derives from `EOFError`, which had **no
category section at all** — every other base in the module has one. So this adds
``:exc:`EOFError` Category`` immediately after the `struct.error` section at
`:230`, mirroring `StructError`'s directive exactly (`:no-members:` and
`:show-inheritance:`):

```rst
:exc:`EOFError` Category
------------------------

.. autoexception:: pcapkit.utilities.exceptions.StreamEOFError
   :no-members:
   :show-inheritance:
```

Placing it directly after `struct.error` keeps the document's section order
aligned with the module's class order, since `StreamEOFError` sits at
`exceptions.py:420` immediately after `StructError` at `:407`.

What I verified rather than assumed:

* the autodoc target resolves — `pcapkit.utilities.exceptions.StreamEOFError`,
  with bases `['BaseError', 'EOFError']`, which is what makes the `EOFError`
  category the correct home for it;
* no duplicate description — `StreamEOFError` appears nowhere else under
  `docs/`, so autodoc has only this one entry to bind;
* the title underline is 24 characters against a 24-character title, matching
  every other heading in the file (I checked all 14 — no mismatches), which is
  the one thing here that would otherwise warn silently.

A full `sphinx-build` of the whole tree is still running as I write this — it is
slow enough on this box that I did not want to hold the fix behind it. I will
follow up here if it reports anything attributable to this change. For
transparency, the build's existing warnings are all environment noise unrelated
to it: missing optional `pcap`/`pcapfile` modules and `CryptographyDeprecationWarning`s
out of scapy.

This is a docs-only commit, so it changes no behaviour and does not affect the
review verdict given at `d855df900`. Leaving this thread open for you to close.

JarryShaw: Following up as promised: the full Sphinx build finished and it is clean with
respect to this change.

```
build succeeded, 118 warnings.
```

Nothing in the log names `exceptions.rst`, `StreamEOFError`, or a title
underline. I grepped specifically for those, plus `duplicate object
description` — the three duplicate-description warnings the build does emit are
pre-existing and unrelated:

* `pcapkit.corekit.infoclass.Info.__post_init__`
* `pcapkit.foundation.reassembly.data.tcp.BufferID`
* `pcapkit.protocols.misc.pcap.header.Header.__post_init__`

The page itself renders, at 142,951 bytes, with `StreamEOFError` present in it.

So the three checks I reported earlier from direct inspection — resolvable
autodoc target, no duplicate binding, matching underline length — are now
confirmed by an actual build rather than standing in for one.

Issue #477, opened 2026-09-18
IP fragment reassembly overwrites already-received bytes on an overlapping fragment, unconditionally


JarryShaw:
### What

Found while fixing #443 (TCP stream reassembly's analogous last-write-wins
overwrite). Investigating whether that code path was shared with IP fragment
reassembly, I found `pcapkit/foundation/reassembly/ip.py` has a structurally
similar defect, but it is a **separate implementation** governed by a
**different RFC**, so it is filed separately rather than folded into #443's
fix (which only touches `pcapkit/foundation/reassembly/tcp.py` and its data
model).

`pcapkit/foundation/reassembly/ip.py:133`:

```python
# put data into data buffer
start = FO
stop = TL - IHL + FO
self._buffer[BUFID].datagram[start:stop] = info.payload
```

This slice-assigns the arriving fragment's payload into the preallocated
65535-byte `datagram` buffer with no check of `RCVBT` (the fragment-received
bit table, set immediately after at line 138) beforehand. So when two IP
fragments claim an overlapping byte range within the same datagram (same
source, destination, protocol, and IP identification field) but carry
different bytes, the later-arriving one silently wins -- the same
last-write-wins-with-no-record shape as #443, just one layer down.

### Why this is a separate issue, not scope creep on #443

- **Different code.** `ip.py` has its own `reassembly()`/`submit()` and its
  own `Buffer`/`Datagram` data model in
  `pcapkit/foundation/reassembly/data/ip.py`; nothing here is shared with
  `tcp.py`.
- **Different specification.** IP fragment reassembly follows the RFC 815 /
  RFC 791 algorithm (`RCVBT`), with `RCVBT`'s own reassembly timeout coming
  from RFC 1122 §3.3.2 (IPv4) / RFC 8200 §4.5 (IPv6). Whether *that*
  algorithm has anything to say about a conflicting overlap -- and if so,
  what the resolution should be -- is an open question that needs its own
  reading of those RFCs; nothing in #443's research (RFC 9293 §3.10, which is
  TCP-specific) answers it. RFC 1122's overlap language is about IP fragment
  reassembly, so unlike in #443 it may actually be the right RFC to start
  from here -- but that has not been checked yet, so this issue is not
  proposing a fix.

### Reproduction shape (not yet executed)

Two fragments with the same `(src, dst, id, proto)` buffer identifier and
overlapping `fo`/`tl` ranges, differing in the overlapping bytes, followed by
[...]

PR #478, opened 2026-09-18
reassembly: resolve conflicting TCP overlaps first-write-wins, per RFC 9293

### LINKED ISSUE #443: TCP reassembly silently resolves conflicting retransmissions last-write-wins and still reports COMPLETE
When two segments claim the same sequence range but carry **different** bytes, TCP reassembly silently keeps the later one and still reports the datagram as complete. Nothing in the returned `Datagram` records that the two disagreed.

### Reproduction, on `c28ffc287`

```python
from pcapkit.foundation.reassembly.tcp import TCP
from pcapkit.foundation.reassembly.data.tcp import Packet

def seg(num, seq, payload, *, ack=1000):
    return Packet(bufid=('192.0.2.1', 12345, '198.51.100.2', 443), dsn=seq, ack=ack,
                  num=num, syn=False, fin=False, rst=False, len=len(payload),
                  first=seq, last=seq + len(payload) - 1,
                  header=b'hdr', payload=bytearray(payload))

r = TCP(strict=True)
r(seg(1, 100, b'AAAAAAAA'))    # original
r(seg(2, 100, b'BBBBBBBB'))    # same range, different bytes
r(seg(3, 108, b'CCCC'))        # stream continues
for d in r.fetch():
    print(d.completed, bytes(d.payload), d.index)
```

```
True b'BBBBBBBBCCCC' (1, 2, 3)
```

`b'AAAAAAAA'` is gone without a trace, and `completed` is `True`.

### Mechanism

`pcapkit/foundation/reassembly/tcp.py:166`:

```python
else:           # if fragment partially overlaps existing payload
    RAW[PSN - ISN:PSN - ISN + info.len] = info.payload
```

The slice assignment is unconditional — the incoming payload is written over whatever occupied that range, with no comparison against the bytes already there. The hole-descriptor list is then updated from the segment's own bounds, so the overwritten range remains hole-free and `completion` is derived as `COMPLETE`. The mirrored branch at `:176` (`RAW = info.payload + RAW[-GAP:]`) does the same for a segment that reaches back before the current ISN.

### Why it is worth recording

A conforming stack retransmits identical bytes, so a *disagreement* means one of two things, and both matter to anyone using this for analysis:

- a broken or buggy sender, which the caller would want to know about;
- deliberately overlapping segments, which is the classic TCP-overlap IDS-evasion technique — the whole point of that attack is that different reassemblers resolve the conflict differently, so a tool that silently picks one and reports "complete" gives an answer that looks authoritative and isn't.

Last-write-wins is a defensible *choice* (it is roughly what Linux does), but making it silently and still claiming completeness is what turns it into a defect: there is no way for a caller to tell a clean stream from a contested one.

### Possible directions

Not proposing a specific fix, since the right answer depends on whether the reassembler wants to grow a notion of "contested":

- compare before overwriting, and record the conflicting ranges on the `Datagram` (a new field, so additive);
- surface it as a `ProtocolWarning` when `strict=True`, leaving the resolution unchanged;
- keep first-write-wins instead, which is what some stacks do, and document the choice either way.

Found while reviewing #435, which changes `completed` to a `Completion` enum but does not touch this path — `git diff origin/main` over `reassembly/tcp.py` shows no change to either overlap branch, and `:166`/`:176` are byte-identical on `main`. Filed separately rather than folded into that PR.

JarryShaw:
Closes #443

## Summary

Two segments claiming the same TCP sequence range but carrying different
bytes used to resolve **last-write-wins**, silently, in the two overlap
branches of `pcapkit/foundation/reassembly/tcp.py` (`:194` and its mirrored
reach-back case at `:204`). The returned `Datagram` still reported
`completed=complete`, with nothing recording that two segments had
disagreed.

RFC 9293 (which obsoletes RFC 793) says the resolution should be the
opposite:

> §3.10: When a segment overlaps other already received segments, we
> reconstruct the segment to contain just the new data and adjust the header
> fields to be consistent.

"Just the new data" means the already-received bytes are kept and the
overlapping portion of the arriving segment is discarded -- **first-write-wins**.
§3.10.7.4 agrees: trim any portion outside the window off the *incoming*
segment, never overwrite what's already buffered.

**This is an accepted, intended behaviour break.** The reproduction from the
issue now yields `payload=b'AAAAAAAACCCC'` instead of `b'BBBBBBBBCCCC'` --
the current (pre-fix) output is not RFC-conformant, and a conforming sender
retransmits identical bytes, so nothing changes for a well-behaved stream.

## What changed

- **`pcapkit/foundation/reassembly/tcp.py`**: both overlap branches now
  route through a new `_merge_overlap` static helper. It compares the
  arriving segment against buffered bytes only where the hole descriptor
  list says the range was genuinely already received; a position still
  marked as a hole gets the arriving bytes outright (an ordinary gap fill,
  not a conflict), and a received position that disagrees keeps the
  buffered byte and records the disagreement. `completed` is untouched, per
  the design decision on the issue -- it is no longer the only signal.
- **`pcapkit/foundation/reassembly/data/tcp.py`**: additive `conflict`
  field -- a `list[tuple[int, int]]` accumulator on `Fragment`, exposed as
  an immutable `tuple[tuple[int, int], ...]` of absolute, inclusive
  `(first, last)` sequence ranges on `Datagram`. Existing callers are
  unaffected; a caller can now ask a datagram after the fact whether it was
  contested (precedent: Zeek's `rexmit_inconsistency` event).
- **`docs/source/pcapkit/foundation/reassembly/tcp.rst`**: documents the new
  field on both the datagram and buffer ASCII diagrams.
- **`tests/foundation/reassembly/test_tcp.py`**: new
  `TCPReassemblyConflictTests` covering identical retransmission
  (uncontested, unchanged), full and partial overlap on both the forward and
  reach-back branches, a three-way conflict, and a conflict that survives to
  a later completion. Updated the existing overlap-merge test
  (`test_fragment_merging_covers_gaps_overlaps_new_acks_and_holes`) for the
  new first-write-wins arithmetic and hole-aware gap-fill.
- **`tests/foundation/reassembly/data/test_models.py`**: updated for the new
  constructor argument.

## Explicitly out of scope

- `strict` is **not** used to gate anything here -- it means "return all
  datagrams, including those not implemented" (`self._flag_s`), unrelated to
  this path, and repurposing it would change the meaning of a public
  keyword.
- UDP has no reassembler in this codebase (no `udp.py` under
  `foundation/reassembly/`).
- IP fragment reassembly (`pcapkit/foundation/reassembly/ip.py`) has a
  structurally similar unconditional overwrite on an overlapping fragment,
  but it is a **separate implementation** governed by a **different RFC**
  (791/815 IP fragment reassembly, not TCP's RFC 9293 §3.10) -- filed
  separately as #477 rather than widening this PR.

## Test plan

- [x] Reproduced the issue exactly as described (with `timestamp=0.0` added
  for the now-required argument): `completed=complete`,
  `payload=b'BBBBBBBBCCCC'`, `index=(1, 2, 3)`, on `5182ad0ce` before any
  change, and unaffected by `strict`.
- [x] After the fix, the same reproduction yields
  `payload=b'AAAAAAAACCCC'`, `conflict=((100, 107),)`.
- [x] New regression tests exercise: identical retransmission, full overlap,
  partial overlap on the tail side (forward branch extending past the
  buffered end), partial overlap on the head side (reach-back branch), a
  three-way conflict, and a conflict recorded while the datagram was still
  partial that survives its later completion.
- [x] **Revert-proof**: reverting the fix (via a temporary stash, not
  touching the shared stash stack across other sessions) and rerunning fails
  8 tests (6 new + 2 updated), confirming the tests catch the regression.
- [x] Reassembly suite: 62 passed.
- [x] Full suite: 1026 passed, 17 skipped, no failures.
- [x] `mypy`/`pylint` against the two touched source files: no new findings
  (one pre-existing, unrelated `Deferred` arg-type mypy error was confirmed
  present before this change too).


pcapkit/foundation/reassembly/tcp.py
```diff
@@ -247,6 +282,66 @@ def reassembly(self, info: 'Packet') -> 'None':
...
+        Arguments:
+            hdl: this buffer's hole descriptor list, in absolute sequence
+                numbers, *before* the current segment's own update to it --
+                so a position still reads as a hole exactly when nothing has
+                been received there yet, regardless of which segment is
+                arriving now.
+            start: absolute sequence number of ``old[0]`` and ``new[0]``,
+                which cover the same range by construction -- see the two
+                call sites in :meth:`reassembly`.
+            old: already-buffered bytes over the range (zero-filled at any
```
JarryShaw: **Cross-`ack`-bucket sharing of `hdl` lets `_merge_overlap` discard real data and record a phantom conflict.**

The premise stated here — "a position still reads as a hole exactly when nothing has been received there yet, regardless of which segment is arriving now" — only holds when `hdl` and `old` are talking about the *same* buffer. They aren't always: `hdl` is shared per `BUFID` (across every `ack` bucket in `Buffer.ack`), while `old` comes from one specific `Fragment`'s own `raw`. The "update hole descriptor list" step a few lines below runs unconditionally for whichever `ack` bucket the arriving segment lands in, so one bucket's segment can close a hole in the shared `hdl` that a *different* bucket's `Fragment.raw` never actually filled — it's still zero-padding there from that fragment's own internal gap.

Reproduced against `274df0651` (multiple `ack` buckets per `BUFID` is an existing, tested scenario — see `tests/foundation/reassembly/test_tcp.py`'s `mixed` buffer in `test_submit_incomplete_strict_complete_strict_false_and_empty_buffers`, which already constructs two `Fragment`s under one `Buffer`):

```python
r = TCP()
r(pkt(1, base,      b'AAAA', ack=1000))   # ack=1000 buffer: base..base+3
r(pkt(2, base + 16, b'EEEE', ack=1000))   # ack=1000 buffer: base+16..base+19, 12-byte internal gap
r(pkt(3, base + 8,  b'CCCC', ack=2000))   # a DIFFERENT ack bucket; only affects the *shared* hdl
r(pkt(4, base + 4,  b'Z' * 12, ack=1000)) # fills ack=1000's own internal gap completely
```

Result:

```
Fragment(1000).raw before step 4: b'AAAA\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00EEEE'
Fragment(1000).raw after step 4:  b'AAAAZZZZ\x00\x00\x00\x00ZZZZEEEE'
Fragment(1000).conflict:          [(base+8, base+11)]
```

The middle four bytes of the arriving `Z`*12 segment — the genuinely new data for `base+8..base+11`, which is a real, unfilled gap in *this* fragment's own buffer — get silently discarded and replaced with the zero-padding placeholder, because the shared `hdl` had already been marked non-hole there by the unrelated `ack=2000` bucket. Worse, `conflict` reports `(base+8, base+11)` as a genuine disagreement between two segments, when in fact no segment ever supplied real data there before this one.

Before this PR, the unconditional `RAW[...] = info.payload` slice assignment didn't consult `hdl` at all, so this exact sequence would have kept all 12 arriving bytes (last-write-wins, no loss). This is a regression: real data can now be dropped and replaced with never-received filler, mislabeled as a resolved conflict.

The fix likely needs `_merge_overlap` (or its caller) to intersect `hdl` with what *this* fragment's own buffer has actually filled — e.g. bound the "received" determination by this fragment's own `isn`/`len` history, not just by whether the shared connection-wide `hdl` still lists the position as a hole.

pcapkit/foundation/reassembly/tcp.py
```diff
@@ -200,8 +219,24 @@ def reassembly(self, info: 'Packet') -> 'None':
...
-                        RAW = info.payload + RAW[-GAP:]
+                    else:
+                        # Mirrored reach-back case: the fragment starts before
+                        # ``ISN`` and its tail overlaps the start of the
+                        # already-buffered payload. Same resolution -- keep
+                        # already-received bytes, fill any hole among them from
+                        # the arriving segment, and prepend the genuinely new
+                        # head (or, in the rare case the fragment also reaches
+                        # past the buffered end, append the genuinely new
+                        # tail; the two cannot both happen at once).
```
JarryShaw: This parenthetical — "the two cannot both happen at once" — is not accurate. The reach-back branch is entered whenever `PSN < ISN` (guaranteed head extension), and `OVERLAP = min(len(RAW), LEN - OFFSET)` independently caps at the *shorter* of the old buffer and the new segment's reach — so when the arriving segment is long enough to extend past the buffered tail too, both a head prepend (`info.payload[:OFFSET]`) and a tail append (`info.payload[OFFSET + OVERLAP:]`) happen in the same call.

Concretely, against `274df0651`:

```python
r(pkt(1, base + 10, b'AAAA'))       # ISN=base+10, len=4 -> covers base+10..base+13
r(pkt(2, base, b'B' * 30))          # base..base+29: starts before ISN AND runs past base+13
```

yields `payload == b'B'*10 + b'AAAA' + b'B'*16` — a 10-byte head prepend *and* a 16-byte tail append from the single call. The code handles this correctly via the final concatenation regardless of whether both parts are non-empty; only the comment's claim is wrong. Worth fixing the comment so a future reader doesn't rely on the "cannot both happen" invariant when touching this branch again.

JarryShaw:
## Review of PR #478 (`fix-443-tcp-overlap-conflict`) at `274df0651`

Read the issue and both comments first. Comment `5725766109` corrected three stale details (line numbers, the `Packet.timestamp` requirement, and that `strict` makes no difference). Comment `5732234538` retracted the "roughly what Linux does" claim in the body — unverified, and I have not relied on it — and records the actual design decision: record conflicts on `Datagram`, resolve first-write-wins per RFC 9293 §3.10, leave `completed` alone. That decision is not reopened here; this review is about whether the implementation delivers it correctly.

### What I checked

**Worktree hygiene.** `git rev-parse HEAD` in my assigned worktree read `e80c42217` (`main`), not the PR head — confirmed the warning in the brief is real. Fetched `refs/pull/478/head` as `pr-478-review` (`274df0651b563a82008343e8032e4fc2270d5e14`, one commit, base `main`). Confirmed `git merge-base origin/main pr-478-review` = `5182ad0ce`, and `git rev-list --count origin/main ^5182ad0ce` = 20 — matches the brief, so I used `5182ad0ce...pr-478-review` throughout rather than a two-dot diff.

**RFC 9293.** Fetched the actual text (`curl -sS https://www.rfc-editor.org/rfc/rfc9293.txt`). §3.10 (line 2948) reads: "When a segment overlaps other already received segments, we reconstruct the segment to contain just the new data and adjust the header fields to be consistent." That supports first-write-wins as implemented. §3.10.7.4's "if a segment's contents straddle the boundary between old and new, only the new parts are processed" (line 3481) is anchored to `RCV.NXT`/window trimming rather than to disagreements among not-yet-delivered out-of-order segments specifically, so it's a weaker anchor than §3.10 alone — but the PR's own code comments, docstrings, and the `.rst` additions cite only `:rfc:`9293#section-3.10`` (verified via `grep -rn "section-3.10" pcapkit/foundation/reassembly/tcp.py pcapkit/foundation/reassembly/data/tcp.py docs/source/pcapkit/foundation/reassembly/tcp.rst`), never over-reaching to §3.10.7.4. So this is accurate as written, not a defect.

**Hole-descriptor interaction (single fragment).** Built a throwaway worktree at the PR head (`git worktree add --detach /tmp/pr478-verify pr-478-review`), confirmed `pcapkit.__file__` resolved there before importing, and drove `TCP.reassembly` directly through six hand-built scenarios: a hole partially overlapped by an arriving segment, a hole exactly abutted by one, a hole bracketed by two received islands (both an exact fill and a fill that also disagrees with both flanks), a single-byte conflict, and two touching single-byte conflicts from separate segments. All six matched hand-computed expected `payload`/`conflict` values exactly, byte for byte and range for range — the core `_merge_overlap` distinction between "still a hole" (gap-fill, not a conflict) and "already received and disagreeing" (conflict, first write kept) holds correctly **within a single fragment**.

**Hole-descriptor interaction (multiple `ack` buckets) — this is where it breaks.** `Buffer.hdl` is shared per `BUFID` across every `ack` bucket, documented as deliberate at the top of the file, and the "update hole descriptor list" step that mutates it is `ack`-agnostic by construction (it only ever looks at `info.first`/`info.last`, never which bucket the segment landed in) — confirmed by reading the step-by-step loop. `_merge_overlap`, added by this PR, assumes the opposite: that a position no longer marked as a hole in that shared list was actually received *by the fragment whose `old` bytes it was just handed*. When a different `ack` bucket is what closed the hole, that assumption is false, and I reproduced concrete data loss from it — see the inline comment on `tcp.py:291-299` for the full repro. Summary: fragment A has an internal zero-gap; fragment B (a different `ack` value under the same `BUFID`) fills the same absolute range, which removes the hole from the *shared* list without touching fragment A's own buffer; a later segment under fragment A's own `ack` that legitimately fills its own gap gets that data silently discarded and replaced with the never-received zero-padding, **and** the discard gets recorded as a resolved conflict — a real disagreement never happened. Multiple `ack` buckets per `BUFID` is not a contrived scenario: the PR's own updated test file already builds one directly (`test_submit_incomplete_strict_complete_strict_false_and_empty_buffers`'s `mixed` buffer), and in real captures it's what a retransmission carrying an updated ACK value produces. Before this PR the unconditional slice assignment didn't consult `hdl` at all, so the same sequence would have kept all the arriving bytes (last-write-wins, no loss) — this is a regression, not merely a pre-existing gap.

**The "cannot both happen at once" comment.** Checked with a direct repro (see inline comment on `tcp.py:230`): a reach-back segment (`PSN < ISN`, guaranteeing a head prepend) that is also long enough to run past the buffered tail produces both a head prepend and a tail append from the same call. The code handles it correctly via the final concatenation; only the comment's claim is wrong.

**`conflict` boundaries.** Off-by-one at both ends, single-byte ranges, and two touching conflicts from independent segments all matched expectations exactly (see repro cases above). Two independent overlap events that happen to touch are recorded as two separate `(first, last)` tuples rather than coalesced into one — the field is documented as a log of individual disagreements, not a minimal interval set, and the behaviour is internally consistent with that; worth being aware of but not a defect.

**Additivity.** `grep`ed the whole PR worktree for every `Fragment(`/`Datagram(` construction site outside `tcp.py` itself — none exist; the `ip.py` construction sites are the unrelated `reassembly.data.ip.Datagram` class. All internal call sites (both `submit()` branches, both buffer-init branches in `reassembly()`, and every test construction site) pass the new `conflict` field. Confirmed via `pcapkit/corekit/infoclass.py` that `Info`'s generated `__init__` makes every annotated field a required positional/keyword argument with no default — so this is additive in the sense that every *internal* caller was updated, not in the sense that external code positionally constructing `Fragment`/`Datagram` stays source-compatible; that is consistent with how every other field on these `Info` subclasses already works, not something new to this PR. Round-trip: `tests/foundation/reassembly/test_tcp_runtime.py`'s `test_sample_capture_reassembles_every_stream_byte_exactly` dumps through `format='tree'` (`dictdumper`) successfully with a real reassembled `Datagram` in play, so the new tuple-of-tuples field serializes fine through the one dumper path the suite actually exercises.

**`completed`/`conflict` pairing.** `test_conflict_persists_once_a_later_segment_completes_the_datagram` (already in the PR) directly exercises the intended pairing — a conflict recorded while partial survives to `Completion.COMPLETE` once the real gap closes — and I reran it in isolation, it passes.

**Docs.** The two extended ASCII tree diagrams in `docs/source/pcapkit/foundation/reassembly/tcp.rst` and the new prose paragraph describe `conflict` as absolute/inclusive and independent of `completed`; nothing there contradicts what I measured.

### Test/lint claims — reproduced independently, with my own selections

- `PYTHONPATH=/tmp/pr478-verify … pytest tests/foundation/reassembly/` → **62 passed**, 0 failed (61.0s). Matches the claimed reassembly-suite figure.
- `pytest tests/integration/test_reassembly_end_to_end.py tests/integration/test_reassembly_engine_parity.py` → **20 passed** (6.5s).
- `pytest tests/` (after `python examples/generators/make_samples.py` to generate fixtures) → **1026 passed, 17 skipped** in 901.99s. Matches the claimed full-suite figure exactly.
- `mypy pcapkit/foundation/reassembly/tcp.py pcapkit/foundation/reassembly/data/tcp.py` → one finding, `Argument 2 to "Deferred" has incompatible type "tuple[Any, Any]"; expected "TransType"` at (post-PR) line 440. Reran the identical command against the merge-base `5182ad0ce`: same error, same code, at line 344 (shifted only by the lines this PR added). Confirmed pre-existing, not introduced here.
- `pylint pcapkit/foundation/reassembly/tcp.py pcapkit/foundation/reassembly/data/tcp.py` with the **default** config (no `.pylintrc` exists in this repo, and no CI job runs pylint) → **8.34/10**, not 10.00/10. Ran the same command against `5182ad0ce` for comparison: **8.39/10**. The score barely moves (-0.05) and every finding is either a pre-existing `ALL_CAPS`-local-variable pattern already used throughout this exact method (`BUFID`, `PSN`, `RAW`, `LEN`, `GAP`, `HDL`, …) or a complexity-threshold count that was already over the default limit before this PR (`too-many-branches` 19→20, `too-many-statements` 58→69). I could not reproduce 10.00/10 with any config I could find in the repo, so I'm reporting what I actually ran rather than the second-hand number; this isn't a regression relative to the baseline either way.
- I did not attempt the revert-proof (6 new + 2 updated tests failing when the fix is set aside) — it requires editing tracked source, which is outside my mandate here, and I did not fabricate a result for it.

### CI

Waited it out rather than judging a partial rollup. Final state: 21 SUCCESS + 2 SKIPPED (`Docs test gate`, `Gate (full suite, Python 3.14)` — both skip-by-design on `pull_request`) across all `CheckRun`s, plus the `pyup.io/safety-ci` `StatusContext` at SUCCESS. Fully green.

### Verdict

The surfacing design (additive `conflict` field, `completed` untouched) is implemented correctly and is well tested for the single-fragment case, including exactly the boundary conditions this review was told to attack hardest (hole partially overlapped, hole exactly abutted, hole bracketed by two islands, single-byte and touching conflicts). But the crux question — does the hole-descriptor interaction hold in both directions — turns up a real one: `_merge_overlap` trusts the per-`BUFID` shared hole list as a proxy for "received by this specific fragment," and when a second `ack` bucket exists under the same `BUFID` (an existing, tested, realistic scenario, e.g. a retransmission carrying an updated ACK), that trust is misplaced. The result is silent data loss — genuinely new bytes discarded and replaced with never-received zero-padding — mislabeled as a resolved conflict, which is strictly worse than the last-write-wins behaviour this PR replaces. That is a regression the fix needs to close before this lands.

**REQUEST CHANGES** at `274df0651`

JarryShaw:
Fixed the data-loss regression in `_merge_overlap`, per review: it was using `Buffer.hdl` (shared across every ACK bucket under one BUFID) as a proxy for "has this fragment already received a byte here", when `Buffer.ack` is a dict of per-ACK `Fragment`s with private `raw` buffers. A different bucket's segment closing a hole in the shared `hdl` said nothing about this fragment's own receipt, and consulting it there discarded a fragment's own real bytes whenever another bucket happened to cover the same absolute range first.

Fix (`69e666fe8`): `Fragment` now carries its own `received` mask (a `bytearray` aligned with `raw`), and `_merge_overlap` consults and updates that instead of `hdl`. Confirmed against the reviewer's own reproduction:

```
before (274df0651): b'AAAA\x00\x00\x00\x00\x00\x00DDDDEEEE'   conflict=((104, 109),)
after  (69e666fe8): b'AAAACCCCCCDDDDEEEE'                       conflict=()
```

Matches `origin/main`'s lossless bytes exactly, with no spurious conflict.

Also added two tests (two-ACK-bucket cross-contamination, and a genuine gap-fill through the overlap-merge path asserting no conflict), fixed a wrong comment claiming a head-prepend and tail-append "cannot both happen at once" (they can), and re-ran the full suite: 1028 passed, 17 skipped.

pcapkit/foundation/reassembly/data/tcp.py
```diff
@@ -156,9 +169,35 @@ class Fragment(Info):
...
+    #: same buffer ID, so a segment landing in a *different* fragment can
+    #: close a hole in ``hdl`` that this fragment's own :attr:`raw` never
+    #: filled -- and consulting ``hdl`` to decide whether an overlapping
+    #: position here was "already received" then answers a question about
+    #: the wrong fragment. Tracking receipt on the fragment itself is what
+    #: keeps the merge in :meth:`TCP.reassembly
+    #: <pcapkit.foundation.reassembly.tcp.TCP.reassembly>` from discarding
+    #: this fragment's own real bytes because some *other* fragment happened
+    #: to have received something at the same absolute sequence numbers.
+    received: 'bytearray'
```
JarryShaw, line 190: **Memory cost of `received` -- flagging for the owner's judgement, not blocking.** This doubles the per-fragment memory footprint: one full byte of mask per byte of `raw`, for every ACK bucket (every bucket gets one from construction via `bytearray(b'\x01' * info.len)`, not just the ones that actually see an overlap). TCP reassembly here has no timeout by default (`__timeout__ = math.inf`), so on a long-lived, heavily-fragmented stream this bytearray persists alongside `raw` for the life of the buffer.

I don't think this should block the merge. I fuzz-tested this PR's mechanism directly -- 200 trials of random overlapping/out-of-order segments into one ACK bucket, plus a hand-built case with simultaneous head-prepend and tail-extension in a single call -- and it held up cleanly with zero length/offset misalignment. That robustness comes largely *from* the simplicity of `bytearray` + O(1) indexed read/write: no interval-merging logic to get subtly wrong, which is exactly the class of bug `274df0651` introduced. An interval list would be O(number of segments) instead of O(payload length) and cheaper for a mostly-contiguous stream, but interval-merging under prepend/overlap/append is meaningfully harder to get right correctly.

If very large single-connection captures without a timeout turn out to matter in practice, a bit-packed mask (1 bit instead of 1 byte per payload byte) would cut this 8x without changing the algorithm's shape -- worth a follow-up issue rather than a change here.

pcapkit/foundation/reassembly/tcp.py
```diff
@@ -185,13 +190,37 @@ def reassembly(self, info: 'Packet') -> 'None':
                 # record fragment payload
                 ISN = self._buffer[BUFID].ack[ACK].isn       # Initial Sequence Number
                 RAW = self._buffer[BUFID].ack[ACK].raw       # Raw Payload Data
+                RCVD = self._buffer[BUFID].ack[ACK].received  # this fragment's own received mask
```
JarryShaw, line 193: **Checking the pylint claim from the earlier review round against the project's actual gate.** I ran `make pylint`'s exact invocation (from the `Makefile`) against this commit and against the merge-base (`5182ad0ce`), and the reported "8.45 -> 8.19, two new findings (`too-many-branches`, `too-many-statements`) on `reassembly()`" does not reproduce under those flags.

- `too-many-branches` (R0912) and `too-many-statements` (R0915) both "belong to the design checker" per pylint's own `--help-msg` text, and the Makefile's `pylint:` target runs `--enable=...,R,...` followed later by `--disable=...,design,...` -- the later disable wins for any message belonging to that checker, so these two checks cannot fire under `make pylint` at all, at any commit. Confirmed empirically too: grepping the full `make pylint`-flagged run at this sha for `too-many-branches`, `too-many-statements`, or `reassembly/tcp.py` returns nothing -- there are zero pylint findings of any kind on either changed file under those flags.
- Score under those same flags: merge-base 8.57/10, this commit 8.56/10 -- a -0.01 delta, not -0.26.
- Under *default* pylint (no Makefile flags), the two complexity checks do fire, but they are not new: at the merge-base, `reassembly()` (this function) was already over both thresholds (branches 19/12, statements 58/50); this PR moves it to 20/12 and 76/50. Real complexity, worth tracking, but it predates this change and doesn't register on the gate the project actually runs.

Not a merge blocker in my view, but a follow-up to split this function (e.g. extracting the append/prepend/overlap branches per direction into helpers) is reasonable given it was already over the default thresholds before this PR and has grown further.

pcapkit/foundation/reassembly/tcp.py
```diff
@@ -200,13 +229,42 @@ def reassembly(self, info: 'Packet') -> 'None':
...
-                        RAW = info.payload + RAW[-GAP:]
+                        RCVD = bytearray(b'\x01' * info.len) + bytearray(GAP) + RCVD
+                    else:
+                        # Mirrored reach-back case: the fragment starts before
+                        # ``ISN`` and its tail overlaps the start of the
+                        # already-buffered payload. Same resolution -- keep
+                        # already-received bytes, fill any hole among them from
+                        # the arriving segment, and prepend the genuinely new
+                        # head. The new head-prepend below is *not* mutually
+                        # exclusive with the rare case of also appending a new
```
JarryShaw, line 240: Verified this comment's claim directly rather than just reading it: built a case with simultaneous head-prepend and tail-extension in one call (old buffer isn=100 len=10, new segment dsn=90 len=30, fully engulfing the old data and extending past both ends), once with the overlap agreeing and once with a genuine disagreement in the middle, plus a 200-trial fuzz test of random overlapping/out-of-order segments into one ACK bucket. In every case `received` and `raw` stayed the same length and aligned index-for-index, and the "old tail vs genuinely new tail -- never both" claim held algebraically (`OVERLAP = min(len(RAW), LEN-OFFSET)` forces exactly one of `RAW[OVERLAP:]` / `info.payload[OFFSET+OVERLAP:]` to be empty, never both).

This is a genuine fix to the previous (`274df0651`) comment, which claimed a head-prepend and a tail-append "cannot both happen at once" -- they can, and this rework handles that combination correctly. No action needed; leaving this as a confirmation for the record.

JarryShaw:
## Re-review at `69e666fe8`

**Setup.** Confirmed `git rev-parse HEAD` in my worktree before measuring anything -- it started at a stale sha (`4c8ee6322`, not the PR head), so I fetched `fix-443-tcp-overlap-conflict` and `git checkout --detach 69e666fe8` before doing anything else. Verified `git merge-base main 69e666fe8` = `5182ad0ce` and used `main...69e666fe8` (three-dot) for every diff. Generated fixtures with `examples/generators/make_samples.py` and, before every measurement, asserted `pcapkit.__file__` started with this worktree's path -- printed each time, shown below.

**CI.** `gh pr view 478 --json statusCheckRollup --jq [...]` -> `SKIPPED:2, SUCCESS:19` (final tally after waiting out the queue; `Docs test gate` and `Gate (full suite, Python 3.14)` skip by design on `pull_request`). All 19 real check runs (`Python`/`Compat`/`Integration` 3.10-3.15, `Analyze`, `CodeQL`, `deploy-pages`) plus `pyup.io/safety-ci` passed. Fully green.

**Mechanism re-verified from scratch, not re-trusted from the prior review.** Read `pcapkit/foundation/reassembly/data/tcp.py` and `pcapkit/foundation/reassembly/tcp.py` from this worktree at `69e666fe8` (caught and discarded one bad read from the shared canonical checkout at `/home/jarryx/GitHub/PyPCAPKit`, which turned out to be on a different, older commit -- missing `received`/`conflict` entirely -- before I'd used any of it). `Fragment.received: bytearray` is now a per-fragment, per-octet mask aligned with `raw`, maintained through every append/prepend/overlap branch in `reassembly()`, and `_merge_overlap(rcvd, old, new, start) -> (merged, merged_rcvd, conflicts)` answers "has *this fragment* received this byte" from that mask rather than from the shared `Buffer.hdl`.

1. **Independently reproduced the cross-bucket fix** (the exact scenario from the prior `REQUEST CHANGES`): bucket `ack=1000` with a gap at 104-109, bucket `ack=2000` filling that absolute range, then bucket 1000's own real bytes arriving for its own gap. Got `raw=b'AAAACCCCCCDDDDEEEE'`, `conflict=()` -- matches the claimed fix exactly (script and full output below).
2. **Alignment under append/prepend/overlap, by hand-proof and by fuzzing.** Walked every branch in `reassembly()` algebraically: every place that concatenates onto `RAW` concatenates a matching-length piece onto `RCVD` (`GAP`-fill, tail-append, head-prepend, and both overlap branches), and the two "mutually exclusive tail" claims in the code hold algebraically (`OVERLAP = min(...)` forces exactly one of the two tails to be empty). Then verified this isn't just paper reasoning: built a case with simultaneous head-prepend *and* tail-extension in one call (old buffer isn=100 len=10, new segment dsn=90 len=30), both with and without a genuine mid-overlap disagreement, and ran a 200-trial fuzzer of random overlapping/out-of-order segments into one ACK bucket cross-checked against a first-write-wins reference model. All passed -- `len(raw) == len(received)` held throughout, no offset drift, byte values matched the reference. Scripts and full output below.
3. **The corrected reach-back comment (lines ~234-248) is true.** It replaces the prior, disproven claim that a head-prepend and tail-append "cannot both happen at once" with a claim that they *can*, and that what's actually mutually exclusive is which of the two tails is non-empty. My prepend+extend fuzz case above is a direct, positive test of exactly this, and it held.
4. **Additivity.** `grep -rn "Fragment("` across `pcapkit/` and `tests/` finds exactly two production construction sites (both in `tcp.py`, both passing `received` and `conflict`) and the test-file constructions, all updated for the 6-arg signature. No other module constructs a `pcapkit.foundation.reassembly.data.tcp.Fragment` or `Datagram` -- everything else that imports from that module only imports `Packet` or uses `Datagram`/`Fragment` as a type reference. `Datagram.conflict` is threaded through both `submit()` branches (`conflict=tuple(buffer.conflict)`).
5. **Full reassembly test suite**: `pytest tests/foundation/reassembly/` -- 64 passed, 18 subtests, including every new `TCPReassemblyConflictTests` case and the six-boundary-case coordinate tests from the prior review round.
6. **mypy**, three flag sets, all pointing at the same single pre-existing, unrelated error: `--follow-imports=silent --ignore-missing-imports --show-column-numbers --show-error-codes` (the Makefile's own `mypy:` target), `--config-file mypy.ini` alone, and no flags at all. All three report only `tcp.py:467: error: Argument 2 to "Deferred" has incompatible type ... [arg-type]` on these two files -- and a diff against the merge-base shows that exact line is byte-for-byte unchanged by this PR. Zero mypy errors introduced.
7. **pylint -- checked, and partially refutes the "8.45 -> 8.19" claim as stated.** Ran the project's actual `make pylint` flags (from the `Makefile`) at this commit and at the merge-base `5182ad0ce` (exported via `git archive` to a scratch dir, no shared state touched). `too-many-branches`/`too-many-statements` cannot fire under those flags at all -- both "belong to the design checker" per pylint's own `--help-msg`, and the Makefile's `--enable=...,R,...` is followed later by `--disable=...,design,...`, which wins. Confirmed empirically: zero matches for either check, or for either changed file, anywhere in the full-package run. Score under those flags: merge-base 8.57/10, head 8.56/10 (-0.01, not -0.26). Under *default* pylint (no flags), the two checks do fire, but they are not new: at the merge-base `reassembly()` was already over both thresholds (19/12 branches, 58/50 statements); this PR moves it to 20/12 and 76/50. Real, but pre-existing and off the project's own gate.
8. **Memory/cost judgement call (not blocking).** `received` doubles the per-fragment footprint -- noted inline, with reasoning for why I think the `bytearray` tradeoff is the right call here rather than an interval list, given TCP reassembly here has no timeout by default.

Every command run and its real output is in the session transcript; the key ones: `git rev-parse HEAD`, `git merge-base main 69e666fe8`, `gh pr checks 478 --watch`, the fixture-generation and `pytest tests/foundation/reassembly/` run (64 passed), the three `mypy` invocations, the two `pylint` invocations (head and merge-base under Makefile flags, plus default-flag runs at both), and two ad hoc scripts (`/tmp/probe_reasm.py`, `/tmp/probe_crossbucket.py`) exercising the mechanism directly against this worktree's `pcapkit`.

I did **not** attempt the revert-proof described in the review brief -- it requires editing tracked source, which is outside a review-only mandate; I did not check whether it still holds.

Three inline comments posted: a memory/cost tradeoff note on `Fragment.received` (judgement call, not a defect), a correction of the pylint claim anchored near `reassembly()`, and a confirmation that the corrected reach-back comment is verifiably true. None of the three are blocking.

**GOOD TO MERGE** at `69e666fe8`

docs/source/pcapkit/foundation/reassembly/tcp.rst
```diff
@@ -256,8 +271,32 @@ Terminology
...
+           |                                      |                 |                  an arriving segment disagreed
+           |                                      |                 |                  with bytes already in 'raw'
+           |                                      |                 |                  |--> (tuple) (first, last),
+           |                                      |                 |                               absolute and
+           |                                      |                 |                               inclusive
+           |                                      |                 |                  |--> ...
            |                                      |--> (int) ACK ...
            |                                      |--> ...
+
+       ``received`` is deliberately **not** derived from ``hdl`` above.
```
JarryShaw, line 292: breaks the comment's inline code section it seems

pcapkit/foundation/reassembly/data/tcp.py
JarryShaw: Is there a better approach? I dislike over excess memory as well

pcapkit/foundation/reassembly/tcp.py
JarryShaw, line 193: make the # comments align in one column

JarryShaw: > **Checking the pylint claim from the earlier review round against the project's actual gate.** I ran `make pylint`'s exact invocation (from the `Makefile`) against this commit and against the merge-base (`5182ad0ce`), and the reported "8.45 -> 8.19, two new findings (`too-many-branches`, `too-many-statements`) on `reassembly()`" does not reproduce under those flags.
> 
> - `too-many-branches` (R0912) and `too-many-statements` (R0915) both "belong to the design checker" per pylint's own `--help-msg` text, and the Makefile's `pylint:` target runs `--enable=...,R,...` followed later by `--disable=...,design,...` -- the later disable wins for any message belonging to that checker, so these two checks cannot fire under `make pylint` at all, at any commit. Confirmed empirically too: grepping the full `make pylint`-flagged run at this sha for `too-many-branches`, `too-many-statements`, or `reassembly/tcp.py` returns nothing -- there are zero pylint findings of any kind on either changed file under those flags.
> - Score under those same flags: merge-base 8.57/10, this commit 8.56/10 -- a -0.01 delta, not -0.26.
> - Under *default* pylint (no Makefile flags), the two complexity checks do fire, but they are not new: at the merge-base, `reassembly()` (this function) was already over both thresholds (branches 19/12, statements 58/50); this PR moves it to 20/12 and 76/50. Real complexity, worth tracking, but it predates this change and doesn't register on the gate the project actually runs.
> 
> Not a merge blocker in my view, but a follow-up to split this function (e.g. extracting the append/prepend/overlap branches per direction into helpers) is reasonable given it was already over the default thresholds before this PR and has grown further.

Let's get this addressed in this PR

pcapkit/foundation/reassembly/data/tcp.py
JarryShaw: Measured the alternatives rather than guessing, and there is a better approach — the current one is paying for the common case to serve the rare one.

`sys.getsizeof`, per fragment, against what each representation costs to record the same fact:

| payload | gaps | `bytearray` mask | bit-packed | gap-interval list |
|---|---|---|---|---|
| 1,500 B | 0 | 1,557 | 245 | 56 |
| 64,000 B | 0 | 64,057 | 8,057 | 56 |
| 1,000,000 B | 0 | **1,000,057** | 125,057 | **56** |
| 1,000,000 B | 10 | 1,000,057 | 125,057 | 776 |
| 1,000,000 B | 1,000 | 1,000,057 | 125,057 | 72,056 |
| 1,000,000 B | 50,000 | 1,000,057 | 125,057 | 3,600,056 |

The shape of the problem: **a clean stream costs a megabyte of mask to record "all received"**, which a 56-byte empty list records exactly as well. Bit-packing is a flat 8× win and keeps the array semantics, so it is the cheap change. But the interval list is asymptotically better wherever it matters and only loses on pathologically fragmented streams — the crossover in the table above is somewhere past a thousand distinct gaps in a megabyte, which is a capture you would already be unhappy about.

**What I would actually do: track the gaps, not the receipts.** The two are complements, and the gaps are the smaller set by construction — they exist only where `reassembly()` inserts zero-fill, which is exactly two places, `bytearray(GAP)` on the forward-append path and on the reach-back path. "Has this fragment received position *p*" becomes "*p* is in no gap interval". For a stream with no gaps at all — the overwhelmingly common case, and the one where `received` is currently pure overhead — the structure is empty and costs nothing.

That is also a smaller change than it sounds, because the invariant is narrower: instead of maintaining a mask in lockstep with every slice operation on `raw` (which is what made the alignment question hard enough to need a 200-trial fuzz test), you record an interval at each of the two sites that create fill, and subtract from it when fill is overwritten by real bytes.

Three things I want to flag honestly before you pick:

- **This is a rewrite of the mechanism, not a tweak.** The current `received: bytearray` is reviewed, fuzz-verified for alignment across 200 random-segment trials, and demonstrably fixes the data loss. Swapping it for intervals means re-earning that confidence, including a fresh revert-proof and fresh alignment testing.
- **Bit-packing is the low-risk option** if you want the memory back now: 8× less, the same array semantics, the same alignment invariant, and the existing fuzz test still covers it. The cost is index arithmetic at every access.
- **`conflict` itself is already an interval list**, so intervals-for-gaps would make the two halves of this feature use one representation instead of two.

My recommendation: **bit-pack now, inside this PR, and file the gap-interval rewrite as its own issue.** It gets you the 8× immediately against a mechanism that is already verified, and keeps a rewrite of just-reviewed reassembly code out of a PR that is otherwise ready. But if you would rather have the better structure once rather than twice, say so and I will do the interval version here instead — it is your call on how much churn this path is worth.

JarryShaw: We've already used gap calculation in TCP reassembly logic. We should keep on that

pcapkit/foundation/reassembly/tcp.py
JarryShaw: Fixed, and this finding was correct in every particular — it is the best catch of the session.

`_merge_overlap` no longer consults `hdl` at all. `Fragment` carries its own per-octet `received` mask, so the question "has *this* fragment received a byte here" is answered from this fragment's own state rather than inferred from something every ACK bucket shares. The docstring now says the mask is deliberately **not** derived from `hdl`, and why, so the next reader does not re-derive the shortcut.

Verified before/after on your own cross-bucket scenario, two detached worktrees at the exact shas, `pcapkit.__file__` asserted in each:

```
274df0651 : b'AAAA\x00\x00\x00\x00\x00\x00DDDDEEEE'   conflict=((104, 109),)
69e666fe8 : b'AAAACCCCCCDDDDEEEE'                      conflict=()
e80c42217 : b'AAAACCCCCCDDDDEEEE'                      (pre-PR main, lossless)
```

Six genuinely-received bytes no longer discarded, and no phantom conflict recorded where there was no disagreement. Your point that this was **strictly worse than the last-write-wins it replaced** was the thing that made it a blocker rather than a nit — last-write-wins was at least lossless.

Two things worth recording about why the PR's own tests could not catch it: all six boundary cases used a single ACK bucket, so the defect was structurally outside their reach, not merely missed. The new `test_one_ack_buckets_hole_closing_does_not_leak_receipt_into_another` covers two buckets under one BUFID. I ran the revert-proof myself, since a review-only mandate cannot: reverting `tcp.py` and `data/tcp.py` to `274df0651` while keeping the new tests fails 3, and restoring gives 23 passed / 5 subtests.

Resolving. The memory cost of the mask is being addressed separately in this PR, per the owner's decision on that thread.

pcapkit/foundation/reassembly/tcp.py
JarryShaw: Fixed — the comment was wrong and you disproved it with a case, which is the right way to retire a false claim.

The parenthetical asserting a head-prepend and a tail-append "cannot both happen at once" is gone. It now states what is actually mutually exclusive: with `OVERLAP = min(len(RAW), LEN - OFFSET)`, exactly one of `RAW[OVERLAP:]` (old tail preserved) or `info.payload[OFFSET + OVERLAP:]` (new tail appended) is non-empty — which is a real invariant, unlike the one it replaced.

The code was always correct here; only the prose was false. That distinction matters for the record, because a wrong comment next to right code is the kind of thing that survives for years and then misleads someone into "simplifying" the code to match it.

Independently confirmed by #478's re-reviewer, which built the simultaneous head-prepend-and-tail-extension case directly (old buffer `isn=100 len=10`, new segment `dsn=90 len=30`, engulfing and extending past both ends) and checked the replacement claim holds algebraically.

Resolving.

pcapkit/foundation/reassembly/data/tcp.py
JarryShaw: Taking this as the decision it is, and it is not yet in this PR — so this thread stays open until it is.

Your reasoning is better than my recommendation and I want to record why, since I had argued for bit-packing. `reassembly()` **already computes gaps**: the two `GAP` expressions are the only places zero-fill enters `raw` at all — the forward-append `GAP = PSN - (ISN + LEN)` and the reach-back `GAP = ISN - (PSN + LEN)`. Recording intervals therefore reuses a quantity the algorithm derives anyway, where a mask of either width invents a second parallel representation of the same fact. And storing the intervals in **absolute sequence numbers** matches what `conflict` already does, so both halves end up in one coordinate system instead of two.

The work exists and is measured, on ``wip/443-gap-intervals`` at ``fc968d6dd`` — deliberately parked off this branch so a half-applied representation change could not sit on a PR that was otherwise reviewable. What it establishes so far:

- **The alignment problem disappears, which was the real risk.** The old mask had to be re-prefixed when a reach-back prepend revised `isn` downward — that shifting is exactly the class of bug the earlier `274df0651` regression came from. With absolute intervals nothing needs shifting: the gap list is untouched by the `isn` revision. I predicted that and asked to be told plainly if it turned out false; it held.
- **`conflict` output is unchanged** on every case reached so far — a cross-bucket stream still yielding `()`, a conflicting retransmission still `((100, 107),)`, an identical retransmission `()`, a plain gap-fill `()`. That is the property that must hold: this is an internal representation change, not a behaviour change.
- The pre-existing reassembly suite is still exactly 64 passed / 18 subtests with the new tests deselected.
- Memory, per fragment, 1 MB payload with no gaps: **1,000,057 bytes for the bytearray, 125,057 bit-packed, 56 for an empty gap list.** A clean stream was paying a megabyte to record "everything arrived". At the gappy end my figure and the author's disagree (776 vs 824 at ten gaps) and the explanation offered for the difference does not hold — `sys.getsizeof((1,2))` is 64 here for both of us — so I am treating the gappy numbers as unsettled and having them re-derived rather than quoting either.

Three things are still open on that branch and I am not pretending otherwise: no revert-proof on the new `_merge_overlap`, `docs/source/pcapkit/foundation/reassembly/tcp.rst` still documents the `received` field that the change removes, and the full suite has not been re-run. It is under independent review now, with the reviewer specifically asked to hunt for a `conflict` divergence rather than re-run the author's own cases, and to say whether the alignment invariant is genuinely gone or merely unhandled.

Once that comes back clean it lands here, on this branch, and I will close this thread against the merged code rather than against a promise.

pcapkit/foundation/reassembly/tcp.py
JarryShaw: "Let's get this addressed in this PR" — understood, and I owe you an apology for how long this sat. I had been treating this thread as informational because the comment above it concluded "not a merge blocker", and I did not re-read your reply underneath it. That is my error: your instruction is the last word on the thread, and I have been reporting this PR as merge-ready for several checks while it was outstanding.

So, to be explicit about what is being addressed, because the thread contains two different claims and only one of them is real:

**The gate delta is not real** and I am not going to manufacture work to chase it. `R0912`/`R0915` belong to pylint's `design` checker, and the `Makefile`'s `pylint:` target enables `R` and then *later* disables `design`, so the later disable wins and neither check can fire under `make pylint` at any commit. The measured score delta on the project's actual gate is 8.57 → 8.56, which is -0.01 rather than the -0.26 the earlier round reported. That part of the earlier review was wrong and stays refuted.

**The complexity is real**, and that is what gets fixed. Under default pylint the two checks do fire, and `reassembly()` was already over both thresholds before this PR — branches 19 against a limit of 12, statements 58 against 50 — with this change taking it to 20 and **76**. The statement count is the striking one: the function has grown by roughly a third and is now over 50% past the threshold. A function that size mixing append, prepend, overlap and submission concerns is how the `274df0651` data-loss regression got in and stayed invisible through six boundary tests, so the split is worth doing on its merits rather than to satisfy a checker that is switched off.

The plan, sequenced deliberately in that order:

1. Land the gap-interval representation from the thread above, since it rewrites the interior of this same function — splitting first and then rewriting the pieces would mean doing the same work twice and reviewing it twice.
2. Extract the append, prepend and overlap branches into named helpers, so `reassembly()` reads as the dispatch it actually is.
3. Re-measure default-pylint branches and statements, and report the before/after numbers here rather than asserting it is better.

The constraint on the refactor: **the visible behaviour must not move.** `conflict`, `completed`, payload bytes and the `Datagram` shape all stay exactly as they are, verified against the current branch rather than against the tests alone — a pure extraction that changes output is not a refactor. I will close this thread with those numbers and that evidence once it lands, not before.

pcapkit/foundation/reassembly/data/tcp.py
JarryShaw: Landed at `7ce6c200d` (representation commit `2a795938c`). `Fragment.received` is gone; `Fragment.gap: list[tuple[int, int]]` replaces it at `pcapkit/foundation/reassembly/data/tcp.py:206`.

Your reasoning was better than my recommendation, and it is worth saying why. `reassembly()` **already computes gaps** — the two `GAP` expressions are the only places zero-fill enters `raw` — so intervals reuse a quantity the algorithm derives anyway, where a mask of any width invents a second parallel representation of the same fact. Storing them in **absolute** sequence numbers also puts them in the same coordinate system as `conflict`, instead of two.

**The alignment problem the mask had is structurally gone, not merely unhandled.** The old per-octet mask had to be re-prefixed whenever a reach-back prepend revised `isn` downward — exactly the shifting that produced the `274df0651` regression. With absolute intervals nothing indexes a gap entry relative to `raw`, so no fixup exists to get wrong. Verified by inspecting internal state through two successive `isn` revisions (`base+100 → base+40 → base`): an entry created under an `isn` that no longer exists stays present, unmodified, and still correct against `raw` at the new `isn`, with no shifting path executed.

**I verified the visible output is unchanged myself**, running identical inputs through two detached worktrees at `a1c50d4df` and `7ce6c200d` and diffing:

```
conflict_same_bucket:  (b'AAAAAAAACCCC', ((100, 107),))
identical_retrans:     (b'AAAAAAAACCCC', ())
plain_gap_fill:        (b'AAAABBBBCCCC', ())
reach_back:            (b'AAAA' + 96 zero octets + b'DDDD', ())
double_reach_back:     (b'AAAA' + zeros + b'DDDD' + zeros + b'EEEE', ())
exact_boundary_fill:   (b'AACD', ((101, 101),))
TWO_ACK_BUCKETS:       (b'AAAA', ((100, 103),))  (b'CCCC', ())  (b'XXXXYYYY', ())
```

`diff` exit 0 — byte-identical on all seven, zero exceptions. The two-bucket case matters most: it is the shape this PR's original six boundary cases structurally could not reach, all having used a single ACK bucket, which is why they missed the cross-bucket data loss earlier. Here one bucket reports its conflict and the other reports `()`, so receipt state stays private per bucket.

An independent review of the same change across 11 cases — seven of its own design, including a segment spanning two separate gaps in one merge call, a position that is a gap then closed then retouched with different bytes, and `GAP == 0` on both paths — also found byte-identity and no divergence.

Tests at `7ce6c200d`, measured in a fresh worktree with fixtures regenerated and `pcapkit.__file__` asserted against it: `tests/foundation/reassembly/test_tcp.py` **25 passed / 205 subtests**; `tests/foundation/reassembly/` **73 passed / 218 subtests**.

**On memory — correcting a figure I gave you earlier, which was wrong.** I reported 776 bytes for ten gaps and said it reproduced. It reproduces my own formula, which is not the same as being right: the formula used `sys.getsizeof([])` = 56 for a container that over-allocates to **184** when built by `append`, and assumed 72 bytes per tuple where the real cost is **64**. Measured properly on the repo venv, Python 3.14.7:

| | bytes |
|---|---|
| clean 1 MB stream, old `bytearray` mask | **1,000,057** |
| clean 1 MB stream, empty gap list | **56** |
| 10 gaps — list container | 184 |
| 10 gaps — container + tuples | **824** |
| 10 gaps — container + tuples + int objects | 1464 |

So the megabyte-to-56-bytes case that motivated this is solid, and the earlier "824 versus 776" disagreement was my error rather than a real discrepancy.

`docs/source/pcapkit/foundation/reassembly/tcp.rst` is updated rather than renamed, since the semantics **invert** — the old field recorded what had been received, `gap` records what is still missing. The ASCII block now reads `'gap' : (list) sequence ranges still zero-fill placeholder in 'raw'` with `(tuple) (first, last)`, and the prose records both why it is not derived from `hdl` and that it follows `conflict`'s absolute-inclusive convention.

Resolving.

pcapkit/foundation/reassembly/tcp.py
JarryShaw: Landed at `7ce6c200d` (split commit `9ecbcee70`). `TCP.reassembly()` now dispatches to three private helpers in `pcapkit/foundation/reassembly/tcp.py`:

```
102:  def reassembly(self, info)                          # the dispatch
214:  def _reassemble_append(self, info, fragment, PSN)
266:  def _reassemble_prepend(self, info, fragment, PSN)
331:  def _update_hole_descriptors(self, info, BUFID, ...)
371:  @staticmethod
372:  def _merge_overlap(gap, old, new, ...)
```

**The before/after you asked for, measured rather than asserted.** Default pylint on `pcapkit/foundation/reassembly/tcp.py`, in two detached worktrees:

```
a1c50d4df  R0912: Too many branches (20/12)     (too-many-branches)
           R0915: Too many statements (76/50)   (too-many-statements)

7ce6c200d  (neither message fires)
```

Both are gone outright, not merely reduced — 76 statements against a limit of 50 was over 50% past it.

**And the part of the original claim that was wrong stays refuted rather than quietly fixed.** The "8.45 → 8.19, two new findings" from the earlier review round cannot happen under this project's actual gate: `R0912` and `R0915` belong to pylint's `design` checker, and the `Makefile`'s `pylint:` target enables `R` and then *later* disables `design`, so the later disable wins and neither check can fire under `make pylint` at any commit. The measured score delta on that gate was **8.57 → 8.56**, i.e. -0.01 rather than -0.26. So nothing here was done to satisfy a switched-off checker, and **no suppressions, pragmas or config changes were added** — the complexity was real under default pylint and predated this PR, which is why it was worth fixing on its merits.

**The refactor is behaviour-preserving, established against the code rather than against the tests** — the tests did not catch the last regression in this file either. Identical inputs through detached worktrees at `a1c50d4df` and `7ce6c200d`, diffed:

```
conflict_same_bucket:  (b'AAAAAAAACCCC', ((100, 107),))
identical_retrans:     (b'AAAAAAAACCCC', ())
plain_gap_fill:        (b'AAAABBBBCCCC', ())
reach_back:            (b'AAAA' + 96 zero octets + b'DDDD', ())
double_reach_back:     (b'AAAA' + zeros + b'DDDD' + zeros + b'EEEE', ())
exact_boundary_fill:   (b'AACD', ((101, 101),))
TWO_ACK_BUCKETS:       (b'AAAA', ((100, 103),))  (b'CCCC', ())  (b'XXXXYYYY', ())
```

`diff` exit 0 across all seven, zero exceptions — `conflict`, payload bytes and `completed` all unchanged, including the two-ACK-bucket case that this PR's original tests could not reach.

Suites at `7ce6c200d`, fresh worktree with fixtures regenerated and `pcapkit.__file__` asserted against it: `test_tcp.py` **25 passed / 205 subtests**, `tests/foundation/reassembly/` **73 passed / 218 subtests**.

One note on ordering, since it was deliberate: the representation change from the thread above went in **first** and the split second. Splitting first would have meant refactoring branch bodies that the representation change then rewrote — the same work, and the same review, twice.

Resolving.

pcapkit/foundation/reassembly/tcp.py
```diff
@@ -190,51 +190,52 @@ def reassembly(self, info: 'Packet') -> 'None':
                 # record fragment payload
                 ISN = self._buffer[BUFID].ack[ACK].isn        # Initial Sequence Number
                 RAW = self._buffer[BUFID].ack[ACK].raw        # Raw Payload Data
-                RCVD = self._buffer[BUFID].ack[ACK].received  # this fragment's own received mask
+                GAPS = self._buffer[BUFID].ack[ACK].gap        # this fragment's own gap list
```
JarryShaw: align the `#` comments on ssame column

JarryShaw: Done at `4f7063f4f`.

Worth saying plainly that this was reported back to me as needing no change — the claim being that the extraction had superseded the flagged line and that every remaining comment group was already internally consistent. I measured the columns rather than accepting that, and it was wrong. Each of the two extracted helpers opened with a block carrying comments at **three** different columns:

```
        ISN = fragment.isn       # col 29
        RAW = fragment.raw       # col 29
        GAPS = fragment.gap      # col 29
        LEN = fragment.len
        GAP = PSN - (ISN + LEN)  # col 36
        if GAP >= 0:             # col 24
```

Every comment was individually one space from its own code, which is how it passed a by-eye check, but the block reads as ragged — which is the thing you were pointing at.

Both blocks now align at **column 33**, two past the longest line in the run (the `GAP` expression). The two helpers are also now identical where they are parallel: `_reassemble_prepend`'s guard had been left at 24 because an intervening ``fragment.__update__`` call separates it from the rest of its run, so a purely adjacency-based pass missed it.

Two things left alone deliberately, rather than silently:

* The lone comment on the past-the-buffered-end guard in ``_reassemble_append`` (`# fragment reaches past the buffered end`). It is a single-comment run with no counterpart in the sibling helper, so a shared column is vacuous for it and forcing it out to 51 would leave a long gap after a short statement.
* The ``OFFSET``/``OVERLAP`` pairs and the ``__update__`` keyword pairs, already internally aligned at 51 and 27 respectively.

For completeness, the columns across all four functions now:

```
reassembly                 25 x1, 29 x7, 41 x2
_reassemble_append         27 x2, 33 x5, 36 x1, 51 x2
_reassemble_prepend        27 x2, 33 x5,        51 x2
_update_hole_descriptors   63 x8
```

**Comment-only, verified mechanically rather than asserted**: stripping trailing `#` comments and all whitespace from every added and removed line leaves the two sides identical, so no code changed. 10 insertions, 10 deletions, one file. `tests/foundation/reassembly/test_tcp.py` unchanged at **25 passed / 205 subtests**.

A comment-only follow-up does not affect the review state, so the re-review already running on the two code commits still stands on its own. Resolving.

JarryShaw:
Re-reviewed at `583a5d1a7` (confirmed via `git rev-parse HEAD` after `git checkout 583a5d1a7`), specifically the two commits that voided the prior GOOD TO MERGE: `2a795938c` (`Fragment.received: bytearray` → `Fragment.gap: list[tuple[int, int]]`) and `7ce6c200d` (splitting `TCP.reassembly()` into `_reassemble_append`/`_reassemble_prepend`/`_update_hole_descriptors`).

**1. Is the split a pure extraction?** Read `reassembly()` and the three new helpers line by line against `a1c50d4df`'s single function. Both `_reassemble_append`/`_reassemble_prepend` capture `ISN`/`RAW`/`GAPS` as locals with exactly the same scoping the inline code had -- in particular, `_reassemble_prepend` captures the *old* `ISN` before calling `fragment.__update__(isn=PSN)`, matching the original's local-variable capture-before-mutate order exactly. The mutual exclusivity of "old tail survives" vs "new tail survives" in the reach-back overlap branch is unchanged (`OVERLAP = min(len(RAW), LEN - OFFSET)` caps it identically). No state is read after being mutated out of order, no early return changes control flow, and `_update_hole_descriptors` is a verbatim lift of the `HDL` loop with no logic touched. Found no divergence from a pure extraction.

**2. `conflict` divergence hunt on cases I built myself** (two detached worktrees, `a1c50d4df` vs `583a5d1a7`, identical driver against both):

- A segment spanning **two separate existing gaps plus the two real-data ranges between them**, in one `_merge_overlap` call (`AAAAA` real, gap, `BBBBB` real, gap, `CCCCC` real, then one 20-byte segment covering all of it with conflicting bytes) → identical on both trees: `raw=AAAAAXXXXXBBBBBXXXXXCCCCC`, `conflict=[(base+10,base+14),(base+20,base+24)]`; new tree's `gap=[]` matches the old tree's received-mask-derived holes (`[]`).
- A gap **closed** by one segment, then **retouched** by a later segment with *different* bytes → identical `conflict=[(base+5,base+9)]` on both -- confirms the closed gap entry is actually retired rather than merely masked, since a live gap entry there would have suppressed the conflict.
- `GAP == 0` on **both** the forward-append and reach-back-prepend paths (exact abutment, no hole) → `conflict=[]`, `gap=[]` on both trees. Also confirmed by code reading: both helpers guard `GAPS.append(...)` with `if GAP > 0:`, so a `GAP == 0` call never appends an inverted `(x, x-1)` entry.
- A **conflict + a partial gap-fill + a new-head-prepend, all in the same `_merge_overlap` call** (prepend side: old fragment `isn+20..24`, prepend creates a gap at `isn+15..19`, then a reach-back segment supplies a new head, disagrees with the old real data, and partially fills the gap, all at once) → byte-identical `raw`, identical single `conflict` range, and the new tree's surviving trimmed `gap` entry (`(isn+18, isn+19)`) matches the old tree's received-mask-derived holes exactly.

All 6 cases matched to the byte; none diverged. Ran a 5000-trial randomized stress test on top of that (biased toward `GAP==0` and negative/overlapping offsets, single bucket) checking every `gap` entry is genuinely all-zero in `raw`: 0 violations.

**3. Interval hygiene.** 5000 randomized trials, checked for (a) any zero-length or inverted `gap` entry ever stored, (b) any two `gap` entries adjacent or overlapping, (c) every `gap` entry genuinely all-zero in `raw` at the time it's checked. **Zero violations.** This confirms the reasoning: a new `gap` entry is only ever created at one of the two current edges of the buffer (the far end in `_reassemble_append`, the near end in `_reassemble_prepend`), and both of those edges are always real payload by construction (the buffer only ever grows by splicing real payload at its extremities), so a fresh entry never lands adjacent to another gap entry, and `_merge_overlap`'s trim-or-remove logic never produces a zero-length remnant (the `first < lo` / `last > hi` guards ensure a surviving slice is always non-empty).

**4. Test adequacy -- enumerated.** Exactly **one** test in `test_tcp.py` uses two or more ACK buckets: `test_one_ack_buckets_hole_closing_does_not_leak_receipt_into_another` (`ack=1000` and `ack=2000`), and that test predates this PR's two reviewed commits -- it's part of the earlier `274df0651`/`69e666fe8` work and asserts only `.conflict`/`.payload`, never `.gap`. **Both tests added by the two commits under review here are single-ACK-bucket only**: `test_a_simultaneous_head_prepend_and_tail_extension_in_one_overlap_call` and `test_gap_list_matches_the_zero_filled_positions_across_200_random_trials` (the latter's own docstring says "segments fed into a single ACK bucket"). So there is no committed test that drives the new `gap` mechanism across multiple ACK buckets. I did **not** build a dedicated multi-bucket probe of `.gap` itself -- what I have is a code-reading argument: `_reassemble_append`, `_reassemble_prepend`, and `_merge_overlap` only ever read or write `fragment.gap`, the specific ACK bucket's own `Fragment`, never `self._buffer[BUFID].hdl` or any other bucket's `Fragment`, so the isolation the existing multi-bucket test proved for the old `received`/`conflict` fields extends to `gap` by the same code path, not by a fresh test I ran. Worth a follow-up test; not something I found broken.

**5. `test_models.py`.** Diffed the PR head against `origin/main`'s tip directly. `test_ip_data_models_and_package_aliases` already carries main's `conflict` field (from #482, unrelated to this PR) untouched by this PR. `test_tcp_data_models_and_package_aliases` is the only section this PR touched, updating `Fragment(...)`'s 5th positional arg from a `received` bytearray to `gap=[]`. No corruption, no cross-contamination between the IP and TCP sections.

**6. Docs.** Read the full 374-line `tcp.rst`. The ASCII `.. code-block:: text` block (lines 250-291) is intact and correctly indented; the two `gap`-explanation paragraphs sit outside the block, not wedged into it at the wrong indent. `automethod::` lists only the two public methods (`reassembly`, `submit`); the new private helpers aren't autodoc'd, consistent with the pre-existing `_merge_overlap` convention. All `autoclass::`/`.. module::`/`.. currentmodule::` directives name the real defining modules (`pcapkit.foundation.reassembly.data.tcp`, `pcapkit.foundation.reassembly.tcp`), not a re-export. RFC 815's `RCVBT` and the one English use of "received" in prose are the only surviving occurrences of the word, both legitimate.

**Measurements I ran myself** (fixtures generated via `PYTHONPATH=<worktree> .../make_samples.py`, confirmed `pcapkit.__file__` resolved into the worktree before every run):

- `pytest tests/foundation/reassembly/test_tcp.py -v`: **25 passed, 205 subtests passed**, 0 failures.
- `pytest tests/foundation/reassembly/ -v`: **75 passed, 218 subtests passed**, 0 failures, 0 skips (the subtest count matches the quoted baseline of 218; the top-level pass count I measured is 75 rather than 73, but there are no failures either way so it doesn't change the picture).
- `pylint pcapkit/foundation/reassembly/tcp.py --disable=all --enable=R0912,R0915`: **10.00/10** -- neither too-many-branches nor too-many-statements fires post-split (previous run 7.73/10 on the pre-split file).

**CI.** The PR moved twice during this review (the owner pushed follow-ups): `583a5d1a7` -> `b248e3651` (test-only commit touching only `tests/toolkit/test_scapy_unit.py`, unrelated to this PR's files) -> `4f7063f4f` (a comment-alignment-only commit touching `pcapkit/foundation/reassembly/tcp.py`, verified by diff to contain zero code changes -- only trailing-comment columns shift). Each push cancelled the in-flight run for the prior sha via GitHub's concurrency group, which is why `583a5d1a7` itself never shows a completed run of its own. I polled the now-stable head `4f7063f4f` to full settlement: **21 success + 2 skipped-by-design (`Gate (full suite, Python 3.14)`, `Docs test gate`) = 23/23 checks, 0 failures, 0 pending, 0 cancelled.**

**Verdict.** No code defect found in the split or the gap-list rewrite -- every adversarial case built to try to break it (spanning two gaps in one call, retouching a closed gap, both `GAP==0` paths, conflict-plus-gap-fill-plus-new-head-prepend in one call, and 5000 randomized hygiene trials) matched the pre-rewrite behavior exactly, and CI is fully green on the current head. The one real, actionable gap is test coverage, not correctness: no committed test drives the `gap` mechanism through two-or-more ACK buckets. Worth a follow-up test; not a blocker.

GOOD TO MERGE `583a5d1a7` (the two commits that followed during this review -- `23bddffd5`/`b248e3651` and `4f7063f4f` -- touch only an unrelated test file and comment formatting respectively, change no reassembly logic, and CI is green on the resulting head `4f7063f4f`).

PR #481, opened 2026-09-18
mh: reject a wrong-type MN-ID identifier in-library, not as a stdlib leak

### LINKED ISSUE #469: MH MN-ID: bytes and str are documented interchangeably but each subtype accepts only one, and the wrong one fails with a bare stdlib exception
## Summary

`MH._make_opt_mn_id` documents `bytes | str` as accepted identifier types for
every MN-ID subtype, but each subtype's field accepts only **one** of the two.
Passing the other builds a schema that cannot be packed, and fails with a bare
stdlib exception rather than an in-library one.

This is the same family as #467 (which covered `int`) and #448 (which covered
the `IPv6_Address` width), but it is about `bytes`/`str` and is not fixed by
either.

## Reproduction

Measured on `fa128959e` (current `main`). Identical on PR #468's branch, so
**pre-existing and not a regression from that PR** — I checked both trees.

```python
from pcapkit.const.mh.option import Option
from pcapkit.const.mh.mn_id_subtype import MNIDSubtype
from pcapkit.protocols.internet.mh import MH

proto = object.__new__(MH)

# NAI's field is text, so bytes cannot be encoded
proto._make_opt_mn_id(Option.MN_ID_OPTION_TYPE,
                      subtype=MNIDSubtype.NAI,
                      identifier=b'node@example').pack()

# IMSI's field is raw octets, so str cannot be packed
proto._make_opt_mn_id(Option.MN_ID_OPTION_TYPE,
                      subtype=MNIDSubtype.IMSI,
                      identifier='12345').pack()
```

```
NAI  + bytes -> AttributeError: 'bytes' object has no attribute 'encode'
IMSI + str   -> struct.error: argument for 's' must be a bytes object
NAI  + str   -> length=13, packs 15 octets            (correct)
```

So the defect is symmetric: `bytes` fails for the one text subtype, and `str`
fails for the octet subtypes. Only the matching pairing works.

Both failures happen at `pack()` rather than at construction, and neither is an
`instance of `pcapkit.utilities.exceptions.BaseError`, so a caller cannot catch
them with the library's own exception hierarchy.

## Why it matters

The type union is documented, not accidental — `_make_opt_mn_id`'s `identifier`
parameter is annotated `bytes | str | IPv6Address | int`, and the
`Schema_MNIDOption.__init__` stub carries a matching union. So the signature
promises `bytes` and `str` interchangeably for all subtypes when in fact the
choice is fixed per subtype by whether that subtype's field is a `StringField`
or a `BytesField`.

`NAI` is the only text subtype; `IMSI`, `P_TMSI`, `EUI_48_address`,
`EUI_64_address`, `GUTI` and `DUID` are octet subtypes.

## Suggested direction

The same shape as #467's fix: decide per subtype, then either convert (encode a
`str` for the octet subtypes, decode `bytes` for NAI — noting that an encoding
has to be chosen and named if so) or reject with a `ProtocolError` saying which
type that subtype takes. #467's fix already added exactly that kind of
per-subtype rejection for `int`, so the mechanism and message style are in
place to extend.

Whichever is chosen, the documented union should end up describing what is
actually accepted.

## Provenance

Surfaced while verifying #468 (the fix for #467). Its author flagged the
`bytes`-for-NAI half and correctly left it out of scope; the `str`-for-octet-
subtypes half is mine, found while measuring the first. Both verified on `main`
and on the PR branch before filing.

JarryShaw:
Closes #469

## Summary

`_make_opt_mn_id` documents `identifier` as `bytes | str | IPv6Address | int`,
but each MN-ID subtype's field (`mn_id_selector`) accepts only one of the first
two: a `StringField` for `NAI`, a `BytesField` for the other six. Handing the
wrong one -- or a value of neither type -- escaped as a bare stdlib exception
instead of a `ProtocolError`. #468 already closed the `int` half of this
family (#467); this PR closes the rest, per #469's own scoping ("only its
remaining half" -- the `int` half is untouched).

## What changed

`pcapkit/protocols/internet/mh.py::MH._make_opt_mn_id`:

- The `IPv6_Address` branch now wraps the `ipaddress.IPv6Address(identifier)`
  conversion in `try`/`except ValueError`, re-raising as `ProtocolError`. This
  catches `ipaddress.AddressValueError` (itself a bare `ValueError`) for any
  identifier `ipaddress.IPv6Address` cannot parse, without swallowing the
  sibling `ProtocolError` raised just above it for an out-of-range `int`.
- A new `elif subtype_val == Enum_MNIDSubtype.NAI` branch rejects any
  non-`str` identifier with a `ProtocolError` before the `len()` call that
  used to leak `TypeError`/`AttributeError`.
- The remaining `else` branch (the six octet subtypes) rejects any identifier
  that isn't `bytes`/`bytearray` with a `ProtocolError`, for the same reason.

`tests/protocols/internet/test_mh_unit.py` gains
`test_mh_mn_id_option_rejects_wrong_type_identifier_per_subtype`, sweeping all
eight subtypes against every wrong type (including empty-value boundaries),
and asserting the matching type still works.

## Measurements

Re-measured myself on `e80c42217` (current `main`) before changing anything,
through the public maker, not a hand-built schema:

| subtype | wrong identifier | before | after |
|---|---|---|---|
| `NAI` | `b'\x01\x02'`, `[1,2]`, `{'a':1}` | `AttributeError` | `ProtocolError` |
| `NAI` | `1.5`, `None`, `IPv6Address('::1')` | `TypeError` | `ProtocolError` |
| `NAI` | `b''` (empty) | `AttributeError` | `ProtocolError` |
| six octet subtypes | `'user@realm'`, `[1,2]`, `{'a':1}` | `struct.error` | `ProtocolError` |
| six octet subtypes | `1.5`, `None`, `IPv6Address('::1')` | `TypeError` | `ProtocolError` |
| six octet subtypes | `''` (empty), `memoryview(...)` | `struct.error` | `ProtocolError` |
| `IPv6_Address` | `b'\x01\x02'`, `'user@realm'`, `1.5`, `None`, `[1,2]`, `{'a':1}`, `b''`, `bytearray(16)`, `memoryview(16)` | `ipaddress.AddressValueError` (a `ValueError`) | `ProtocolError` |

Correct-type identifiers are unaffected: `str` for `NAI`, `bytes`/`bytearray`
for the six octet subtypes, an `IPv6Address`/valid `int`/valid `str` for
`IPv6_Address` -- all still construct and pack, including the empty-value
boundary (`NAI` + `''`, the six + `b''`). The `int` half from #468 is
unaffected: valid ints still convert per subtype, and the negative/`NAI`/
`>=2**128` rejections still raise exactly as before.

Revert-proof: reverting only `mh.py` (test file left in place) makes the new
test fail with 64 subtest failures across the wrong-type sweep; restoring the
fix makes it pass again.

**Build**: `tests/protocols/internet/test_mh_unit.py` -> 38 passed, 408
subtests passed. Full suite `tests/` -> 1032 passed, 17 skipped, 1707 subtests
passed. `mypy --config-file mypy.ini pcapkit/protocols/internet/mh.py` -> 97
errors, down from a 98-error baseline on unmodified `main` (the one resolved
is this method's own `len(identifier)` "incompatible type... expected Sized"
complaint, now that `isinstance` narrows the union before each `len()` call);
the remaining 2 are pre-existing and unrelated to this method (lines outside
`_make_opt_mn_id`).

## Judgement call: `bytearray`, `memoryview`, and ASCII `str`

- **`bytearray` is accepted** alongside `bytes` for the six octet subtypes.
  Measured: `struct.pack('Ns', bytearray(...))` already round-trips
  correctly today, so rejecting it would tighten behaviour that was never
  broken, for no benefit.
- **`memoryview` is rejected** even though it looks equally bytes-like.
  Measured: `struct.pack('Ns', memoryview(...))` raises
  `struct.error: argument for 's' must be a bytes object` -- it does not
  actually survive the pack call, so letting it through would just relocate
  the leak rather than fix it.
- **A `str` that happens to be pure ASCII is *not* auto-encoded** for the
  octet subtypes, and **a `bytes` that happens to decode cleanly is *not*
  auto-decoded** for `NAI`. Either direction reintroduces the same "accepts a
  value that means the wrong thing" class of bug #467 removed for `int`,
  just relocated to `bytes`/`str`.

## Out of scope

Nothing further found in this half; no new issues filed.


pcapkit/protocols/internet/mh.py
```diff
@@ -7664,10 +7664,15 @@ def _make_opt_mn_id(self, type: 'Enum_Option', option: 'Optional[Data_MNIDOption
...
                 subtype has a wire form for one), an :obj:`int` of any value
-                with the ``NAI`` subtype, or an :obj:`int` of ``2**128`` or
+                with the ``NAI`` subtype, an :obj:`int` of ``2**128`` or
                 above with the ``IPv6_Address`` subtype (whose wire form is a
                 fixed 16 octets, unlike the other subtypes, which have no
-                ceiling and simply pack into more).
+                ceiling and simply pack into more), or ``identifier`` is of a
+                type its subtype's field cannot hold at all: anything but
+                :obj:`str` for ``NAI``, anything but :obj:`bytes`/
+                :obj:`bytearray` for the other six, or anything
```
JarryShaw, line 7673: Minor doc-precision nit, non-blocking: read in isolation, ``anything but bytes/bytearray for the other six`` implies an ``int`` also raises ``ProtocolError`` here, but it does not -- ``int`` is accepted and converted for these six subtypes (the earlier ``elif isinstance(identifier, int):`` branch intercepts it before this method's own new type gate is ever reached). Confirmed: ``_make_opt_mn_id(subtype=IMSI, identifier=5)`` does not raise.

The ``Args:`` section above already says an ``int`` is accepted for every subtype but ``NAI``, so a careful reader reconciles the two, but the ``Raises:`` bullet on its own overstates what raises -- exactly the category of thing #468 had to correct once (a ``Raises:`` clause not matching actual behavior). Suggest something like ``anything but bytes/bytearray/int for the other six`` (or an explicit note that int is handled separately above).

pcapkit/protocols/internet/mh.py
```diff
@@ -7727,7 +7732,23 @@ def _make_opt_mn_id(self, type: 'Enum_Option', option: 'Optional[Data_MNIDOption
...
+                # ValueError subclass (c.f. #467, #468, #469).
+                try:
+                    identifier = ipaddress.IPv6Address(identifier)
+                except ValueError as error:
+                    raise ProtocolError(
+                        f'{self.alias}: [OptNo {type}] MN-ID subtype IPv6_Address '
+                        f'identifier must be an ipaddress.IPv6Address, an int, or '
+                        f'bytes/str it accepts, not {identifier!r}') from error
             id_len = 16
         elif isinstance(identifier, int):
```
JarryShaw, line 7753: Observation, not a defect in this PR: since ``bool`` is a subclass of ``int`` in Python, ``isinstance(identifier, int)`` here also catches ``True``/``False`` and routes them through the pre-existing (#468) int-conversion path rather than through this PR's new type gates. Measured on this PR's head: ``identifier=True`` for ``IPv6_Address``/the six octet subtypes produces ``NO RAISE`` (``True`` -> ``ipaddress.IPv6Address(1)`` / ``(1).to_bytes(...)`` respectively), and for ``NAI`` it correctly raises ``ProtocolError`` (int is rejected there per #468).

This isn't a new leak -- it's `int`/`bool` aliasing that predates this PR and is explicitly out of scope per the issue's "settled decisions not to reopen" (the int half shipped with #468). Flagging only because a bool identifier silently becoming address ``::1``/``::`` or a 1-octet id is a slightly surprising corner if anyone hits it later; no action requested here.

JarryShaw:
Review at head `37f026846` (verified: `git rev-parse HEAD` after `git checkout --detach FETCH_HEAD` on `refs/pull/481/head` returned `37f0268468894b7d760403a143f4a47af375b361`, matching `headRefOid` exactly — the worktree had started on `main`/`e43ad4354` and I re-detached before measuring anything).

## Scope of the diff

`git diff --stat main...HEAD` (three-dot, merge-base `e80c42217`): only `pcapkit/protocols/internet/mh.py` (+72/-3) and `tests/protocols/internet/test_mh_unit.py` (+112). No other files touched.

## CI

Waited it out rather than judging a partial rollup. Final state, `gh pr view 481 --json statusCheckRollup`:

```
CheckRun  Docs test gate                 SKIPPED
CheckRun  Gate (full suite, Python 3.14) SKIPPED
CheckRun  (all 21 others: Analyze, Compat/Integration/Python 3.10-3.15, deploy-pages, CodeQL) SUCCESS
StatusContext pyup.io/safety-ci          SUCCESS
```
21 SUCCESS + 2 SKIPPED-by-design + 1 StatusContext SUCCESS = fully green.

## Independent verification of the PR's own measurements

All run myself in this worktree, `pcapkit.__file__` asserted to start with the worktree root before every run, fixtures generated first via `PYTHONPATH=<worktree root> .venv/bin/python examples/generators/make_samples.py`.

- **mypy**, `python -m mypy --config-file mypy.ini pcapkit/protocols/internet/mh.py`: **98 errors** on merge-base `e80c42217` (detached, ran, then returned to PR head) → **97 errors** on `37f026846`. The one that disappeared is exactly the claimed one — merge-base has `mh.py:7767: error: Argument 1 to "len" has incompatible type "bytes | str | IPv6Address"; expected "Sized"`, which is gone at the PR head; the remaining 2 (`mh.py:7604` prefix arg-type, `mh.py:7971` CGA-extensions list-invariance) are both outside `_make_opt_mn_id` and present at both revisions.
- **Unit file**: `pytest tests/protocols/internet/test_mh_unit.py -q` → `38 passed, 8 warnings, 408 subtests passed in 37.40s`. Matches exactly.
- **Full suite**: `pytest tests/ -q` → `1032 passed, 17 skipped, 11137 warnings, 1707 subtests passed in 872.08s`. Matches exactly.
- **Revert-proof, run myself** (copied `mh.py` out, overwrote it with `git show e80c42217:pcapkit/protocols/internet/mh.py`, left the new test file in place, ran only the new test, then restored the PR-head `mh.py` byte-for-byte and re-ran to confirm a clean `git status`): reverted → `64 failed, 1 passed, 12 subtests passed`; restored → `1 passed, 76 subtests passed`. The 64-failure figure matches the PR description exactly.

## Attack 1 — the `except ValueError` ordering (mh.py:7717-7752)

Read the control flow rather than trusting the tests alone, as asked. The out-of-range-int `ProtocolError` (line 7731) sits in its own `if` statement, *sequentially before and outside* the `try` (lines 7745-7751) — not nested inside it. Raising unwinds the stack immediately, so if that `if` fires, the `try` below is never entered; structurally, the `except ValueError` cannot see that raise. The `try` body itself contains only the stdlib call `ipaddress.IPv6Address(identifier)` — no in-library raise lives inside it to be swallowed either.

I also independently probed `ipaddress.IPv6Address()` directly against every exotic type I threw at it (`None`, `float`, `list`, `dict`, `set`, `frozenset`, a generator, a `__len__`-only object, an `encode`-only object, `memoryview`, `bytearray`, wrong-length `bytes`) — every single one raises `ipaddress.AddressValueError` (confirmed `isinstance(e, ValueError) == True` in each case), never a bare `TypeError`. So the `except ValueError` is complete against everything reachable, not just the cases the new test happens to try.

Bonus: I checked whether the *existing* (unmodified, #468-era) test would even catch the specific regression this attack worries about — merging the out-of-range check into the same `try`. It would: `test_mh_mn_id_option_converts_int_identifier_per_subtype` asserts `self.assertIn('below 2**128', str(ctx.exception))` for that case, and a merged-try version would produce the generic "must be an ipaddress.IPv6Address..." message instead, failing that assertion even though the exception class would still be `ProtocolError`.

## Attack 2 — completeness of the type gate

Ran an independent 8-subtype × 12-exotic-type sweep (96 combos: `None`, `float`, `list`, `dict`, `set`, `frozenset`, generator, `HasLen()`, `HasEncode()`, `bool True/False`, `memoryview`) through the public maker directly (not the PR's own test). Zero non-`BaseError` leaks. The only "NO RAISE" results were `bool` for the six octet subtypes and `IPv6_Address` (14 of 96) — and that's `bool`-is-`int` in Python (`isinstance(True, int) == True`), routed through the pre-existing, settled #468 int-conversion path before this PR's new gates are ever reached, not a gap this PR introduced. `NAI` correctly still rejects `bool` (int is rejected there too, per #468). Left a non-blocking inline comment noting it.

Also checked bytes/str subclasses and out-of-range/unrecognized `subtype_val` values (0, -1, 300, 999) against the same exotic types: all correctly rejected via `ProtocolError`, no leaks anywhere.

## Attack 3 — the two judgement calls

Verified both claimed measurements independently, directly against `struct.pack`: `struct.pack('4s', bytearray(b'ABCD'))` → `b'ABCD'` (round-trips); `struct.pack('4s', memoryview(b'ABCD'))` → `struct.error: argument for 's' must be a bytes object`. Confirmed exactly as claimed.

I'd also point out the same principle is applied consistently to `IPv6_Address`, where `bytearray` is *rejected* even though it's accepted for the six octet subtypes — this isn't an inconsistency, because CPython's `ipaddress.IPv6Address.__init__` only takes the packed-bytes fast path for an exact `bytes` instance (`isinstance(address, bytes)`); a `bytearray`/`memoryview` of the correct 16-byte length falls through to the `str(address)` fallback and fails with `AddressValueError` regardless (confirmed directly). So "does it survive the actual conversion call this branch uses" is applied uniformly across both branches, not just asserted for one. I agree accepting `bytearray` but not `memoryview` is a defensible, measured line rather than an arbitrary one.

## Attack 4 — would the new test catch a re-break

`test_mh_mn_id_option_rejects_wrong_type_identifier_per_subtype` gives every one of the 8 subtypes its own dedicated wrong-type loop (`NAI`: 1, six octets: 1 shared loop, `IPv6_Address`: 1) — no subtype is left out of a loop, which is precisely the shape of #468's own miss (excluding `IPv6_Address` from the negative-int sweep). Assertions pin `isinstance(ctx.exception, ProtocolError)` (the concrete class, not just `BaseError`) plus a message substring naming the subtype and the accepted type — a real pin, if a slightly looser one than the pre-existing test's full-string `assertIn('below 2**128', ...)`. It also separately pins the `bytearray`-accepted judgement call (`schema.identifier == bytearray(...)`, `schema.pack()[3:] == ...`) and the correct-type/empty-value boundary. I ran this test standalone against the revert and confirmed it fails hard (64 subtests) without the fix.

## Attack 5 — docstring/comment accuracy

Enumerated all 6 distinct `ProtocolError`-raising conditions in the code (negative int; int≥2**128 for `IPv6_Address`; int for `NAI`; type `ipaddress.IPv6Address` itself rejects; non-`str` for `NAI`; non-`bytes`/`bytearray` for the six) against the `Raises:` clause — all 6 are represented, unlike #468's own prior 2-of-3 omission. One precision nit, left inline: "anything but bytes/bytearray for the other six" reads as if `int` also raises there, but it doesn't (int is accepted/converted, intercepted earlier). Non-blocking — the `Args:` section already disambiguates it, but the `Raises:` bullet standing alone overstates it, which is the same category of thing #468 had to fix once.

## Verdict

No blocking defects found. CI is fully green, the diff is scoped to exactly the two files it should touch, every quantitative claim in the PR description reproduced exactly under independent measurement, the exception-swallowing control flow is safe by construction (not just by test), an independent 96-combo exotic-type sweep found no leaks beyond a pre-existing/settled `bool`-is-`int` corner, and the two measured judgement calls (`bytearray` in, `memoryview` out) hold up and are applied consistently across both branches. Two non-blocking nits posted inline (a docstring precision issue and the bool/int observation) — neither affects correctness.

**GOOD TO MERGE** at `37f026846`

pcapkit/protocols/internet/mh.py
JarryShaw: Fix this

pcapkit/protocols/internet/mh.py
JarryShaw: Fix this in the PR

pcapkit/protocols/internet/mh.py
JarryShaw: Fixed at `35834b69b` — you were right that the bullet overstated what raises, and the fix went in verbatim along the lines you suggested.

The `Raises:` bullet now names `int` explicitly, and says why it is there rather than leaving a reader to reconcile it against `Args:` two paragraphs up:

```
                :obj:`str` for ``NAI``, anything but :obj:`bytes`/
                :obj:`bytearray`/:obj:`int` for the other six -- an :obj:`int`
                is converted rather than rejected there, per #468 -- or anything
                :class:`ipaddress.IPv6Address` itself does not accept for
```

Your measurement re-confirmed at the current head, forcing the interpreter onto this tree (`sys.path.insert` + asserting `pcapkit.__file__` before importing):

```
IMSI identifier=5 -> b'\x05'
```

So `int` is accepted and converted for the six octet subtypes exactly as you said, and the docs now state it. This was the category of defect #468 had to correct once already — a `Raises:` clause not matching behaviour — which is why it was worth fixing rather than filing.

Resolving.

pcapkit/protocols/internet/mh.py
JarryShaw: Fixed at `35834b69b`. You filed this as "no action requested", and it has been actioned anyway at the repo owner's instruction — so flagging plainly that **this PR now carries a behaviour change beyond what you reviewed**: a `bool` identifier is rejected for all eight subtypes rather than silently converted.

Your diagnosis was exactly right, including the mechanism. The guard is placed **before** the subtype dispatch, which is the part that matters here:

```
7711:        if isinstance(identifier, bool):
7733:        if subtype_val == Enum_MNIDSubtype.IPv6_Address:
```

That ordering is not incidental. A guard placed *after* the `IPv6_Address` branch cannot catch that subtype — `True` would still become `IPv6Address(1)`, that is `::1`, precisely the surprise you described. The first attempt at this fix made that mistake, as did #468's negative-int guard before it, so the line numbers are the check rather than the passing test.

Measured at `35834b69b`, all eight subtypes against both values, on this tree (asserted via `pcapkit.__file__` before import):

```
combinations: 16 | ProtocolError: 16 | other: NONE
sample: MH: [OptNo 8] MN-ID identifier must not be a bool, not True -- pass int(True) if the numeric value is what is wanted
```

And confirming it is not collateral damage to the `int` half that shipped with #468:

```
IPv6_Address identifier=0x1234 -> IPv6Address('::1234')
IMSI         identifier=0x1234 -> b'\x124'
IMSI         int(True)         -> b'\x01'
```

So real `int` identifiers convert unchanged, and `int(True)` remains available as the way to pass the numeric value — which is what the error message points the caller at. A test sweeping eight subtypes against `{True, False}` covers it; reverting the guard fails exactly sixteen subtests.

Resolving.

pcapkit/protocols/internet/mh.py
```diff
@@ -7664,10 +7664,16 @@ def _make_opt_mn_id(self, type: 'Enum_Option', option: 'Optional[Data_MNIDOption
...
                 above with the ``IPv6_Address`` subtype (whose wire form is a
                 fixed 16 octets, unlike the other subtypes, which have no
-                ceiling and simply pack into more).
+                ceiling and simply pack into more), or ``identifier`` is of a
+                type its subtype's field cannot hold at all: anything but
+                :obj:`str` for ``NAI``, anything but :obj:`bytes`/
+                :obj:`bytearray`/:obj:`int` for the other six -- an :obj:`int`
+                is converted rather than rejected there, per #468 -- or anything
+                :class:`ipaddress.IPv6Address` itself does not accept for
+                ``IPv6_Address`` (c.f. #469).
```
JarryShaw, line 7684: This commit's own message says two nits were fixed on owner request: the ``Raises:`` bullet's "int" omission, and adding the ``bool`` guard itself. The first is fixed here (lines 7672-7674 now say ``bytes``/``bytearray``/``int`` and point at #468). The second lands a real, well-placed, well-tested guard (`isinstance(identifier, bool)` at line 7711, strictly before the subtype dispatch at line 7733 -- verified `bool(True)` is correctly rejected for every one of the 8 subtypes including `IPv6_Address`, where a post-dispatch guard would have missed it and produced `::1`).

But the docstring was not updated to describe *that* new behaviour. Neither ``Args:`` (line ~7649: "An ``int`` is accepted for every subtype except ``NAI``") nor ``Raises:`` (lines 7664-7676) says anything about `bool` now being rejected for every subtype. A reader of this docstring who calls `_make_opt_mn_id(..., identifier=True)` learns nothing from the docs about why it raises, or that `int(flag)` is the escape hatch -- both of which only appear in the runtime error message and the inline `NOTE:` comment.

Suggest adding a clause to ``Raises:``, e.g. "or `identifier` is a `bool` (accepted nowhere despite `bool` being an `int` subclass -- pass `int(flag)` for the numeric value)."

This is a docs-only nit, same class as the one this commit already fixed elsewhere in the same docstring -- not a functional issue. Code, guard placement, and test coverage (all 8 subtypes x {True, False}, verified independently including via the `option=` data-model path, which the new test does not itself exercise but which shares the same code path) all check out.

JarryShaw: Confirmed and fixed at `e41d3b720`. Good catch — and it is slightly worse than you filed it, which is worth stating.

You described it as an omission. It is also a **false statement**: the `Args:` sentence read

```
identifier: Identifier. An :obj:`int` is accepted for every
    subtype except ``NAI``.
```

and `bool` is an `int` subclass, so `True` *is* an `int` for every purpose that sentence is making a claim about — and it is now rejected. So the docstring did not merely fail to mention the new behaviour, it asserted the opposite of it. Verified before fixing: the only occurrence of ``bool`` anywhere in that docstring at `35834b69b` was the ``subtype_reversed: 'bool'`` parameter annotation, nothing about the identifier.

`Args:` now carries the carve-out and points at `Raises:` for the reason; `Raises:` leads with it and says *why* rejection is the right answer rather than only that it happens:

```
        Raises:
            ProtocolError: If ``identifier`` is a :obj:`bool`, for any subtype
                -- :obj:`bool` is an :obj:`int` subclass, so ``True`` would
                otherwise be converted by two different paths below, to
                ``::1`` for ``IPv6_Address`` and to a one-octet identifier for
                the other six, neither of which a caller passing a flag can
                plausibly have meant; pass ``int(...)`` to get the numeric
                value (c.f. #469). If ``identifier`` is a negative :obj:`int`
```

You noted this is the same class of defect the commit itself had just fixed two lines below. That is exactly why it was worth fixing rather than filing — a `Raises:` clause not matching behaviour is the thing #468 already had to correct once.

Docstring only, no behaviour change, and measured rather than assumed:

```
bool rejected:            16 / 16 subtype x {True, False}
IMSI 0x1234           ->  b'\x124'          (int half untouched)
test_mh_unit.py       ->  39 passed, 424 subtests passed
mypy --config-file mypy.ini on mh.py -> the same 2 pre-existing errors
                                        as before the change (:7604, :7995/:7987,
                                        shifted only by the 8 added lines)
```

The `git diff` is 11 insertions / 3 deletions, all inside the docstring.

Resolving.

PR #484, opened 2026-09-18
toolkit(scapy): scale IPv4 fragment offset to octets before reassembly

### LINKED ISSUE #483: scapy toolkit adapter passes raw 13-bit fragment offset unscaled to IP reassembly
### What

Found while fixing #477. `pcapkit/toolkit/scapy.py:160`:

\`\`\`python
data = IP_Packet(
    ...
    fo=ipv4.frag,                              # fragment offset
    ...
)
\`\`\`

\`ipv4.frag\` is scapy's raw IPv4 fragment-offset field — a 13-bit value counted in 8-octet units, per the wire format — not a byte offset. Every other pcapkit toolkit adapter either multiplies by 8 explicitly or uses a value the underlying engine has already scaled:

- \`pcapkit/toolkit/dpkt.py:215\`: \`fo=ipv4.offset * 8\`
- \`pcapkit/toolkit/pypcapfile.py:326\`: \`fo=ipv4.off * 8\`
- \`pcapkit/toolkit/pcap.py:67\` / \`pcapkit/toolkit/pcapng.py:70\`: \`fo=ipv4_info.offset\` (pcapkit's own \`ipv4.py\` sets \`offset=int(schema.flags['offset']) * 8\` at \`pcapkit/protocols/internet/ipv4.py:307\`, i.e. already in octets)
- \`pcapkit/toolkit/scapy.py:225\` (IPv6 fragment header, same file): \`fo=ipv6_frag.offset * 8\` — scapy's IPv6 fragment path *does* scale it; only the IPv4 path at line 160 does not

### Reproduction (not yet executed against the reassembler, confirmed only at the field level)

\`\`\`python
>>> import scapy.all as sc
>>> p = sc.IP(frag=5, flags='MF')
>>> bytes(p)[6:8].hex()
'2005'
\`\`\`

\`0x2005\` decodes as flags=\`001\` (MF set) + a 13-bit offset of \`5\` — i.e. scapy stored the raw unit count \`5\`, not a byte offset of \`40\`. \`pcapkit.toolkit.scapy.ipv4_reassembly\` would hand this packet to \`IP.reassembly()\` as \`fo=5\`, when every other engine reading the same bytes would report \`fo=40\`.

### Expected impact

Any capture read through the scapy engine (\`pcapkit.foundation.engines.scapy\`) containing IPv4 fragments would compute fragment placement 8x too small, corrupting reassembly for that engine specifically — silently, since nothing validates \`fo\` against the fragment's own length. IPv6 fragments read through the same engine are unaffected (the Fragment-header path already multiplies by 8).

### Scope

Not fixing here — this is a single-line, single-adapter fix (\`fo=ipv4.frag * 8\`) but wants its own verification against an actual reassembly run through the scapy engine (e.g. a scapy-crafted fragmented IPv4 packet fed through \`pcapkit.foundation.engines.scapy\`) and its own test, rather than being folded into #477/#482, which own \`pcapkit/foundation/reassembly/ip.py\` and its data model, not \`pcapkit/toolkit/scapy.py\`.

JarryShaw:
Closes #483

## What

`pcapkit/toolkit/scapy.py`'s `ipv4_reassembly` passed Scapy's raw 13-bit
`ipv4.frag` field straight through as `fo`, the byte offset the reassembler
indexes its buffer with. Scapy's `frag` is in on-wire 8-octet units per
:rfc:`791#section-3.1`, not bytes — so `fo` came out 8x too small for every
fragment except the first.

Every other adapter already gets this right:

- `dpkt.py`: `fo=ipv4.offset * 8`
- `pypcapfile.py`: `fo=ipv4.off * 8`
- `pcap.py` / `pcapng.py`: `fo=ipv4_info.offset`, where pcapkit's own
  `ipv4.py` already scaled it (`offset=int(schema.flags['offset']) * 8`)
- this same file's IPv6 path, a few lines below: `fo=ipv6_frag.offset * 8`

Only the scapy IPv4 path was missing the `* 8`.

## Reproduction (before the fix)

At the field level:

```pycon
>>> import scapy.all as sc
>>> bytes(sc.IP(frag=5, flags='MF'))[6:8].hex()
'2005'
```

`0x2005` decodes as flags `001` (MF) + a 13-bit offset of `5` — the raw unit
count, not a byte offset of 40.

End to end, through the public API (`pcapkit.toolkit.scapy.ipv4_reassembly`
into `pcapkit.foundation.reassembly.ipv4.IPv4`): feeding a `frag=0`/MF
fragment carrying 40 octets of `A`, then a `frag=5` final fragment carrying
8 octets of `B`, reports `fo=5` for the second fragment (expected byte
offset 40). The reassembler writes it inside the first fragment's span
instead of after it, and returns a **completed** datagram truncated to 13
octets: `b'AAAAABBBBBBBB'`, instead of the full 48 octets
(`b'A'*40 + b'B'*8`).

After the fix (`fo=ipv4.frag * 8`), the second fragment reports `fo=40` and
the datagram reassembles to the full 48 octets.

## The fix

One line in `pcapkit/toolkit/scapy.py`'s `ipv4_reassembly`:

```python
fo=ipv4.frag * 8,                          # fragment offset
```

with a comment recording the 8-octet-unit fact, mirroring the comment
already present on the IPv6 path in the same file.

## Tests

`tests/toolkit/test_scapy_unit.py`:

- Extended the existing `test_ipv4_and_ipv6_reassembly` with an `fo`
  assertion at the field level (mirroring the IPv6 assertion already there).
- Added `test_ipv4_reassembly_scales_fragment_offset_through_the_reassembler`,
  which drives the actual reassembler with two crafted fragments and asserts
  on the reassembled payload — not just on `fo` in isolation — since a
  fragment-offset-only check can't distinguish a real fix from one that
  merely relocates the same corruption elsewhere.

Both fail without the `* 8` scaling (`2 != 16` and `5 != 40` respectively)
and pass with it.

## Scope

Does not touch `pcapkit/foundation/reassembly/ip.py`, its data model, or
`tests/foundation/reassembly/test_ip.py` — those belong to #482. Does not
change any other toolkit adapter; none showed the same defect.

## Test plan

- [x] `pytest tests/toolkit/test_scapy_unit.py -v` — 5 passed (confirmed the
      two touched assertions fail without the fix, pass with it)
- [x] `pytest tests/toolkit tests/foundation/reassembly tests/foundation/engines tests/interface -q` — 230 passed, 15 skipped
- [x] `pytest tests -q` (full suite) — 1035 passed, 17 skipped
- [x] `mypy --config-file mypy.ini pcapkit/toolkit/scapy.py` — same 6
      pre-existing errors as on `origin/main`, none introduced by this change
- [x] scapy 2.7.0 is importable in the repo venv; scapy-dependent tests ran
      (not skipped) in this environment


tests/toolkit/test_scapy_unit.py
```diff
@@ -165,6 +170,58 @@ def test_ipv4_and_ipv6_reassembly(self) -> None:
...
+        datagram, = reasm.datagram
+        # only ``Completion.COMPLETE`` is truthy
+        self.assertTrue(datagram.completed)
+        # under the defect this comes back truncated to 13 octets
+        # (``b'AAAAABBBBBBBB'``): fragment 2 overwrote bytes 5-12 of
+        # fragment 1's span instead of being appended at byte 40, and the
+        # datagram's declared total length is computed from the corrupted
+        # (unscaled) offset of the final fragment
+        self.assertEqual(len(datagram.payload), 48)
+        self.assertEqual(bytes(datagram.payload), b'A' * 40 + b'B' * 8)
```
JarryShaw, line 223: Non-blocking suggestion: this test proves the payload-corruption half of #483's defect (the wrong bytes at the wrong offset), but it doesn't check the other symptom that the same corruption would have produced through #482's new ``conflict`` machinery.

I traced ``IP.reassembly()`` / ``IP._detect_conflicts()`` in ``pcapkit/foundation/reassembly/ip.py`` (unchanged by this PR) against this exact fixture under the pre-fix, unscaled ``fo``:

- Fragment 1 (``fo=0``, 40-octet payload) sets ``RCVBT`` bits for blocks 0-4 (``FO // 8`` through ``FO // 8 + (TL - IHL + 7) // 8`` = 0..5) and writes ``datagram[0:40] = b'A' * 40``.
- Fragment 2, with an unscaled ``fo=5`` instead of the correct ``40``, writes at ``datagram[5:13]``. ``_detect_conflicts`` runs *before* that write, over ``[5, 13)``: every one of those positions falls in an already-``RCVBT``-set block (0 and 1, both fully claimed by fragment 1), and ``tdl`` is still ``-1`` at that point (fragment 2's own ``MF=0`` update happens after), so the ``tdl < 0`` branch of the guard is satisfied everywhere. ``datagram[pos]`` is ``'A'`` and ``payload[index]`` is ``'B'`` for the whole span, so this reports a single conflicting run, i.e. ``conflict == ((5, 12),)`` -- a spurious conflict manufactured entirely by the unscaled offset, on data that never actually overlapped on the wire.

So a fix that scaled ``fo`` correctly but left some other consumer of the raw value unscaled would still be caught by the existing payload-bytes assertion, but a regression that reintroduced the offset bug would show up two ways: corrupted payload bytes *and* a non-empty ``conflict`` tuple on otherwise non-overlapping fragments. Since this is precisely the interaction #482 and #483 have with each other, asserting ``self.assertEqual(datagram.conflict, ())`` alongside the existing payload assertion would pin that down directly rather than leaving it to be inferred from the payload check.

Not asking for a change before merge -- the existing assertions already conclusively demonstrate the fix works -- just flagging the gap since it's exactly the interaction worth covering here.

JarryShaw:
## Review summary

Reviewed at `67dad967e51a811377536ffb5d242184af6831d1` (the tip of this PR's own commit, and the merge-base with the branch's current head `5e78093e0482ab73bcd505f1c2b0e5e0e7a6811f` -- the branch has since had `main` merged into it, but `pcapkit/toolkit/scapy.py` and `tests/toolkit/test_scapy_unit.py` are byte-identical between the two shas, so this review covers the PR's actual content either way).

### What changed
`pcapkit/toolkit/scapy.py`'s `ipv4_reassembly` now passes `fo=ipv4.frag * 8` instead of the raw `ipv4.frag`, with a `NOTE:` comment, plus two new/extended tests in `tests/toolkit/test_scapy_unit.py`.

### Checks performed

1. **Only one consumer of `ipv4.frag` in this module.** `grep -n frag pcapkit/toolkit/scapy.py` shows exactly one use of `ipv4.frag` (the fixed line) and the IPv6 path's own `ipv6_frag.offset * 8`, which already had the scaling. No sibling was left unscaled.

2. **Verified the reassembler's octet contract by reading `pcapkit/foundation/reassembly/ip.py`'s `reassembly()` directly** (unchanged by this PR -- identical between `fcd853032` and `67dad967e`): line 134-135 indexes the datagram buffer with `start = FO; stop = TL - IHL + FO` directly in octets, while lines 154-156 index `RCVBT` with `start = FO // 8; stop = FO // 8 + (TL - IHL + 7) // 8`. So the buffer wants octets and only `RCVBT` divides by 8, confirming the fix's contract is correct, not backwards.

3. **Interaction with #482's `conflict` feature:** `ip.py` (where `conflict` was added) is untouched by this PR. No existing test anywhere in `tests/` encodes the old buggy offset as expected behaviour (`grep -rln conflict tests/` finds only `test_ip.py`, `test_infoclass.py`, and `reassembly/data/test_models.py`, none of which reference scapy). The new regression test drives the real `IPv4()` reassembler and asserts on reassembled payload bytes (`b'A' * 40 + b'B' * 8`), not just on `fo` in isolation -- so it's a real regression test, not a tautology. I left one non-blocking inline suggestion: the test doesn't assert `datagram.conflict == ()`, and by tracing `_detect_conflicts` against this exact fixture I could show the pre-fix unscaled offset would have produced a spurious `conflict == ((5, 12),)` on data that never actually overlapped on the wire -- worth asserting directly given it's the exact #482/#483 interaction, but not required since the payload-bytes assertion already conclusively pins the fix.

4. **Test adequacy.** `test_ipv4_and_ipv6_reassembly` now asserts both the raw `ipv4.frag == 2` and the scaled `reassembled.fo == 16`, which shows the multiplication actually happened rather than restating a single number. The new `test_ipv4_reassembly_scales_fragment_offset_through_the_reassembler` exercises a real two-fragment reassembly and checks payload bytes. Not covered, and worth having eventually but not blocking here since the fix is a single scalar multiply independent of fragment count/order: a three-fragment datagram, out-of-order arrival, and a final fragment whose length isn't itself a multiple of 8 (a genuinely partial `RCVBT` tail block).

5. **scapy actually ran.** Confirmed scapy 2.7.0 importable in this environment and `pcapkit.__file__` resolving to my worktree before running anything. `tests/toolkit/test_scapy_unit.py` ran for real (no skips):
   ```
   PYTHONPATH=<worktree> .venv/bin/python -m pytest tests/toolkit/test_scapy_unit.py -v
   5 passed in 4.12s
   ```
   Broader selection (`tests/toolkit tests/foundation/reassembly tests/foundation/engines tests/integration`):
   ```
   305 passed, 17 skipped, 232 subtests passed in 250.19s
   ```
   No failures in either run.

6. **The comment and RFC citation.** The comment's `:rfc:` role points at RFC 791 section 3.1, which is the IPv4 header format section that defines the Fragment Offset field in 8-octet units -- accurate. The comment explains *why* (the reassembler indexes in octets while the wire field is in 8-octet units, mirroring the IPv6 path) rather than just restating `* 8`.

### CI
```
gh api repos/JarryShaw/PyPCAPKit/commits/67dad967e/check-runs --paginate --jq ...
success=21 skipped=2 unfinished=0 failed=0
```
The 2 skips are `Gate (full suite, Python 3.14)` and `Docs test gate`, which skip by design on `pull_request` -- fully green otherwise. All `Python 3.10`-`3.15`, `Compat`, `Integration`, `CodeQL`/`Analyze`, and `deploy-pages` checks passed.

### Verdict

**GOOD TO MERGE** `67dad967e51a811377536ffb5d242184af6831d1`

One non-blocking inline suggestion posted above (strengthening the regression test with a `conflict == ()` assertion); nothing blocking found.

PR #504, opened 2026-09-19
test: enumerate the 38 __proto__ dispatch-registry entries

### LINKED ISSUE #496: No enumerating harness for the 38 __proto__ dispatch-registry entries (the OSPF regression's blind spot)
Found by strand 2 of the post-wave-1 consistency sweep — *named* test gaps rather than a coverage percentage. Counts below are mine, measured on `5959a437c`.

This is a test-coverage proposal, not a defect: every registry entry named here dispatches correctly today. The case for it is that a defect of exactly this kind has already shipped once, and the existing guard against it is hand-written rather than enumerated.

## The gap

`pcapkit/foundation/registry/protocols.py` defines the registries that decide **which `Protocol` class parses the next layer**. Measured sizes:

| registry | keyed by | entries |
|---|---|---|
| `Link.__proto__` | EtherType | 7 |
| `Internet.__proto__` | TransType | 16 |
| `TCP.__proto__` | port | 4 |
| `UDP.__proto__` | port | 3 |
| `SCTP.__proto__` | PPID | 2 |
| `Frame.__proto__` | pcap LinkType | 3 |
| `PCAPNG.__proto__` | LinkType | 3 |

**38 entries, and no test enumerates them.** To be precise about the claim, since several test files do reference `__proto__`: what is missing is a harness that walks each registry and asserts that dispatching *through* each registered code actually parses a packet into the registered class. `tests/protocols/test_registry_runtime.py` and similar check that an entry *resolves to the right class object*, which is a weaker property — it is satisfied by a table whose target could not parse a packet if it tried.

`tests/protocols/test_dispatch_bindings_unit.py` exists precisely because this class of bug shipped. Its own docstring records that OSPF *"was reachable from no table at all and could not have parsed a packet if it had been"*. But it hand-picks its cases — 11 test functions covering Q-in-Q S/C-Tag, OSPF, L2TP, and the FTP/HTTP ports — rather than enumerating, so it guards the entries someone remembered and not the registry.

The contrast worth drawing: the **option/parameter** registries already have exactly the harness being asked for. `tests/protocols/test_option_roundtrip_unit.py` with `examples/generators/options.py` enumerates 18 families via `FAMILIES` and *fails when a registered code has no case*. The `__proto__` family never got the equivalent.

## The specific entries with no dispatch test

Ranked by what a slipped regression would cost. All were probed and **all dispatch correctly today** — these are unguarded, not broken:

1. **`Internet.__proto__` for AH (51), HIP (139), Mobility_Header (135), SCTP (132)** — four of the highest-traffic entries in the library, each with only an identity check (`AH.__index__() == TransType.AH`) and no parse. The option round-trip generator constructs `MH(...)`/`HIP(...)`/`SCTP(...)` **directly**, bypassing `Internet.__proto__`; and `test_option_coverage_runtime.py` asserts only the chain *prefix* (`'Ethernet:IPv6'`), so the continuation into MH/HIP/AH/SCTP is asserted nowhere. Probed: `Ethernet:IPv4:AH:TCP`, 46 frames of `Ethernet:IPv6:HIPv2` and 99 of `Ethernet:IPv6:MH` from the real fixture, plus a synthetic `Ethernet:IPv4:SCTP`.
2. **`Link.__proto__[0x8035]` (RARP)** — identity-level only. Probes as `Ethernet:RARP:Raw`. One trap for whoever writes the test: ARP/RARP/InARP/DRARP's reported name comes from the wire `oper` field (`link/arp.py:176-190`), not from the dispatching ethertype, so an ethertype-0x8035 frame with `oper=1` legitimately reports `Ethernet:ARP`. The test needs `oper` in `{3,4}` or it silently exercises the wrong alias.
3. **`Frame.__proto__[LinkType.IPV4]`/`[IPV6]` (228/229)** — every pcap test uses `ETHERNET` or the fallback-triggering `NULL`; nothing builds a raw-IP capture with no link layer. Probed directly: both parse (`IPv4:TCP:Raw`, `IPv6:TCP:Raw`).
4. **`PCAPNG.__proto__`'s matching two entries** — **not probed.** The sweep judged that hand-building a valid minimal SHB+IDB+EPB stream risked a false negative worth more than the check, and reported it as unverified rather than claiming a result. Recorded here the same way: same `ModuleDescriptor` targets and same dispatch path as the `Frame` entries above, but not independently confirmed.

## Cleared, so this ground is not re-audited

* `Internet.__proto__` for HOPOPT (0), IPv6_Opts (60), IPv6_Route (43) — genuinely dispatch-tested by the #425 regression test in `test_ipv6_extension_unit.py`, which builds real IPv6 datagrams per next-header.
* IPv6_Frag (44), ESP (50), IPv4 (4, IP-in-IP) — covered end to end (`'Ethernet:IPv6:IPv6-Frag:UDP:Raw'`; `'Ethernet:IPv4:ESP:IPv4:ICMP'`).
* **All** `TCP.__proto__` (20, 21, 80, 8080) and `UDP.__proto__` (80, 1701, 8080) ports — exercised by `test_dispatch_bindings_unit.py`. No gap.
* `SCTP.__proto__` — both PPIDs (60, 66) exercised in `test_sctp_unit.py::test_ppid_dispatch_hook`; PPID 66's Raw-only degrade is documented (DTLS-wrapped, no DTLS implementation), not a gap.
* `EXPECTED_FAILURES` is **not stale** — imported the module (it cannot be grepped or ast-parsed, as the dict uses `**` unpacking): 59 entries, with `ipv6-route-type/RPL_Source_Route_Header` the sole surviving ipv6-route one. `test_option_roundtrip_unit.py`: 6 passed / 358 subtests.

## Suggested shape

A `FAMILIES`-style enumerator over the seven `__proto__` tables, mirroring `examples/generators/options.py`: build one minimal envelope per registered code, run it through `pcapkit.extract()` (or the frame class directly for the pcap-level tables), and assert the protochain reaches the registered class. That converts 11 hand-picked cases into 38 enumerated ones and would have caught the OSPF regression that motivated the existing file.

## Not covered

A full sweep of all 907 `raise` sites against `assertRaises` coverage was out of scope. "Behaviour asserted only in prose" belongs to strand 1 and was not separately passed over here. `const/http/*`, ESP's cipher/integrity enums, ARP hardware/operation, OSPF authentication and VLAN priority are open field-value vocabularies rather than code→dissector registries, so they do not fit this harness shape and were not audited. The full `tests/` suite was not run as a baseline.

JarryShaw:
Closes #496. Test-only: `examples/generators/dispatch.py` and `tests/protocols/test_dispatch_registry_unit.py`, 1,199 insertions, **no library code touched**.

## The gap

`pcapkit/foundation/registry/protocols.py` defines the registries that decide which `Protocol` class parses the next layer. **38 entries across seven tables** — `Link.__proto__` (7), `Internet.__proto__` (16), `TCP.__proto__` (4), `UDP.__proto__` (3), `SCTP.__proto__` (2), `Frame.__proto__` (3), `PCAPNG.__proto__` (3) — and none of them had an enumerating harness.

Existing coverage checked that an entry *resolves to the right class object*, which is satisfied by a table whose target could not parse a packet if it tried. `tests/protocols/test_dispatch_bindings_unit.py` exists because this class of bug shipped — its docstring records that OSPF *"was reachable from no table at all and could not have parsed a packet if it had been"* — but it hand-picks 11 cases rather than enumerating.

This is the `__proto__` equivalent of what `tests/protocols/test_option_roundtrip_unit.py` and `examples/generators/options.py` already give the option registries, and it follows their structure.

## Two things found while building it, both worth reading

**1. The first draft's expectation was self-referential.** `cases()` built each `Case` with `target=_resolve(entry)` straight from `registry.items()`, so the expected target was read out of the very table under test. It proved "dispatch honours whatever the table says", not "the table says the right thing". I breakage-tested retargeting `AH` from `ah.AH` to `raw.Raw` with the registry size unchanged and got **5 passed, 45 subtests** — not caught.

Fixed with `PINNED_TARGETS`, a hand-written `label -> (module, name)` table independent of the registry. `Case` now carries `target` (pinned) and `registered` (whatever the live table resolves to) as separate fields, and a mismatch between them is itself a failure.

**2. Even a pinned target was not enough, for a genuinely subtle reason.** `case.target in frame.protochain` *still* missed the retarget, because `ProtoChain.__contains__`/`index()` match on alias strings and `.id()` names rather than class identity — and `Raw` adopts the dispatching protocol's own name (`pcapkit/protocols/misc/raw.py`: `alias = self._info.protocol.name`). So an entry keyed on `TransType.AH` but retargeted to `Raw` still renders as `'...:AH'` and satisfies a string check, with no `AH` instance anywhere in the chain.

Fixed by asserting strict class identity against `frame.protochain.protocols` — the tuple of actual `type(instance)` per layer. Worth knowing beyond this PR: **a protochain string containing a protocol's name is not evidence that protocol parsed anything.**

## Both breakage shapes, run by me against the final code

**Deletion** — `AH` line removed from `Internet.__proto__`:
```
FAILED ...::test_cases_cover_every_table_named_in_the_issue
AssertionError: 37 != 38 : expected exactly 38 entries across the seven __proto__ tables
1 failed, 6 passed, 81 subtests passed
```

**Retargeting** — `AH` → `Raw`, registry size unchanged:
```
SUBFAILED(case='internet/AH') ...::test_dispatch_reaches_target_or_is_a_recorded_degrade
SUBFAILED(case='internet/AH') ...::test_registry_currently_matches_pinned_target
AssertionError: <class 'pcapkit.protocols.misc.raw.Raw'> is not <class 'pcapkit.protocols.internet.ah.AH'> :
  internet/AH: the registry currently resolves to Raw, but PINNED_TARGETS expects AH. Either the
  registry regressed, or PINNED_TARGETS is stale and needs updating to match a deliberate change.
2 failed, 7 passed, 81 subtests passed
```

Caught by **two independent tests**, both naming `internet/AH`, with a message that distinguishes the two possible causes. Restored after each: **7 passed, 83 subtests passed**, clean tree.

## Coverage

All **38 of 38** entries, including the ones that previously had only an identity check and no parse test: AH (51), HIP (139), Mobility_Header (135), SCTP (132), RARP (0x8035), and the raw-IP `Frame.__proto__` pair (228/229). The `PCAPNG.__proto__` IPV4/IPV6 pair — which the audit flagged as unverified because hand-rolling a minimal SHB+IDB+EPB stream looked riskier than the check was worth — is now confirmed working, built through `pcapkit`'s own `PCAPNG`/`Context` construction API rather than by hand.

34 of 38 reach their target cleanly. **4 are recorded in `KNOWN_DEGRADED` rather than fixed** — this module reports defects, it does not carry workarounds:

* `link/Novell_Inc_0x8137` and `internet/IPX_in_IP` — both hit `0 is not a valid Socket` (#492, fixed in #503; once that merges these entries should be deleted and the cases should pass).
* Both SCTP PPIDs (60, 66) — reaching NGAP needs a well-formed PER-encoded `NGAP-PDU`, out of scope for a minimal probe. Same placeholder and same documented reason as the existing `test_sctp_unit.py::test_ppid_dispatch_hook`.

`KNOWN_DEGRADED` is a ratchet, not a skip-list: a case in it must degrade in the **recorded** way, the test tells you to delete the entry once the defect is fixed, a separate test rejects entries naming cases that no longer exist, and another keeps degraded cases a small minority.

## A trap for whoever extends this

ARP/RARP/InARP/DRARP's reported name comes from the wire `oper` field (`pcapkit/protocols/link/arp.py:176-190`), **not** from the dispatching ethertype, so an ethertype-0x8035 frame with `oper=1` legitimately reports `Ethernet:ARP`. The RARP case uses `oper` in `{3,4}` for that reason; anything else exercises the wrong alias while appearing to pass.

## Verification

`test_dispatch_registry_unit.py`: **7 tests, 83 subtests, 0 failures**. Full `tests/protocols/` blast radius: **470 passed, 1341 subtests, 0 failures**.

Three-dot `origin/main...HEAD` confirms only the two files. (The two-dot form additionally shows `pcapkit/__init__.py` — that is `main`'s own `1.5.0b3` bump appearing as a reversal, not a change here; `git merge-tree` against `main` produces a tree whose `__version__` is `1.5.0b3`.)

This module constructs its own octets and reads nothing under `examples/captures/`, so it belongs to the unit tier and runs on a fresh checkout with nothing generated.


JarryShaw:
## Diagnosis of the `Python 3.15` failure — it was never version-specific

The failing job's log (reachable via `gh api .../actions/jobs/<id>/logs --allow-escape-sequences`, which works while the run-level log is still gated) named it precisely:

```
SUBFAILED(case='sctp/PayloadProtocolIdentifier_3GPP_NG_Application_Protocol')
AssertionError: 'malformed NGAP-PDU' not found in
  'NGAP: decoding needs the optional "pycrate" dependency, which is not installed;
   pip install pypcapkit[NGAP]'
  : still degrades, but not in the recorded way
```

Both SCTP NGAP entries in `KNOWN_DEGRADED` pinned their degradation reason to the string `'malformed NGAP-PDU'`. **That reason is environment-dependent.** Where `pycrate` is installed the PER decoder runs and reports a malformed PDU; where it is not, NGAP reports the missing dependency instead. Both are the same placeholder-payload limitation, not two different defects — and the entry accepted only the first.

`.github/workflows/unit-tests.yml:82` installs `.[test]`, which **deliberately excludes `[NGAP]`** (`pyproject.toml:114-122` keeps `pycrate` out of `all` on size and licence grounds). So **no** CI leg has it. `Python 3.15` merely reported first; 3.10–3.14 were still queued and would have failed identically.

The generator's own comment asserted the opposite — that the failure occurs "the same way it would on any checkout regardless of whether the optional `pycrate` dependency is installed". That was factually wrong and is corrected, since it is the assumption that produced the failure.

## The fix

`Degrade.fragment` now accepts a tuple, meaning *any* of its substrings is an acceptable reason, and both NGAP entries record both reasons. The assertion's message tells a future reader to **add** a reason rather than widen the entry to match anything, so this cannot decay into a blanket pass.

No library code touched; still test-only.

## Verification — reproducing CI's condition locally

`pycrate` installs as `pycrate_asn1dir`/`pycrate_asn1rt`/`pycrate_core`; there is no top-level `pycrate` module, so a bare `import pycrate` is not a valid presence check. I blocked the real module names with a `sys.meta_path` finder and confirmed NGAP then emits CI's exact string:

```
NGAP: decoding needs the optional "pycrate" dependency, which is not installed; pip install pypcapkit[NGAP]
```

| condition | before the fix | after |
|---|---|---|
| `pycrate` blocked (CI's condition) | **2 failed**, 7 passed, 81 subtests | **7 passed, 83 subtests** |
| `pycrate` present (local) | 7 passed, 83 subtests | **7 passed, 83 subtests** |

Both `SUBFAILED` lines in the "before" run name the two SCTP cases, matching CI's failure exactly.

The full unit tier — the same command CI runs, `pytest -q --ignore=tests/integration --ignore-glob='*_runtime.py' --ignore-glob='*_regression.py'` — passes in this branch's worktree on 3.14: **924 passed, 5 skipped, 1867 subtests, 0 failed**.

## Note on what this was not

This is the harness having pinned an environment-dependent detail too narrowly. It is **not** a library defect, and the two cases still degrade exactly as recorded — they simply have two legitimate ways of saying so.

JarryShaw:
## Pushed `52bc0603c` — the two IPX `KNOWN_DEGRADED` entries are gone, because #503 fixed them

Merging `main` into this branch brought in #503, which added the missing `Socket(0x0000)` member. Both IPX entries recorded those cases as *degrading*, and this harness asserts a recorded degrade **still degrades** — so leaving them in fails, by design:

```
AssertionError: True != False : internet/IPX_in_IP was recorded as degrading
  (same defect as link/Novell_Inc_0x8137 -- both dispatch to
   pcapkit.protocols.internet.ipx.IPX (#492))
  but now reaches its target ('Ethernet:IPv4:IPX:Unknown').
  If the defect is fixed, delete its KNOWN_DEGRADED entry.
```

That is the ratchet doing its job rather than a problem: the entry could not silently rot into a permanent exemption. Removed from both the test copy and the generator copy.

**Verified the cases genuinely reach IPX rather than merely stopping to fail** — worth distinguishing, since "no longer errors" and "parses correctly" are different claims:

```
link/Novell_Inc_0x8137: reached=True chain='Ethernet:IPX:Unknown'
internet/IPX_in_IP:     reached=True chain='Ethernet:IPv4:IPX:Unknown'
```

So this PR now also serves as an independent check on #503: IPX dispatch works end to end through a harness that knew nothing about that fix.

Tests, in both environments: **7 passed, 83 subtests** with `pycrate` present, and **7 passed, 83 subtests** with it blocked — the latter reproducing CI's condition, since `.[test]` excludes `[NGAP]`.

`KNOWN_DEGRADED` now holds only the two SCTP NGAP cases, which are a placeholder-payload limitation of the probe rather than a library defect.

Pushed as a follow-up commit rather than an amend, since the merge commit `81893a7e6` sits in between and amending would have rewritten it.

PR #509, opened 2026-09-19
refactor: extract duplicated wire-unit arithmetic across mh/ipv6_route/hopopt/ipv6_opts

### LINKED ISSUE #495: Duplicated wire-unit arithmetic wants a shared module: 25 MH sites, 126 option sites, and two near-identical option files
Found by strand 5 of the post-wave-1 consistency sweep — the strand the owner added at `c54727865`: *check if there're duplicated common methods/logic that can be extracted into a shared module*. Measurements below are mine, taken on `5959a437c`.

Unlike the other sweep issues, **nothing here is currently producing a wrong answer.** Every site is consistent today. The case for acting is that this exact duplication has already caused defects twice — #487 and #398 — and the remaining copies are the same pattern awaiting the same fate. Filed as structural work, not as a defect.

## The evidence that this is not aesthetic: #398 fixed one bug six times

`29dfd351b` — *"ipv6/mh: fix option padding, and with it construction, which was wholly broken (#398)"*:

```
pcapkit/protocols/internet/hopopt.py           | 130 ++++++++---
pcapkit/protocols/internet/ipv6_opts.py        | 131 ++++++++---
pcapkit/protocols/internet/mh.py              | 130 ++++++++++-
pcapkit/protocols/schema/internet/hopopt.py    |  46 +++-
pcapkit/protocols/schema/internet/ipv6_opts.py |  46 +++-
pcapkit/protocols/schema/internet/mh.py        |  47 +++-
```

One conceptual bug — a `+2`/`-2` mismatch between the whole-option length and the on-wire Option Length field — diagnosed once and then patched independently in six places with near-identical diffs. The surviving explanation, `mh.py:7446-7453`:

> `Data_PadOption.length` counts the *whole* option, whereas `Schema_PadOption.length` is the `Option Length` field — two octets fewer... Copying one into the other unconverted is why re-making a parsed `PadN` used to come back two octets too long.

#487 is the other precedent: `Hdr Ext Len` computed in four places, with four different wrong answers.

## 1. `hopopt.py` and `ipv6_opts.py` are the same implementation twice

The strongest case in the codebase, because **there is no specification reason for these two to differ**: Hop-by-Hop Options and Destination Options draw on the same IANA registry (RFC 8200 §4.3/§4.6), which is why every per-option method is the same in both files.

Measured:

```
hopopt.py    1953 lines
ipv6_opts.py 1965 lines
full diff:    406 changed lines out of ~1950
```

The overwhelming majority of those 406 lines is renaming — `HOPOPT`→`IPv6_Opts`, `hopopt`→`opt`, docstring rewrap — not different logic.

**Two corrections to how this reached me.** The sweep reported 666 changed lines; I measure **406**. It also reported that diffing the `def` lines "produces zero output"; it does not — `ipv6_opts.py` has an extra `alias` method, and the options entry point is named `_read_hopopt_options` versus `_read_ipv6_opts_options`. The method sets are *near*-identical, not identical. The conclusion holds; the specific claims were overstated.

A shared options base class, or a mixin the two thin header classes compose, would collapse this and would have made #398 a one-place fix.

## 2. MH message length: 25 identical read-side sites, while the write side was already unified

Every `_read_msg_*` in `mh.py` computes `length=(header.length + 1) * 8`. Confirmed count: **25** occurrences, all textually identical — no drift yet.

That value feeds `self._decode_next_layer(mh, schema.next, length - mh.length)` at `mh.py:1369`, which is precisely the role `Hdr Ext Len` played in #487.

The asymmetry is the point: the **write** side already got the treatment, unified into one expression at `mh.py:1438` (`length=(len(data_val) + 6) // 8 - 1`), while the read side kept 25 copies. One `_mh_message_length(hdr_len_field)` helper, called from all 25, mirrors what `IPv6_Route._make_hdr_ext_len`/`ipv6_route_data_length` already do.

## 3. `IPv6_Route` read side: the unfinished half of #487/#489

`ipv6_route.py:447, 514, 562, 615` — all four `_read_data_type_*` methods repeat `length=header.length * 8 + 8`. Confirmed count: **4**.

#489 unified the *write* side into `_make_hdr_ext_len`/`ipv6_route_data_length`. This read-side "total header octets" figure is a third related-but-distinct quantity — `8 + 8k` here versus `ipv6_route_data_length`'s `4 + 8k` — and is still computed four ways. Small blast radius, but immediately adjacent to work just merged, so cheap to finish now: an `ipv6_route_header_length()` beside the existing helper.

## 4. Option `+2` arithmetic: 126 sites across three files

Every `_read_opt_*` across `mh.py` (92 sites), `hopopt.py` (17) and `ipv6_opts.py` (17) computes `length = <schema>.length + 2`, adding back the Type+Length prefix that RFC 6275 §6.2 and RFC 8200 §4.3 exclude. This is the quantity #398 got wrong.

Only four make-side places consume a stored `.length` (`mh.py:7457`, `mh.py:9139`, `hopopt.py:1384`, `ipv6_opts.py:1396`) and all four correctly subtract 2 — four correct sites beside 126 duplicated ones with no shared helper between them.

**The honest severity, which the sweep stated well:** the option round-trip harness would *not* catch a fresh regression here, because `_make_opt_*` methods other than Pad recompute length from `len(value)` rather than reading `.length` back. Wire bytes stay correct; what drifts is the informational `.length` on a parsed packet. That is also *why* it has not bitten again — but nothing stops a new option's maker from following Pad's original pattern rather than its fixed one.

## Deliberately not proposed: the TCP/IP reassembly pair

Recorded so a future sweep does not "fix" it. `reassembly/tcp.py` and `reassembly/ip.py` look like near-duplicates and must **not** be merged: they resolve overlapping data in opposite directions because their specifications demand it. RFC 791 lines 1892-1894 — *"either identically or through a partial overlap, this procedure will use the more recently arrived copy in the data buffer and datagram delivered"* — is last-write-wins; RFC 9293 §3.10 is first-write-wins. IP is last-write-wins, TCP is first-write-wins, by specification. Verified against the RFC text; #443 settled the TCP side deliberately.

## Retired: extraction not worth it, with reasons

* **`pcap.py`/`pcapng.py` don't scale fragment offset by 8 while `dpkt.py`/`scapy.py`/`pypcapfile.py` do.** Not duplication that drifted: the native engines consume pcapkit's own already-scaled `Data_IPv4.offset`, whereas the third-party engines expose the raw 8-octet-unit field and must scale it themselves. The same relationship, correctly expressed differently because the inputs' units differ. (Note this is adjacent to the real defect filed separately for `IPv4._make_data`, which *is* a missing conversion.)
* **`pypcap.py` / `pcap_ct.py`** raise `UnsupportedCall` throughout — no dissection capability, nothing to extract.
* **HIP's `length=schema.len * 8 + 8`** (`hip.py:545`) — a single call site, since HIP has one top-level packet format. No extraction candidate.

## Not covered

`pcapkit/protocols/link/**` and `application/**` were not swept for the same arithmetic patterns, and `tcp.py`, `hip.py`'s parameter handlers and `arp.py` were not read method-by-method — targeted greps surfaced nothing there, which is weaker than having checked.

JarryShaw:
Closes #495 — the last open item from the post-wave-1 consistency sweep. Pure refactor: **168 duplicated arithmetic sites replaced by five helpers, with no behaviour change.**

## Why this is not tidying

Two shipped defects came from exactly this duplication. **#487** computed `Hdr Ext Len` in four places and got four different wrong answers. **#398** (`29dfd351b`) was one conceptual `+2`/`-2` unit mismatch that had to be diagnosed once and then patched independently in **six** files — `hopopt.py` 130 lines, `ipv6_opts.py` 131, `mh.py` 130, plus all three schema files at 46/46/47.

## What changed

| scope | sites | helper |
|---|---|---|
| `ipv6_route.py` total header octets | **4** | `ipv6_route_header_length()`, beside the existing `ipv6_route_data_length()` |
| `mh.py` message length | **25** | `MH._mh_message_length()` |
| `mh.py` option length | **92** | `MH._mh_option_length()` |
| `hopopt.py` option length | **17** | `HOPOPT._hopopt_option_length()` |
| `ipv6_opts.py` option length | **17** | `IPv6_Opts._ipv6_opts_option_length()` |

The `mh.py` message-length case is the clearest: the **write** side was already unified at `mh.py:1438` while the read side stayed duplicated 25 ways, and that value feeds `_decode_next_layer(mh, schema.next, length - mh.length)` — the same role `Hdr Ext Len` played in #487.

`ipv6_route_header_length()` is defined as `4 + ipv6_route_data_length(hdr_ext_len)`, which keeps the two related-but-distinct quantities visibly related: `8 + 8k` total header octets versus `4 + 8k` data octets. Confusing those two *was* #487.

## Deliberately NOT done

**The `hopopt.py`/`ipv6_opts.py` shared base class.** Those two files are one implementation twice (406 changed lines out of ~1950, overwhelmingly renaming), and collapsing them would subsume a large part of this issue — but it is a cross-cutting change that deserves its own issue and its own review, so the option-length helpers here are **per-class** rather than one shared helper across the two files. That boundary was drawn on purpose.

Also untouched, and already retired with reasons in #495: `pcap.py`/`pcapng.py` not scaling fragment offsets (they consume already-scaled values, unlike the third-party adapters, so the asymmetry is correct); `pypcap.py`/`pcap_ct.py` raising `UnsupportedCall` throughout; HIP's single `length=schema.len * 8 + 8` call site, which has nothing to share with.

## The trap this change had to avoid

`reassembly/tcp.py` and `reassembly/ip.py` look like near-duplicates and **must not be merged**: RFC 791 lines 1892-1894 mandate last-write-wins for IP, RFC 9293 §3.10 mandates first-write-wins for TCP, and #443 settled that deliberately. Neither file is in this diff — **0 reassembly files**, verifiable from `git diff --name-only origin/main...HEAD`.

## Verification

**Helper arithmetic checked independently of the tests** — 240 comparisons over `k` = 0..39 against the original expressions, **0 mismatches**:

```
ipv6_route_header_length(k) == k*8 + 8      ipv6_route_data_length(k) == 4 + k*8
MH._mh_message_length(k)    == (k+1)*8      MH._mh_option_length(k)   == k + 2
HOPOPT._hopopt_option_length(k) == k + 2    IPv6_Opts._ipv6_opts_option_length(k) == k + 2
```

**Every old pattern is gone, and the counts match #495's measurements exactly:**

```
                            main   branch
mh  (header.length + 1) * 8   25      0
ipv6_route  length * 8 + 8     4      0
hopopt      .len + 2          17      0
ipv6_opts   .len + 2          17      0
```

**Byte-identity against `main`, which is the only proof that matters for a refactor.** Construct-then-parse across IPv6-Route source-route with 0/1/2/3 addresses, six MH message types, and the HOPOPT/IPv6-Opts Router-Alert option, dumped to JSON from both trees:

```
BEFORE (main):     md5 d797c44b0c1b
AFTER  (#495):     md5 d797c44b0c1b
diff: IDENTICAL
```

Critically, **0 error-ish lines in that output** — so this is a comparison of real results, not of two identical crashes. (A driver that dies the same way on both trees reports "identical" while proving nothing; that happened earlier in this programme and is worth guarding against explicitly.)

Tests on the directly affected paths: `test_option_roundtrip_unit.py` + `test_mh_unit.py` + `test_ipv6_extension_unit.py` → **98 passed, 860 subtests passed**. `EXPECTED_FAILURES` inspected by importing the module (it cannot be grepped — `**` unpacking): **59 entries, none needing an update**.

## One deliberate omission

No dedicated unit tests for the five helpers. That follows the house precedent set by #487/#489: `IPv6_Route._make_hdr_ext_len` and `ipv6_route_data_length` have no dedicated tests either and are exercised through the existing round-trip suite. Adding a new convention here rather than following the existing one seemed the wrong call for a refactor, but it is a judgement worth disagreeing with — the arithmetic check above is a script, not a committed test.


pcapkit/protocols/schema/internet/ipv6_route.py
```diff
@@ -62,6 +62,33 @@ def ipv6_route_data_length(hdr_ext_len: 'int') -> 'int':
     return 4 + hdr_ext_len * 8
 
 
+def ipv6_route_header_length(hdr_ext_len: 'int') -> 'int':
```
JarryShaw, line 65: Defined but not used anywhere?

JarryShaw: It is used — four call sites, in the four `_read_data_type_*` methods. The definition lives in the schema module beside its sibling `ipv6_route_data_length()`, and the protocol module imports it:

```
pcapkit/protocols/internet/ipv6_route.py:45   from ...schema.internet.ipv6_route import ipv6_route_header_length
pcapkit/protocols/internet/ipv6_route.py:448      length=ipv6_route_header_length(header.length),   # _read_data_type_none
pcapkit/protocols/internet/ipv6_route.py:515      length=ipv6_route_header_length(header.length),   # _read_data_type_src
pcapkit/protocols/internet/ipv6_route.py:563      length=ipv6_route_header_length(header.length),   # _read_data_type_2
pcapkit/protocols/internet/ipv6_route.py:617      length=ipv6_route_header_length(header.length),   # _read_data_type_rpl
```

Those four are exactly the `header.length * 8 + 8` sites #495 lists, and `grep -c 'header\.length \* 8 + 8'` on this branch returns **0** where `main` returns **4**.

It is placed in the schema module rather than on the class deliberately: `ipv6_route_data_length()` (added by #489) is already there, and the two are the related-but-distinct quantities that #487 got wrong — `8 + 8k` total header octets versus `4 + 8k` data octets. Defining the new one as `4 + ipv6_route_data_length(hdr_ext_len)` keeps that relationship visible in one place instead of leaving two similar-looking formulas to drift apart again.

**One thing your question did surface, though, and it is worth deciding rather than leaving:** neither helper is in that module's `__all__` —

```python
__all__ = [
    'IPv6_Route',
    'RoutingType',
    'UnknownType', 'SourceRoute', 'Type2', 'RPL',
]
```

so `ipv6_route_data_length` has been absent since #489 and `ipv6_route_header_length` follows the same pattern. That makes the new one *consistent* rather than newly inconsistent, which is why I left it alone — but if you would rather both were exported (and picked up by the schema `.rst`, which currently references neither), that is a one-line change and I am happy to add it here.

PR #511, opened 2026-09-19
fix(vendor): retire the IPX socket scrape, its data source is gone (#507)

### LINKED ISSUE #507: IPX vendor crawler is doubly broken: Wikipedia 403s the default User-Agent, and the socket-number table is gone
Found while fixing #492 (PR #503) and deliberately left out of it. **Both halves verified against the live page just now**, on `main` at `d2d9bb80c` — not inferred from the code.

## Two independent breakages in one crawler

`pcapkit/vendor/ipx/socket.py` scrapes `https://en.wikipedia.org/wiki/Internetwork_Packet_Exchange#Socket_number`.

**1. Wikipedia rejects `requests`' default User-Agent.**

```
default-UA GET -> HTTP 403
browser-UA GET -> HTTP 200, 113743 bytes
```

So the crawler cannot fetch the page at all as written. Wikipedia has been progressively blocking generic library User-Agents; this is not specific to this repo.

**2. Even with a working fetch, the table it wants is gone.** The page now carries exactly **one** `wikitable`, and it is the wrong one:

```
wikitable count on the live page: 1
  table 0: headers ['Octets', 'Field']
```

That is the IPX *header format* table. The well-known-socket-number table this crawler parsed has been removed from the article. So there is nothing left to scrape even after fixing the User-Agent — the data source itself is gone.

## What #503 did instead, and why that is not a fix

#503 needed `Socket(0x0000)` to exist, because it is IPX's own class default and a legitimate wire value (see #492). Rather than depend on a scrape that currently yields nothing, `process()` now prepends the member **unconditionally**, with a comment saying so.

That is deliberately a workaround. It gets the one needed member in regardless of the scrape, but it does not make the crawler work, and **the next regeneration will still lose every socket the scrape used to supply.** Nobody should read #503 as having fixed this.

## Options, none of them free

1. **Fix the User-Agent and find a new source.** The IANA-style authority for IPX socket numbers no longer really exists — Novell's registry is long defunct — so the realistic sources are an archived revision of the Wikipedia article, or a hand-maintained table checked into the vendor script.
2. **Pin to an archived revision.** `oldid=` on the Wikipedia URL, or a Wayback snapshot. Stable and reproducible, but frozen; acceptable given the registry is effectively closed.
3. **Retire the scrape and hand-maintain the enum**, as a deliberate decision recorded in the vendor script. Given IPX is a dead protocol with a closed socket registry, this may honestly be the right answer — a crawler for data that will never change again is cost without benefit.

My own leaning is 3 or 2, precisely because IPX is not going to gain new socket numbers. But that is a maintenance-policy call rather than a technical one, so it belongs to whoever owns the vendor pipeline.

## Worth checking at the same time

Other vendor crawlers may hit the same 403. This one was noticed only because #492 sent someone into it. A sweep of `pcapkit/vendor/**` for crawlers that fetch with a default User-Agent would size that, and is **not** something I have done — this issue reports one crawler, verified, and nothing wider.

JarryShaw:
Closes #507.

## Both halves of the issue re-verified against the live page

| fetch | result |
|---|---|
| default `requests` User-Agent | **403**, 126 bytes |
| descriptive User-Agent | **200**, 114117 bytes |

403 body: `Please set a user-agent and respect our robot policy https://w.wiki/4wJS.`

Live page (revid 1372814585) carries **1** `wikitable`, headers `['Octets', 'Field']` — the IPX *header format* table. `0x0451` and `NetWare Core Protocol` appear nowhere in the HTML. The `Socket_number` section still exists but now holds two sentences of prose and no table.

Removal revision is **1371327031** (2026-08-25), comment *"…deleted some irrelevant technical tables"*.

## One correction to the issue's framing

#507 says the next regeneration "will still lose every socket the scrape used to supply". Measured, it does not — it fails outright (`IndexError` at `vendor/ipx/socket.py:82`), and `vendor/__main__.py:59` catches and warns, so the constant file is simply never rewritten. Silent loss would require the page to regain a fourth `wikitable`.

## Why retirement rather than pinning an archived revision

The archived revision was verified viable on the data side — `oldid=1368657333` returns 200 with 4 `wikitable`s, `table[3]` headers `['Socket number', 'Protocol']`, 19 body rows, and running the old `process()` against it reproduces today's constant file exactly.

It is not viable on the *code* side. The User-Agent is applied in `pcapkit/vendor/default.py:404` — `page = requests.get(self.LINK)`, no headers — so **any** non-`None` `LINK`, archived or not, takes the same 403. Fixing that belongs in the shared base class and affects three other crawlers (see below), so it is deliberately out of scope here.

The policy argument settles it, and it is recorded in the script: the registry was Novell's and is closed. Novell's own documentation still says *"developers can contact Novell to reserve well-known sockets"* — and there is no longer a Novell to contact, so a crawler has nothing to pick up on its next run.

## Per-entry sourcing

- **Novell, *IPX Addressing*, Table 2** (`novell.com/documentation/nw6p/ipx_enu/data/hvvqznoa.html`, 200, still live) — `0x451` NCP, `0x452` SAP, `0x453` RIP, `0x455` NetBIOS, `0x456` Diagnostics. This is also the citation the Wikipedia article itself used.
- **RFC 1132** line 104 — `0x8060`
- **RFC 1791** — `0x9091` TCP, `0x9092` UDP, `0x9093` IPXF

Stated honestly in the script: `0x0001`, `0x0002`, `0x0003`, `0x0457`, `0x4003` rest on the archived revision alone. RFCs 1132, 1362, 1377, 1551, 1552, 1553, 1634, 1791, 1973 and 2043 were swept for those codes — **0 hits in all ten**. `0x0000` is not a registry entry but IPX's own field default, per #492/#503.

## No loss — proven by byte identity

`pcapkit/const/ipx/socket.py` regenerates **byte-identically**: same md5, 15 members, same names, values and order. `Socket(0x0000)` → `Unspecified`, so #503's member is intact.

Byte identity was chosen deliberately as the strongest available proof, which meant keeping the `LLC_4` member name and its `#: LLC [ 4 ]` comment — a Wikipedia footnote artefact (`<ref name="RFC1132" />` rendering as `[4]`). Cleaning that up would rename a public member, so it is flagged with a `# NOTE:` instead.

## Verification

Regeneration completes with both `requests.get` and `requests.Session.request` patched to raise, confirming no network path remains.

`tests/vendor/test_ipx_socket_unit.py` — 9 tests, 16 subtests — pins `LINK is None`, byte-for-byte reproduction of the committed constant, the full 15-member list, registry/enum agreement, `Socket(0)`, `RANGES` order, and rejection of out-of-range values.

**The tests are mutation-tested, because the first attempt was vacuous** — `purge_modules` re-imports from disk, so in-memory mutations were discarded and every mutation "passed". Redone by mutating the file on disk with a hash-checked restore: dropping `0x9093` fails 2 tests, dropping `0x0000` fails 3, renaming `LLC_4` fails 2, reordering `RANGES` fails 2.

| suite | result |
|---|---|
| unit tier, before | 934 passed, 8 skipped, 1952 subtests |
| unit tier, after | 943 passed, 8 skipped, 1968 subtests |

Delta is exactly the +9 tests / +16 subtests added. `isort -l100 -ppcapkit --check-only` clean, matching what `cron-vendor.yml` runs.

## Follow-ups this surfaced, deliberately not fixed here

- **`pcapkit/vendor/ipx/packet.py` is broken identically** — same article, same removal revision, `find_all('table', class_='wikitable')[1]` at `:82` against a page that now has one table.
- **Three other crawlers take the same 403**: `vendor/vlan/priority_level.py`, `vendor/ftp/return_code.py`, `vendor/ipx/packet.py`. The fix is a descriptive User-Agent at `vendor/default.py:404`.
- **The three `pcapng/*` crawlers point at a dead URL** — `www.ietf.org/staging/draft-tuexen-opsawg-pcapng-02.html` 404s regardless of User-Agent. Working replacements: `ietf.org/archive/id/…` and `datatracker.ietf.org/doc/html/…`.


JarryShaw:
## Review: CHANGES NEEDED — prose and citations, not code

The three load-bearing claims all hold under independent measurement. What needs fixing is the sourcing prose plus one docs line, and **none of it touches `pcapkit/const/ipx/socket.py`**, so none of it costs the byte-identity proof.

**Verified.** Byte identity: regenerated by constructing `Socket()` for real — md5 `7548ec44c461a5c1d30d57751ae54f42`, 3298 bytes, before and after, matching `origin/main` at `980e52f0c` too; tree left clean; all 15 members and `Socket(0x0000)` -> `Unspecified` intact. No network: same md5 with `requests.get/post/request`, `Session.request/send`, `HTTPAdapter.send` and `socket.connect/connect_ex/getaddrinfo/create_connection` all patched to raise. Tests not vacuous: all four claimed mutations reproduce exactly (2 / 3 / 2 / 2), and 11 more of mine — 5 vendor-side, 6 const-side — are all caught, 15/15. Suite 943 passed / 8 skipped / 1968 subtests, matching the PR. Transcription from `oldid=1368657333` is row for row with **zero** discrepancies.

**Needs changing:**

1. `pcapkit/vendor/ipx/socket.py:116-119` — the name `TCP over IPXF` contradicts its own citation. RFC 1791 §3 is *"Running TCP Over IPX"* (line 99) and line 129 says *"running TCP over IPXF is not recommended"*. Value is right; fix the comment, not the member. (`UDP over IPXF` for `0x9092` **is** correct.)
2. `pcapkit/vendor/ipx/socket.py:127-130` — "if not their exact bounds" understates a contradiction: the cited Novell page says *"between 0x4000 and **0x7FFF**"*, not `0x4FFF`.
3. `pcapkit/vendor/ipx/socket.py:81-83` — "the only source found" is refuted, and correcting it *strengthens* the table. IANA's live `xns-protocol-types` has an "Assigned well-known socket numbers" registry (`1 Routing Information`, `2 Echo`, `3 Router Error`, `40-77 Experimental`) citing Xerox XSIS 028112, whose Appendix D gives *"range 1 to 3000 decimal"*, *"all other ... ephemeral"*, column header **octal**. That confirms this PR's own numbers: 3000 = `0x0BB8`, 3001 = `0x0BB9`, octal 40-77 = `0x0020-0x003F`.
4. `pcapkit/vendor/ipx/socket.py:88-96` — the cited Novell Table 2 has **seven** rows; `0x9001` (NLSP) and `0x9004` (IPXWAN) are dropped silently. `0x9004` is reserved by RFC 1634 line 552, which is in the "swept, 0 hits" set — so narrow that claim to the five specific codes.
5. `docs/source/pcapkit/vendor/ipx.rst:39` — still `:members: FLAG, LINK` for a class that no longer defines `LINK`. Across all 117 vendor `autoclass` entries this is now the **only** code/docs mismatch in the repo; 22 of the other 23 LINK-less modules are `:members: FLAG`. The `:47` footnote also still cites the dead `#Socket_number` anchor.
6. PR description — "`IndexError` at `vendor/ipx/socket.py:82`" is not today's behaviour: measured, the 403 raises `RequestException`, caught at `pcapkit/vendor/__main__.py:59`. `IndexError` appears only once the fetch is fixed (confirmed separately at exactly `:82`). The code comment at `:42-44` is right; the description is not. Conclusion unaffected.

**On the `HAS_REQUESTS` guard — I would add plain `requests` to the `test` extra.** The guard is well built and honestly documented, but measured in a `requests`-free venv it is **9 skipped in 0.02s**, and `cron-vendor.yml` has `.[all]` yet never runs pytest — so the pin protects neither path. It is the pin that matters: `test_regeneration_reproduces_the_committed_constant_file` caught 14 of my 15 mutations and was the *only* thing that caught 2 (a description-only edit; deleting a `_missing_` branch). Green CI here means "9 tests did not run", not "the pin passed". `tests/_tiers.py:16-19` already requires the unit tier to pass on `.[test]`, and one line keeps that true while making the pin execute. Cleaner follow-up: move `import requests` at `pcapkit/vendor/default.py:23` into `Vendor._request` — it is used nowhere else — which drops the need for the guard entirely.

**On `LLC_4` — keep it.** Byte identity is this PR's whole evidentiary basis, and the member has four occurrences repo-wide, so renaming with a deprecated alias is cheap later. The reading is confirmed more strongly than stated: the cell markup is `id="cite_ref-RFC1132_4-1"`, and RFC 1132 has only refs `[1]` and `[2]` and never names the socket "LLC".

**Could not verify:** the `origin/main` baseline (934 / 1952) was not re-run — it follows arithmetically from the measured 943 / 1968 minus the measured 9 / 16, with zero failures; the Sphinx docs were not built, so item 5 rests on the directive plus the 117-module scan (the build passes no `-W`); the socket-number column of the 1981 XSIS scan is OCR-garbled, so the `1/2/3` name mapping comes from IANA rather than the scan; #492/#503 history taken as given beyond confirming `Unspecified = 0x0000` resolves. Python 3.14.7, Linux only.

Minimum to merge: items 1-5 and the `test`-extra line.

JarryShaw:
## Review feedback addressed — pushed `dd888da06`

`pcapkit/const/ipx/socket.py` is untouched: md5 `7548ec44c461a5c1d30d57751ae54f42`, 3298 bytes, identical to `origin/main`. The byte-identity proof this PR rests on survives the prose round.

### The review's own prescription does not work, and this is the substantive finding

Adding plain `requests` to the `test` extra does **not** make the nine tests run — it makes **all nine fail**:

```
ModuleNotFoundError: No module named 'bs4'   (from pcapkit/vendor/reg/linktype.py:17)
```

`pcapkit/vendor/__init__.py` carries **17 unconditional crawler imports**, and **8** crawler modules import `bs4` at module scope (`ftp/return_code.py`, `ipx/packet.py`, `ipx/socket.py`, `pcapng/block_type.py`, `pcapng/option_type.py`, `pcapng/record_type.py`, `reg/linktype.py`, `vlan/priority_level.py`). The package's own `try: import bs4 / except ModuleNotFoundError: warn(...)` at `__init__.py:25-30` suppresses only the **warning**, not the `ModuleNotFoundError` those imports raise.

Measured in a clean venv built from a base interpreter, `pip install -e '.[test]'` each time, with `pcapkit.__file__` asserted:

| `test` extra | result |
|---|---|
| as this PR stood | **9 skipped** — every reason "requests not installed (vendor extra)" |
| `+ requests` only | **9 failed** — `ModuleNotFoundError: No module named 'bs4'` |
| `+ requests, beautifulsoup4` | **9 passed, 16 subtests** |

Both are added, plain — no `[socks]`, no `[html5lib]`, since no unit-tier test goes through a proxy or parses HTML. The skip guard is widened from `requests` to `requests` **and** `bs4` so a deps-less checkout still skips rather than errors.

### The suite baseline in this PR's description is not what CI measures

CI installs `.[test]`, and a clean `.[test]` venv gives **869 passed / 82 skipped / 1910 subtests**, going to **878 / 73 / 1926** with this change — a delta of exactly +9 passed, −9 skipped, +16 subtests. The 82 skips are the optional engines (`dpkt`, `scapy`, `pyshark`, `pypcapfile`, `cryptography`, `emoji`), none of which `.[test]` installs. Higher counts quoted earlier came from an environment carrying all extras; **the figures above are what `unit-tests.yml` actually runs.**

### The name-pinning hardening, with its proof

`test_unlisted_sockets_still_resolve` now pins `member.name` as well as the value, because **every branch of the generated `_missing_` returns `value`** — so asserting the value alone passes with range branches deleted. Proof: dropping the `0x0BB9 <= value <= 0xFFFF` masking branch gives

```
8 failed, 1 passed, 6 subtests passed
SUBFAILED 0x0BB9, 0x4000, 0x4FFF, 0x7FFF, 0x8000, 0x8061, 0x9094, 0xFFFF
```

— exactly the eight sockets that branch covered. Restored afterwards, md5 re-verified.

### Corrections to the review and to this description

- **`117` vendor autoclass entries is 119.** 119 `.. autoclass::` under `docs/source/pcapkit/vendor/`; 117 carry `:members:` (the crawler-class count meant), and the other two are `Vendor`/`VendorMeta` with `:no-members:` at `default.rst:10,35`. `ipx.rst:39` was the sole `:members:` mismatch among the 23 LINK-less crawlers and is fixed here.
- **The `LLC_4` argument's premise was wrong.** RFC 1132 has four references, not two: `[1]` IEEE 802.2, `[2]` Novell IPX, `[3]` RFC 791, `[4]` RFC 826 ARP. The conclusion holds — its `[4]` is ARP, and the markup id `cite_ref-RFC1132_4-1` proves the marker is Wikipedia's footnote — but the stated reason should not be repeated. Also, the spaces in `LLC [ 4 ]` are **not in the page**: bare `get_text()` yields `LLC[4]`, and the spaces are an artefact of the old scrape joining cell strings and picking up the `cite-bracket` spans as separate tokens.
- **"Will still lose every socket the scrape used to supply" is not what happens** — the next regeneration *fails outright* and the constant file is simply never rewritten. The 403 fetch failure comes first (`Vendor._request` retries, then raises, and `vendor/__main__.py:59` turns it into a `VendorRuntimeWarning`); the `IndexError` at `vendor/ipx/socket.py:82` is only reachable once the fetch is fixed. Silent loss would need the page to regain a fourth `wikitable`.
- **Per-entry sourcing is stronger than this PR claimed.** IANA's live `xns-protocol-types` registry and Xerox XSIS 028112 (December 1981) Appendix D both source `0x0001`–`0x0003` and three of the five ranges, so **only `0x0457` and `0x4003` rest on the archived revision alone**. The octal reading of Appendix D is settled by arithmetic rather than OCR: 3000 → `0x0BB8`, 3001 → `0x0BB9`, octal 40–77 → `0x0020`–`0x003F`, all matching the transcribed bounds, where decimal 40–77 would give `0x0028`–`0x004D` and does not.

### Recorded in code, deliberately not fixed

`Socket.TCP_over_IPXF` (0x9091) is **misnamed**: RFC 1791 assigns it in §3, "Running TCP Over IPX", which opens "Unlike UDP, TCP runs directly over IPX" and closes "Hence, running TCP over IPXF is not recommended." The value is right; the name came from the archived revision and describes the one arrangement its own RFC advises against. `UDP over IPXF` for 0x9092 *is* accurate. Also absent with good sourcing: `0x9001` NLSP and `0x9004` IPXWAN (Novell's Table 2 has **seven** rows; 0x9004 is reserved by RFC 1634 line 552, RFC 1551 line 532 verbatim, and RFC 1362 line 253), and an entry for all-ones `0xFFFF`, which XSIS reserves to mean "all". And `(0x4000, 0x4FFF)` is **contradicted** by Novell's "between 0x4000 and 0x7FFF are dynamic sockets", not merely rounded.

None renamed or added: each would end the byte-identity proof, so they belong in their own review.

JarryShaw:
## Re-review of `dd888da06`: GOOD TO GO — fixes applied in `1da80b122`

Second cross-review, different model, scoped to the delta `6007b19d2..dd888da06`.

### The load-bearing claim reproduced character for character

It built **three separate clean venvs** from `git archive 6007b19d2`, `pip install -e '.[test]'` fresh each time:

| `test` extra | reproduced |
|---|---|
| as it stood | `9 skipped, 1 warning in 0.02s` |
| + plain `requests` only | `9 failed, 2 warnings in 8.23s` — traceback bottoms out at exactly `pcapkit/vendor/reg/linktype.py:17` |
| + `requests` and `beautifulsoup4` | `9 passed, 2 warnings, 16 subtests passed` |

So **overriding the original review's prescription was correct**: plain `requests` genuinely does not work. It also checked something the claim never addressed — pytest categorises all nine as **FAILED** rather than ERROR even though the exception fires inside `setUp()`, matching the wording used here.

### It found the undetected mutation, as asked

The three masked range branches are each a strict subset of an earlier always-checked-first branch, so they are unreachable at runtime. Widening `0x0020 <= value <= 0x003F` to `0x0025 <= value <= 0x0999` — still fully masked — is **completely invisible** to `test_unlisted_sockets_still_resolve` alone: `1 passed, 14 subtests passed`.

The whole-file run does catch it, but via `test_regeneration_reproduces_the_committed_constant_file` re-deriving from unmutated source and diffing byte-for-byte — **a different mechanism than the name-pinning under review.** So the hardening has a real blind spot for masked-branch *bounds*, while the suite as a whole does not; and a realistic edit (change `RANGES` in the vendor module, then regenerate) would additionally trip `test_range_order_is_preserved`'s exact tuple compare regardless of masking. Recorded rather than changed.

### Two corrections, one of them in committed code — both now fixed in `1da80b122`

**1. The bs4 importer count was wrong in its own context, and the confusion is worth explaining because three different numbers were all correct.**

| tree | unguarded module-scope `bs4` imports |
|---|---|
| `main` | **8** — both `ipx/packet.py` and `ipx/socket.py` still scrape |
| #518's branch | **7** — #518 retires **`ipx/packet.py`**'s scrape |
| **this branch** | **7** — this PR retires **`ipx/socket.py`**'s scrape |

Both branches reach seven by removing a *different* file. An earlier comment of mine asserted eight as a correction to someone measuring seven — that was measured against `main`, so it was true there and wrong as a correction. The figure is now **seven** in `pyproject.toml` and `tests/vendor/test_ipx_socket_unit.py`, verified on this branch. The mechanism is unaffected either way: one unguarded import (`reg/linktype.py`, pulled in via `reg/__init__.py`) is sufficient to reproduce the failure, and it does.

**2. "RFC 1132 has four references" is wrong — it has eight.** Fetched from `rfc-editor.org` and the IETF datatracker independently: [1] IEEE 802.2, [2] Novell IPX, [3] RFC 791, [4] RFC 826/ARP, [5] ISO DIS 8473, [6] Xerox XNSG 068504, [7] RFC 1042, [8] Cohen. The sub-claim that `[4]` is RFC 826/ARP is correct, and the conclusion holds on ground the reviewer verified directly — the raw wikitext reuses `<ref name="RFC1132" />` at the LLC row and the rendered anchor is `cite_ref-RFC1132_4-1`, where the `_4` is **Wikipedia's own footnote position** and `-1` its second use on the page, nothing to do with RFC 1132's bibliography. My "four references" line conflated two unrelated numbering systems and was wrong on the count as well. GitHub-conversation text only, not committed.

### Verified more strongly than this PR claimed

The PR worried that Appendix D of Xerox XSIS 028112 was OCR-garbled and settled the octal radix by arithmetic instead. The reviewer **installed `poppler` and read the scanned 1981 document directly** — page 58, PDF page 67 — and it is **fully legible**: "socket numbers zero and all ones are reserved to mean unknown and all respectively", "Well-known socket numbers have the range 1 to 3000 decimal", "All other socket numbers are ephemeral", and a table header reading **"Well-Known Socket (octal)"** outright, with rows 1/2/3/40-77 exactly as transcribed. So three of the five ranges rest on a directly-legible primary source, and the arithmetic workaround is no longer load-bearing. Novell's live page independently confirms the other two bounds, including that `(0x4000, 0x4FFF)` really is contradicted by Novell's own `0x4000-0x7FFF`.

Also confirmed exactly: **119** autoclass with 117 `:members:` and 2 `:no-members:`; bare `get_text()` yielding `LLC[4]`; the `0x9004` wording in RFC 1634 and 1551 as "protocol exchange" against RFC 1362's "information exchange"; **17** unconditional crawler imports; byte-identity at md5 `7548ec44c461a5c1d30d57751ae54f42`, 3298 bytes, with `git diff origin/main` on that file empty and the whole vendor-module diff being comments only.

### On `TCP_over_IPXF`, agreed — with a better reason than mine

RFC 1791 §3 confirmed verbatim: titled "Running TCP Over IPX", opening "Unlike UDP, TCP runs directly over IPX", closing "Hence, running TCP over IPXF is not recommended." The value `0x9091` is right and only the label is wrong. Defer the rename — and the sharper argument is that **even a deprecated alias changes `extend_enum`'s generated output**, which would end the byte-identity proof that is this PR's entire evidentiary basis. Its own change, later.

### Not independently verified
The full-suite delta (869→878 passed, 82→73 skipped, 1910→1926 subtests) was not re-run, under host-care constraints after an earlier OOM incident; the nine-test delta is confirmed instead. The exhaustive "0 hits" sweep across all ten RFCs for `0x0457`/`0x4003` was not replicated — the three bearing on `0x9004` were spot-checked and matched exactly.

JarryShaw:
**✅ GOOD TO MERGE** — cross-reviewed twice, the `test`-extra fix reproduced in three clean venvs, and `pcapkit/const/ipx/socket.py` byte-identical to `main` (md5 `7548ec44…`).

JarryShaw:
**✅ GOOD TO MERGE** — cross-reviewed twice, the `test`-extra fix reproduced in three clean venvs, and `pcapkit/const/ipx/socket.py` byte-identical to `main` (md5 `7548ec44…`).

Issue #522, opened 2026-09-19
Vendor._request()'s manual-intervention fallback is an unbounded print loop when stdin is not interactive


JarryShaw:
## Summary

`Vendor._request()`'s manual-intervention fallback becomes an **unbounded, full-speed print loop** when stdin is not interactive. It is not a hang waiting for input — `contextlib.suppress(Exception)` swallows the `EOFError` that `input()` raises on a closed stdin, so the loop never waits at all.

`pcapkit/vendor/default.py:460-466`:

```python
while True:
    with contextlib.suppress(Exception):
        input('Press ENTER to continue...')  # nosec
    if os.path.isfile(temp_file):
        break
    print('File not found; please save the page source at')
    print(f'    {temp_file}')
```

With no terminal, `input()` raises `EOFError`, the suppression discards it, `os.path.isfile(temp_file)` is false because nobody could save the page, and the loop repeats immediately — three lines of output per iteration, forever.

## Measured

Observed in the wild during a review of #518: a probe reached this path after a transient fetch failure and produced **26.7 million lines / 2.6 GB** of output before it was killed.

The mechanism reproduces in isolation, with stdin closed:

```console
$ python -c "
import contextlib
n = 0
for _ in range(3):
    with contextlib.suppress(Exception):
        input('Press ENTER to continue...')
    n += 1
print(); print('iterations without blocking:', n)
" < /dev/null
Press ENTER to continue...Press ENTER to continue...Press ENTER to continue...
iterations without blocking: 3
```

Three iterations, no wait. In `_request()` the loop is `while True` with no bound.

## How it is reached

`_request()` tries `MAX_RETRY` direct fetches (`PCAPKIT_VENDOR_RETRY`, default 5), then `MAX_RETRY` more through proxies if `get_proxies()` returns any, and only then falls through to this block — guarded by `CI_MODE` at `default.py:435`, which comes from `PCAPKIT_CI_MODE` at `default.py:36` and defaults to **false**.

So the trigger is: run a crawler whose `LINK` is unreachable, from a non-interactive process, without `PCAPKIT_CI_MODE` set.

## Severity — narrower than it first looks

[...]

PR #524, opened 2026-09-19
fix(vendor): send a descriptive User-Agent and retarget the dead pcapng draft

### LINKED ISSUE #518: Vendor crawlers: three take a Wikipedia 403 on the default User-Agent, three point at a dead IETF URL
Surfaced while fixing #507 (PR #511), which retired one crawler whose data source is gone. A sweep of all 136 files under `pcapkit/vendor/**` shows it is not an isolated case.

Census: **95** classes with a string `LINK`, **92** distinct URLs. Fetching each with the default `requests` User-Agent: **200 → 87, 403 → 4, 404 → 1**.

## The four 403s are all Wikipedia, and all pure User-Agent refusals

| crawler | default UA | descriptive UA |
|---|---|---|
| `pcapkit/vendor/vlan/priority_level.py` | **403**, 126 B | 200, 119717 B |
| `pcapkit/vendor/ftp/return_code.py` | **403**, 126 B | 200, 78192 B |
| `pcapkit/vendor/ipx/packet.py` | **403**, 126 B | 200, 114117 B |
| `pcapkit/vendor/ipx/socket.py` | retired in #511 | — |

403 body: `Please set a user-agent and respect our robot policy https://w.wiki/4wJS. See also https://phabricator.wikimedia.org/T400119.`

Only the header changes between the two columns — no redirect — reproduced across three different pages.

**The fix belongs in one place**: `pcapkit/vendor/default.py:404` is `page = requests.get(self.LINK)` with no headers, in both the direct and proxy branches. A *descriptive* agent is what Wikipedia's policy asks for; a browser spoof is not.

## `vendor/ipx/packet.py` is additionally broken the same way #507 was

Same article, same removal revision (`1371327031`, 2026-08-25). `:82` does `soup.find_all('table', class_='wikitable')[1]` against a page that now carries **one** table, so even with the User-Agent fixed it raises `IndexError`. It needs the same treatment #511 gave the socket crawler — and unlike that one, its constant file was out of scope there.

## The 404 is a genuinely dead URL, unaffected by User-Agent

`pcapkit/vendor/pcapng/{block_type,option_type,record_type}.py` all point at `https://www.ietf.org/staging/draft-tuexen-opsawg-pcapng-02.html` — **404 with any agent**. Its body is a 77968-byte HTML error page, so `default.py:405`'s `if not page.ok or not page.text` catches it on `page.ok`.

Working replacements, both measured:
- `https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-02.html` — 200, 215857 B
- `https://datatracker.ietf.org/doc/html/draft-tuexen-opsawg-pcapng-02` — 200, 233176 B

**The same dead URL also appears in the documentation**, where `sphinx-build -b linkcheck` reports it as 8 broken links (`docs/source/pcapkit/protocols/misc/pcapng.rst:773`, `docs/source/pcapkit/const/pcapng.rst:123-129`). One fix covers both.

## Worth pairing with a reproduction gate

`cron-vendor.yml` runs these weekly *and* on every push to `main`, and auto-commits the result (`:88`, `:98`) — so a crawler defect lands unreviewed. Measured today: **22/22** offline generators regenerate their constant files byte-for-byte, and 87/93 network ones do (the 6 are 5 network-blocked hosts plus 1 genuine diff). A CI gate asserting byte-identical regeneration is ~40 lines and would have caught `pcapkit/const/mh/handover_initiate_status.py`, which is the only one of 134 constant files missing the `:meta private:` marker its template gained in `07bf98d46` — i.e. the weekly crawl has not successfully rewritten it since.

JarryShaw:
Closes #518.

## What was broken

Four crawlers took a **403** on `requests`' default User-Agent — Wikipedia now requires a descriptive one — and three pointed at an `ietf.org/staging` URL that **404s**.

## The User-Agent

New `get_user_agent()` in `pcapkit/vendor/default.py`, built from `importlib.metadata` rather than hard-coded, applied at **both** the direct and the proxy call site. It is **not a browser spoof**: it names the distribution, version and repository, per Wikimedia's robot policy.

Proven metadata-driven rather than literal: faking the metadata to a different name and repository URL makes the function follow it, and the label match is case-insensitive.

**The argument that the header cannot change generated output:** of the 92 distinct crawler URLs, **87 return byte-identical bodies** with and without it. The only five that differ are the four Wikipedia 403→200 and the one dead 404. Identical bodies imply identical output — stronger than running 117 generators twice.

## The pcapng draft — both URLs the issue proposed are wrong

`draft-tuexen-opsawg-pcapng-02` renders its registries as **preformatted ASCII art**: exactly one HTML table (the running header) and **zero** `table-N` ids, while all three crawlers select tables by id. So both `-02` URLs fetch 200 and then raise `IndexError`. Measured by running the real `BlockType.process()` against the fetched page.

Only **`-03`** reproduces all three constant files byte-identically. `-04`, `-05` and ietf `-00` diverge on `block_type`; ietf `-01`/`-02` on all three; ietf `-03` onward raise `ValueError` on a `0x0A0D0AXX` wildcard row — so migrating to the renamed `draft-ietf-opsawg-pcapng` is real work, not a URL swap, and is left as a follow-up.

Chose the static `www.ietf.org/archive/id/` rendering over datatracker: datatracker renders per request, sends no `Last-Modified`, and wraps the document in chrome adding three unrelated tables to a document the crawlers index tables out of.

## `ipx/packet.py` — scrape retired

The table was deleted from the article on 2026-08-25 and the registry is closed, so `DATA` is hand-maintained with per-entry citations and the class defines no `LINK`. `pcapkit/const/ipx/packet.py` regenerates **byte-identically** (md5 `0655b561ccdca4981ef6f7974ebeb66f`).

## `ftp/return_code.py` — a second defect, and why a data file changes

The article now ends its table with a cell-less `<tr class="mw-empty-elt">`, so indexing the first cell raised `IndexError`. **That is why this constant file had never been regenerated.** With the guard in place it regenerates — and gains `CODE_335`, RFC 2228's ADAT response. **53 → 54 members, nothing removed** (verified with `comm`).

To be explicit, since a data change inside an infrastructure fix deserves it: this is **not a hand-edit**. Running the unmodified `request()`/`count()`/`process()`/`context()` pipeline against today's live page reproduces the committed file byte-for-byte, `CODE_335` included. The only other differences are doc comments — the enum *values* for codes 450 and 452 are byte-identical either side.

## Tests — 25 tests, 35 subtests, all three suites new

Mutation-tested by mutating files **on disk** with hash-checked restore. 14 of 14 mutations caught, including dropping the header from each call site independently.

### Proven to fail without their fixes

| revert | result |
|---|---|
| FTP empty-row guard removed | **4 failed, 1 passed** — real `IndexError` at `vendor/ftp/return_code.py:221` |
| case-insensitive label match dropped | **3 failed** — `SUBFAILED` on `Repository`, `REPOSITORY`, `RePoSiToRy` |
| as shipped | **25 passed, 35 subtests** |

The one test that survives the guard removal is the *fixture-shape* assertion, correctly — it exists so that a fixture which stopped reproducing the defect would be caught, rather than silently making the others pass.

**115 of 117 constant files regenerate byte-identically.** The two exceptions are `ipx.socket`, whose crawler #511 retires, and `mh.handover_initiate_status` — see below.

## Found while doing this, reported not fixed

**`pcapkit/const/mh/handover_initiate_status.py` is the only one of 117 constant files missing the generated-file marker, and the cause is a duplicated export.** `MH_HandoverACKStatus` appears **twice** in `__all__` at line 196 of *both* `pcapkit/vendor/mh/__init__.py` and `pcapkit/const/mh/__init__.py`, and `MH_HandoverInitiateStatus` has no import line at all. Since `pcapkit/vendor/__main__.py:87` builds its target list from `__all__`, **that crawler has never run** — 117 defined, 116 reachable. Not failing; unwired. Confirmed three independent ways. Its own file, so a separate change.

A vendor-reproduction CI gate is drafted but deliberately **not** included: two of its three steps fail today by design until that typo is fixed.

Also out of scope here: the dead docs URL still appears 15 times across 3 files, and #522 (the `_request()` fallback loop) was filed from a failure hit while testing this.


JarryShaw:
## Cross-review: GOOD TO GO

Reviewed at head `c41a56605` on a different model, briefed to falsify and to attack the two tests added *after* the earlier review, since nobody had adversarially read those. It ran four parallel subagents over disjoint file sets and confirmed no clobbering at the end (`git status --short` empty, `HEAD` unchanged).

### Corrections to this PR's own text

- **The `IndexError` in the guard-removal repro fires at `vendor/ftp/return_code.py:221`, not 229.** 229 is where the guard sits in the *unmutated* file; deleting its nine lines shifts everything up eight. Description now corrected.
- **The 53 → 54 diff is not purely a clean addition.** Three pre-existing entries' doc comments also shifted — `CODE_229`'s spacing, and the `File unavailable (e.g., file busy).` clause moving from `CODE_452` to `CODE_450`. Still fully covered by the byte-identical regeneration proof and still nothing removed, but "growing from 53 to 54 with nothing removed" undersells that a few strings moved too.

### Independently verified

**Both new tests hold up, and the negative test genuinely constrains.** The adversarial push was to mutate the label comparison to something *looser* rather than stricter — `==` changed to a substring test. The case-insensitivity test still passed superficially, and `test_unrelated_project_url_labels_are_ignored` **failed**, because `'repository-mirror'` contains `'repository'`. So the fix cannot be "loosen the comparison". Separately confirmed the `lru_cache` clearing is load-bearing rather than cosmetic: populating the cache with the real call, then mocking metadata *without* clearing, returns the stale real answer.

**The FTP fixture genuinely reproduces the defect through the real parse path**, re-parsed independently with the installed bs4 4.15.0 + html5lib 1.1: the trailing `<tr class="mw-empty-elt">` really has zero `<td>`/`<th>`. And no network call — traced through `cls.__new__`, which never reaches `_request()`. One of the four failures fails one column over, on `line[1]` rather than `line[0]`, which is exactly why the guard tests `len(line) < 2` rather than `not line`.

**`-02` confirmed unusable:** 1 `<table>` total, **0** ids matching `table-N`, and **all 8** of the three crawlers' exact selectors return zero matches; calling `.request()` against the fetched page raises `IndexError`. `-03` regenerates all three pcapng const files byte-for-byte. Bonus: `-04` was fetched and shown to genuinely diverge on `block_type` via a renamed member (`systemd_Journal_Export_Block` → `..._I_D`).

**The FTP regeneration is mechanical**, re-derived against the live page with no retries and no warnings: md5 `fc9c72adbb1b4730d007ef31d35bcffd` on both sides, and a `comm` over two independent extraction methods gives 53 → 54, one addition, zero removed.

**Roughly 18 of the 115 constant files** were independently regenerated and diffed byte-for-byte across the four agents — including 9 IANA CSV crawlers, `reg.linktype` from tcpdump.org and `vlan.priority_level` from live Wikipedia — all identical. Consistent with, though not exhaustive proof of, 115/117.

Both judgement calls endorsed: retiring a scrape against a **closed** registry trades no freshness for removed fragility, and bundling the regenerated constant file is right because `CODE_335` *is* the observable proof the guard works — splitting it would leave the guard untestable and the data change unmotivated.

### Not fully verified, stated plainly

`beautifulsoup4[html5lib]` is unpinned in `pyproject.toml`, so a future html5lib could in principle normalise away the empty `<tr>` the fixture depends on. Only 1.1 was tested. Low probability, not a blocker — but the fixture-shape assertion exists precisely to fail loudly if that ever happens.

Of the PR's claimed 14-of-14 caught mutations, only the 2 header-drop mutations were independently re-exercised; the other 12 were not reproduced.

### One reviewer-side finding worth recording

One of the four subagents reported that the proxy call site **lacks** the header. That is **wrong** — line 489 reads `requests.get(self.LINK, headers=headers, proxies=proxies)` and `git diff HEAD` is empty. It was a transient read while a sibling agent was mid-mutation-test in a shared worktree; three of the four children observed the same transient state and attributed it correctly. Worth knowing as a hazard of parallel mutation testing rather than anything about this PR.

JarryShaw:
**✅ GOOD TO MERGE** — cross-reviewed clean; 87 of 92 crawler responses byte-identical with the new User-Agent, and the regenerated FTP constant file confirmed as mechanical crawler output rather than a hand-edit.

JarryShaw:
**✅ GOOD TO MERGE** — cross-reviewed clean; 87 of 92 crawler responses byte-identical with the new User-Agent, and the regenerated FTP constant file confirmed as mechanical crawler output rather than a hand-edit.

pcapkit/vendor/pcapng/block_type.py
```diff
@@ -24,14 +24,59 @@
 
     from bs4.element import Tag
 
+###############################################################################
```
JarryShaw, line 27: these notes are not in the doc. so the docstring rendered Sphinx doc in the three files wont be able to see these.

JarryShaw: Right, and that is a fair hit — the whole rationale is a `#` comment block at `block_type.py:27`, and the docs only `autoclass` `BlockType`, so Sphinx never sees a word of it. A reader of the rendered docs has no way to learn why the URL is `-03`.

Worse, the docs currently tell them something false: the dead `staging/…-02` URL is still cited **15 times across 3 files** — `docs/source/pcapkit/vendor/pcapng.rst`, `const/pcapng.rst` and `protocols/misc/pcapng.rst` — plus 7 `.py` sources the docs pull from. So the rendered page would point at a 404 while the code fetches `-03`.

Proposed fix, same shape as the `.. note::` #511 added to `ipx.rst`:

1. A concise `.. note::` in `docs/source/pcapkit/vendor/pcapng.rst` stating once that all three crawlers read `-03`, that `-02` is dead, and that the `table-N` ids only exist from `-03` — the reader-facing half.
2. The URL substitution across all 15 occurrences, so no rendered page cites the dead draft.
3. Trim the `#` block to the implementer-facing remainder (the per-revision reproduction measurements), since that is genuinely comment material rather than documentation.

Shall I fold that into this PR, or keep it separate so the crawler fix stays reviewable on its own? My inclination is to fold it in — a PR that retargets a URL while leaving the docs citing the old one is half a change.

JarryShaw: Fold them in.

JarryShaw: Folded in as `1240093d2`. The reader-facing "why `-03`" now lives in a `.. note::` in `docs/source/pcapkit/vendor/pcapng.rst`, placed once at page scope since it covers all three scraping crawlers, and following the shape of the `ipx.rst` note from #511 — it says that `BlockType`, `OptionType` and `RecordType` all read `-03`, that the `-02` staging URL is dead, and that the `table-1` through `table-10` ids they select on first exist in `-03`. I also retargeted the dead URL in all 31 places that cited it (15 across the three `docs/source/` pages, 16 in the `.py` sources), having first confirmed that all 8 anchor fragments actually exist in the `-03` document; three `-02` references were deliberately left alone, namely the "which is dead" URL in this comment block and two `draft-ietf-opsawg-pcapng-02 §4.2/§4.3` citations in `protocols/misc/pcapng.py`, which quote that revision's own text and are a different draft series. This block is trimmed to the implementer-facing measurements and now points at the note. Verified by building the docs: the note renders on `pcapkit/vendor/pcapng.html`, the warning set is byte-identical to the parent commit at 83 warnings, and no clickable `-02` link survives anywhere in the output.

JarryShaw:
**✅ GOOD TO MERGE** — the `.. note::` addition renders the pcapng `-03` rationale that previously lived only in an invisible `#` comment, the 31 URL retargets are correct, and the two-draft-series distinction for the 4 references deliberately left on `-02` is independently confirmed by fetching both actual IETF drafts: the quoted `if_tzone`/`if_tsoffset` deprecation language exists verbatim in `draft-ietf-opsawg-pcapng-02` and does not exist at all in `draft-tuexen-opsawg-pcapng-03`.

JarryShaw:
## Re-review of the delta (previous verdict covered `c41a56605`; this covers `1240093d2`)

**Isolating the real delta:** `git log --oneline c41a56605..1240093d2` shows one merge from `main` (absorbing 4 unrelated commits) plus one genuinely new commit, `1240093d2 docs: render the pcapng -03 rationale and retarget the dead -02 URL`. Confirmed `git diff c41a56605 23146419d` touches nothing pcapng-related, so that one commit is the entire real delta: **8 files, 58 insertions, 46 deletions** — matching the description.

**The load-bearing claim — two distinct draft series — holds, verified against the actual documents, not just names.** Fetched both drafts directly:
- `draft-ietf-opsawg-pcapng-02`: authored by Tuexen et al., explicitly stated to *replace* `draft-tuexen-opsawg-pcapng` — same lineage, but a genuinely distinct, independently-numbered WG document once adopted (revision counter resets).
- `draft-tuexen-opsawg-pcapng-03`: fetched directly and confirmed the quoted text does **not** exist there — `if_tzone` is a bare "TODO: specify better" stub with no deprecation language, and `if_tsoffset` is tied to `if_tzone` rather than divorced from local-time conversion (the opposite of what's quoted in `pcapng.py`). `if_iana_tzname` doesn't exist in tuexen-03 at all.
- Re-fetched `draft-ietf-opsawg-pcapng-02` to confirm exact wording: *"it SHOULD NOT be used; instead, the if_iana_tzname option SHOULD be used"* (`if_tzone`) and *"This offset is not intended to be used as an offset between local time and UTC"* (`if_tsoffset`) — verbatim matches to `pcapkit/protocols/misc/pcapng.py:1240` and `:1300`. Bumping those citations to tuexen-`-03` would have been a factual error quoting language that document never contained.
- The 4th reference left on `-02` (beyond `block_type.py`'s comment and the two `pcapng.py` citations the commit message calls "3") is the new `.. note::` text itself, which necessarily names the dead `-02` URL to explain why `-03` was chosen — self-consistent with the commit's own count. `grep -rn "pcapng-02"` across `pcapkit/` and `docs/` → exactly 4 hits, matching this breakdown. A third-party corroboration: `tests/protocols/misc/test_pcapng_unit.py:3088` independently cites the same `draft-ietf-opsawg-pcapng-02 §4.2` language, untouched by either commit.

**Anchor verification — all 8 confirmed, not just spot-checked.** Fetched `draft-tuexen-opsawg-pcapng-03`'s actual HTML and checked every anchor used in the diff: 7 section-heading anchors plus `section-4.2-28.2.1` (used in `filter_type.py`'s TODO) — all real and correctly targeted. One forward-looking note (not a defect): paragraph-level ids like that one are more fragile across future draft revisions than heading ids.

**The `.. note::` rendering fix.** Confirmed the removed rationale previously lived only inside a `#`-prefixed source comment (not a docstring) preceding the class definition — invisible to Sphinx autodoc, which only pulls docstrings. RST syntax was verified statically (no Sphinx build available in the review environment — not installed, and standing one up was judged more than a "quick" check for this delta, so static analysis was used instead): correct `.. note::` marker, indentation, and blank-line placement; backtick balance verified programmatically (46 backticks, all paired: 3 `:class:` roles + 10 double-backtick literals); `py_compile` succeeds on all 5 touched `.py` files.

**Rest of the 8 files.** Nothing beyond the note, the 31 URL retargets (exact count confirmed via diff), and the `block_type.py` comment trim — the file's other untouched paragraphs (archive-vs-datatracker byte-count comparison, the "-03 onward `process()` dies on `0x0A0D0AXX`" note) were checked and retained verbatim.

## What remains unverified
No actual Sphinx build was run against this delta specifically (relied on static RST/backtick analysis, which gives high confidence but isn't a substitute for a real render) — this is consistent with the previous review's build comparison, which covered the base commit this delta sits on top of, not this delta's own note addition in isolation.

## Verdict
No regressions found in the delta; the one load-bearing judgment call (two-draft-series distinction) is independently confirmed against primary sources. Recommend merge.

Issue #525, opened 2026-09-19
Malformed TCP SACK raises ProtocolError or FieldValueError depending on process state


JarryShaw:
## Summary

Which exception a malformed TCP SACK option produces depends on **whether the schema layer has already been exercised in the same process**. The two exceptions are siblings — neither is a subclass of the other — so no single `except` clause catches both, and which one you get is not a property of the input.

Found while fixing #519; reported separately because it is a behavioural defect rather than a docstring one.

## The two paths

- **Clean process:** `_read_mode_sack` reaches its own length check and raises `ProtocolError`.
- **After `tests/protocols/schema/` has run in the same process:** the schema is unpacked *before* `_read_mode_sack` is reached, and `FieldValueError` comes out of `pcapkit/corekit/fields/collections.py:189` instead.

Both derive from the `pcapkit.utilities.exceptions` hierarchy but along different branches, so `except ProtocolError` silently misses the second case and `except FieldValueError` misses the first.

## Why it matters

A consumer writing defensive code around `Extractor` cannot write a correct handler. `except ProtocolError` looks right — it is what the docstring documents and what a clean run produces — and then fails to catch anything in a long-lived process that has already parsed other packets. That is the worst shape for this kind of bug: it works in tests and in short scripts, and stops working under sustained use.

The state dependence appears to be schema-field caching or lazy initialisation deciding whether the unpack happens eagerly, but **the precise mechanism is not established** — only the observable difference is.

## Repro sketch

Parse a TCP packet carrying a SACK option whose length is not `2 + 8n` (for example 11 or 14), twice:

1. in a fresh interpreter, and
2. after importing and exercising the schema tests in the same interpreter.

The raised type differs between the two.

## Suggested direction

Decide which exception is correct for a malformed option and make the other path raise it — most plausibly by having the schema-layer failure wrapped into `ProtocolError`, since that is the documented contract and the layer a caller of `Extractor` is aware of. A test asserting the union of both types, as `tests/protocols/transport/test_tcp_sack_length_unit.py` currently does, is a workaround that records the problem rather than fixing it.

## Notes

The missing length validation itself is fixed separately as part of #519 — `_read_mode_sack` previously accepted lengths 11, 14 and 2 with no exception at all, unlike its siblings `_read_mode_sackpmt` (raises on `length != 2`) and `_read_mode_echo` (raises on `length != 6`). This issue is only about *which* exception escapes once validation exists.

[...]

Issue #506, opened 2026-09-19
IPv4.from_data() cannot rebuild a parsed packet: ProtocolUnbound on a Raw payload


JarryShaw:
Found while verifying #494's fix (PR #499) and deliberately left out of it. Re-measured on `main` at `d2d9bb80c`, after #499, #500, #501, #502 and #503 all merged — **still broken**.

## `IPv4.from_data()` cannot rebuild a parsed packet

```python
proto = object.__new__(IPv4)
raw = proto.make(offset=5, protocol=6, payload=b'\xaa'*8).pack()
info = IPv4(io.BytesIO(raw), len(raw)).info
proto.from_data(info)
```

```
ProtocolUnbound: unsupported type <class 'pcapkit.protocols.misc.raw.Raw'>
```

Both with and without IPv4 options — measured separately, identical result.

## This is not #494, and #494's fix did not cause it

#494 fixed `_make_data`: the fragment offset was returned unscaled, and `.options` was read unconditionally. Both are fixed and merged. But `_make_data` is only *one input* to `from_data`, and the remaining fault is downstream.

The failure mode **shifted** across that fix, which is worth recording so nobody mistakes one for the other:

| tree | error |
|---|---|
| before #499 | `FieldValueError: Field options has invalid value` |
| after #499 (current `main`) | `ProtocolUnbound: unsupported type <class 'pcapkit.protocols.misc.raw.Raw'>` |

So #499 moved the wall, it did not build it. The earlier error came from `_make_data` handing back a parsed options container; the current one comes from `_make_payload` being given a parsed `Raw` payload it has no binding for.

## Where to look

`_make_payload` / `_make_ipv4_options`. Parsing produces a `Raw` instance for an unrecognised payload and a parsed options container for the options; the make path expects the constructor-shaped inputs `make()` takes, and the two shapes are not the same. The round trip is only closed for packets built by hand, not for packets read off the wire.

## Scope worth checking before fixing

This may not be IPv4-specific. `_make_payload` is inherited, so any protocol whose parse yields a `Raw` payload could have the same hole. That is **unverified** — I measured IPv4 only — and sizing it is the first task.

A fix wants a test that round-trips a *parsed* packet rather than a hand-built one, since that is the case the current tests never exercise. `tests/protocols/internet/test_ipv4_unit.py` covers `_make_data` in isolation, which is why this survived #494.

PR #535, opened 2026-09-20
fix(tcp): enforce the SACK length rule its docstring already promised (#519)

### LINKED ISSUE #519: #501 left behind: 6 phantom Raises: clauses and 11 Args: name mismatches
#501 fixed "Extractor's exception mismatch and 40 phantom/stale `Args:` labels". A full AST sweep of 496 files / 3334 functions (2237 with docstrings) shows the class is not exhausted. Every finding below was reproduced by two independently written scanners.

## Phantom `Raises:` — 6 remaining

#501 fixed exactly one (`Extractor.__iter__`, `IterableError` → `CallableError`).

`pcapkit/protocols/application/httpv2.py` gives an unusually clean proof: of its **11 `_read_http_*` methods, all 11 document `Raises: ProtocolError`; 6 genuinely raise it, 5 cannot.**

- `:413` `_read_http_none` — and the mechanism is visible: `:417` is `#raise ProtocolError(f'HTTP/2: [Type {frame}] invalid format')`, commented out and replaced by `warn(…, ProtocolWarning)` at `:418`
- `:459` `_read_http_data`
- `:514` `_read_http_headers`
- `:822` `_read_http_goaway`
- `:903` `_read_http_continuation`

Plus `pcapkit/protocols/transport/tcp.py:904` `_read_mode_sack`, which documents `ProtocolError: If length is **NOT** multiply of 8 plus 2.` — the body at `:907-917` only builds `Data_SACK`, with no length check anywhere.

All 22 commented-out `raise` statements in the package were swept; **only** the `httpv2:417` case coincides with a docstring still claiming it.

Note these split into two different fixes: the five `httpv2` ones look like stale copy-paste where the docstring should go, whereas `_read_mode_sack`'s absent length check looks like the *validation* is what is missing.

## `Args:` naming a parameter that does not exist — 11

| location | docstring says | signature has |
|---|---|---|
| `pcapkit/foundation/registry/foundation.py:74` | `engine:` | `name` |
| `pcapkit/protocols/misc/pcap/frame.py:132` | `module:` | `protocol` |
| `pcapkit/protocols/schema/application/httpv2.py:133` | `schema:` | only `packet` |
| `pcapkit/protocols/schema/misc/pcapng.py:214` | `pkt:` | `packet` |
| `pcapkit/protocols/schema/misc/pcapng.py:235` | `pkt:` | `packet` |
| `pcapkit/vendor/mh/binding_ack_flag.py:136` | `soup:` | `data` |
| `pcapkit/vendor/mh/binding_update_flag.py:135` | `soup:` | `data` |
| `pcapkit/vendor/mh/handover_ack_flag.py:131` | `soup:` | `data` |
| `pcapkit/vendor/mh/handover_initiate_flag.py:131` | `soup:` | `data` |
| `pcapkit/vendor/pcapng/option_type.py:272` | `data: CSV data.` | `soup` (parsed HTML — wrong name *and* wrong description) |
| `pcapkit/vendor/vlan/priority_level.py:57` | `data:` | `soup` |

The six `vendor/**` cases all carry `# pylint: disable=arguments-renamed`, so the rename was known and only the docstring was missed. `pcapkit/vendor/default.py` uses `data` for `process`/`count`/`context`, so `data` is canonical. The labels look swapped between the `process`/`context` pair.

`pcapkit/protocols/misc/pcap/frame.py:132` is the most consequential: `Frame.register` has no `**kwargs`, so `module=` is a hard `TypeError`.

Three of this class were already fixed under separate work (`foundation/reassembly/reassembly.py:499` and `foundation/traceflow/traceflow.py:430` documenting `name` where the parameter is `protocol`; plus `registry/foundation.py`).

A further **5 sites document `self:` as a parameter** — `pcapkit/dumpkit/common.py:128,166,185`, `pcapkit/protocols/misc/pcap/frame.py:403`, `pcapkit/protocols/misc/pcapng.py:1100` — which napoleon renders as `:param self:`. Real but cosmetic.

## `Args:` completeness — the other blind spot

42 functions have an `Args:` section that omits a real parameter. **21 of the 42 are in `pcapkit/protocols/internet/hip.py`.** Two are demonstrably unintended because their own siblings prove the intent:

- `hip.py:3905` `_make_param_reg_response` documents only `code`/`param`/`version` but takes `lifetime`, `reg_response`, `reg_response_default`/`_namespace`/`_reversed` and `**kwargs`. Its sibling `_make_param_reg_failed` at `:3945` documents all of them.
- `hip.py:4287` `_make_param_route_dst` omits `symmetric`, `must_follow`, `hit`; sibling `_make_param_route_via` at `:4658` documents all three.

And a clean house-style measurement: of 450 functions having both `**kwargs` and an `Args:` section, **438 document `**kwargs` and 12 omit it — all 12 in `hip.py`.**

## Not defects, recorded so the count stays honest

- `pcapkit/corekit/multidict.py:176` documents `KeyError` while raising `MissingKeyError` — which *is* a `KeyError` subclass, so the docstring is literally true. Naming the real class would be nicer; it is not drift.
- `pcapkit/protocols/internet/esp.py:735` documents `cryptography.exceptions.InvalidTag` — a scanner false positive; it genuinely propagates from the third-party AEAD `finalize()`.
- **387 functions raise without any `Raises:` section at all.** A real systemic gap (only 278 docstrings have one), but a blanket fix is a very large mechanical change with little reader benefit. The 6 *wrong* clauses matter far more than the absent ones.

JarryShaw:
Closes #519.

#501 fixed one wrong exception name and forty stale `Args:` labels by hand. #519 showed the class was not exhausted. Re-deriving its census at the current base turned up one finding that is **not a docstring defect at all**, plus a third defect class the issue never mentions.

## The one behavioural change, and why it is not a docstring deletion

`TCP._read_mode_sack` documented `ProtocolError: If length is **NOT** multiply of 8 plus 2` and never checked it. The tempting reading — and the first one taken while working this — is that the clause was stale, like the other five phantoms, and should be deleted. It is the opposite. Three things settle it:

- :rfc:`2018` gives the SACK option a 2-octet header followed by 8-octet left/right edge pairs, so a well-formed length really is `8n + 2`.
- Both neighbouring readers validate their own lengths rather than delegating: `_read_mode_sackpmt` on `length != 2`, `_read_mode_echo` on `length != 6`, with the same message. `_read_mode_sack` was the one option reader documenting a length rule it never enforced.
- Nothing upstream enforced it either. The schema field is `ListField(length=lambda pkt: pkt['length'] - 2, item_type=SchemaField(length=8, schema=SACKBlock))`, which consumes as many whole 8-octet items as it finds and silently ignores the tail.

So the docstring was right and the code was wrong. Inferring the behaviour from a neighbouring docstring — inside an issue about docstrings not matching behaviour — is exactly the trap here, and fuzzing the real parse path is what caught it.

## Re-derived census

Taken at this base, not from the issue, which counted against an older one:

| class | at `c8fd97bcd` | remaining |
|---|---|---|
| phantom `Raises:` | 6 | **0** |
| `Args:` naming a parameter that does not exist | 13 | 9, all allowlisted |
| swallowed section headers | 3 | 1, allowlisted |

The `Args:` figure is **13 here and was 14 when #519 was written**: #511 retired the IPX socket scrape and renamed that crawler's `soup` parameter to `data`, incidentally fixing one. The issue's own figure of 11 undercounted by three.

The nine that remain sit in files owned by other in-flight changes (#506, #524, #527, #531, and the registry docstring change) and are recorded in `KNOWN_DEFECTS` with a reason each, under a test asserting every one **still reproduces** — so the list cannot rot into a description of bugs that are gone. That test has already earned its keep: it failed on exactly the IPX entry during the rebase, which is how the stale entry was found and removed.

## What can and cannot fail a test — both framings, honestly

**The docstring corrections cannot fail a test individually.** A label is not executed, so there is no fails-before to show for them and none is manufactured here. What pins them is `tests/test_docstring_contract.py`, which *derives* the answer from the code rather than snapshotting today's wrongness — so a docstring written next year is checked next year. Verified by reintroducing each class into a scratch tree:

```
baseline, all fixes in place          exit 0
phantom Raises: reintroduced         exit 1   "1 phantom Raises: clause"
bad Args: label reintroduced         exit 1   "1 docstring(s) name a parameter that does not exist"
Returns: re-indented into Args:      exit 1   "1 section header(s) indented into a parameter block"
```

Two of those are not merely cosmetic, and both were measured rather than argued:

- **The swallowed `Returns:` loses documentation from the built page.** Run through napoleon's `GoogleDocstring` directly, the over-indented form yields `:param Returns:` and **no `:returns:` at all**; the fixed form yields `:returns:`. `napoleon_include_private_with_doc = True` in `docs/source/conf.py`, so these private methods do publish.
- **The `Args:` mismatches are hard `TypeError`s.** None of the four functions takes `**kwargs`. `Frame.register(code=..., module=IPv4)` really does raise `TypeError: Frame.register() got an unexpected keyword argument 'module'`, while `protocol=` is accepted.

**The SACK check can and does fail before the fix.** Lengths 3, 11, 14, 17, 19 and 25 all parsed clean on `c8fd97bcd` with the SACK option present in `tcp.info['options']`; all six now raise `ProtocolError: TCP: [OptNo 5] invalid format`. Independently reproduced in a throwaway worktree at `origin/main` with only the new test file copied in: exit 1 with `SUBFAILED` for all six lengths, against exit 0 on this branch.

The two `Raises:` assertions are **complementary, not redundant**, and `_read_http_none` proves it: `any(header.flags)` in its body makes the conservative reachability check judge it possibly-raising, so only the commented-out-raise check catches it. Measured — reachability exit 0, commented-raise exit 1 with `httpv2.py:417 _read_http_none() documents 'ProtocolError' but its only raise is commented out`. Delete either test and one of the six phantoms stops being covered.

## Deliberately not done

- **`**kwargs` completeness in hip.py.** All 12 functions carrying both `**kwargs` and an `Args:` section without documenting `**kwargs` are in hip.py; 438 of 450 document it elsewhere. That set of 12 *includes* `_make_param_reg_failed` and `_make_param_route_via`, the very siblings this change copies its wording from — so fixing 4 of the 12 is what would make hip.py inconsistent. Left whole, reported rather than half-fixed.
- **RFC 2018's block-count bound.** The check implements exactly the documented rule, `(length - 2) % 8 == 0`. `length=2` is degenerate and still parses; five blocks or more cannot be expressed at all, since the 4-bit data offset caps the options area at 40 octets. Recorded in the test as a decision rather than quietly tightened past the docstring.
- **Reported, not fixed** (other streams own the files): `Args:` mismatches at `pcapkit/vendor/ipx/packet.py`, four `pcapkit/vendor/mh/*`, `pcapkit/vendor/pcapng/option_type.py`, `pcapkit/vendor/vlan/priority_level.py`, `pcapkit/foundation/registry/foundation.py`, `pcapkit/protocols/internet/ipv4.py:1212`; and the swallowed `Returns:` at `pcapkit/protocols/internet/ipv4.py:1550`.
- **#525** records a genuine separate defect found here: which exception a malformed SACK produces is process-state dependent, `ProtocolError` from `_read_mode_sack` or `FieldValueError` from `ListField.unpack`, and they are siblings so no single `except` catches both. The test asserts the union, recording the problem rather than solving it. One correction to the original write-up: a deliberate attempt to force the `FieldValueError` path by running `tests/protocols/schema/` first did **not** reproduce it, so the trigger is not pinned down and the test docstring no longer claims it is.

## Tests

```
tests/test_docstring_contract.py                          5 passed / 10 subtests   exit 0
tests/protocols/transport/test_tcp_sack_length_unit.py    3 passed / 10 subtests   exit 0
tests/test_tier_guard.py                                 25 passed / 19 subtests   exit 0
tests/protocols/transport/ + schema unit                 81 passed / 73 subtests   exit 0
```

Both new files are unit tier by the guard's own path rule, read no capture under `examples/captures/`, and need no `_tiers.py` registration. `test_docstring_contract.py` imports no pcapkit at all — it reads source under its own root — so no editable install can shadow what it measures. Exit codes were read directly throughout, since `pytest-subtests` is not a dependency and pytest 9.1.1 prints a failing subtest's parent as `PASSED`.

One commit, rebased onto `c8fd97bcd`.


JarryShaw:
Pushed an update (`6bfdebe81` → `b1124264e`, one commit, same 8 files). No behavioural change; the SACK check and every docstring correction are byte-identical. What changed is `tests/test_docstring_contract.py`'s own prose, after an independently written second scanner was run against the same baseline.

It **reproduced all three counts** (6 phantom `Raises:` / 13 `Args:` mismatches / 3 swallowed headers) and the residuals (0 / 9 / 1), including that the `Args:` figure is 13 here rather than the 14 counted against the older base. It also raised three things:

1. **A wrong rationale in the checker itself, now corrected.** `_documented_names` justified measuring indentation relative to the section header by saying Python 3.13 dedents docstrings at compile time. That is true of `__doc__` and **irrelevant here** — this module reads `ast.get_docstring(node, clean=False)`, which preserves raw source indentation. Measured on 3.14.7 for a method whose `Args:` sits at eight spaces:

   ```
   get_docstring(clean=False)  ->  indents [0, 8, 12, 12]
   get_docstring(clean=True)   ->  indents [0, 0,  4,  4]
   compiled __doc__            ->  indents [0, 0,  4,  4]
   ```

   The implementation was right, for a different reason — nesting depth moves the absolute column, so a module-level function has `Args:` at 4 and entries at 8 where a method has 8 and 12. A wrong rationale inside a checker for wrong rationales is worth correcting explicitly rather than leaving.

2. **Two latent false-negative paths, now documented with measurements.** `_raised_names` resolves only `Name` and `Attribute` raise targets (across `pcapkit/`: 856 `Name`, 2 `Attribute`, **0** `Subscript` — so currently inert), and `ast.walk` attributes a nested `def`'s `raise` to the enclosing function (11 documented functions contain a nested `def`, 3 raise inside it). Both can only *miss* a phantom, never fail a correct docstring. The second turns out to be the **correct** answer rather than a gap: `HIP._read_param_locator_set` documents `ProtocolError` and raises none itself, but calls a `_read_locator` local helper that does, once per locator — so the exception genuinely propagates and the clause is not a phantom. It would only mislead for a nested `def` that is *returned* rather than called, which this package does not do.

3. **One reported defect was a false positive**, and is now recorded as a trap rather than acted on. `Raw.__post_init__` documents `error` and `alias`, has neither in its signature, and assigns a *local* `alias` from `self._info.protocol.name` — which reads exactly like a docstring naming a parameter that does not exist. It is not: `__post_init__` forwards `**kwargs` to `unpack`, which dispatches to `read`, and `read:66-67` declares both as keyword-only parameters and uses them at `:81-82`. The local merely shares a name with the keyword. Reporting it would have asked for correct documentation to be deleted — which is exactly what the `**kwargs` exclusion exists to prevent, so it earned a place in the module docstring.

Local: contract 5 passed / 10 subtests, SACK 3 passed / 10 subtests, tier guard 25 passed / 19 subtests, all exit 0.

JarryShaw:
**✅ GOOD TO MERGE** — the one behavioral fix (SACK length validation) independently fails-before/passes-after for all 6 invalid lengths with exit codes verified directly, the swallowed-`Returns:` claim is confirmed literally by feeding both the broken and fixed docstring text through napoleon's own parser, and every re-derived count (phantom `Raises:` 6, `Args:` mismatches 13-at-base, swallowed headers 3-at-base) matches the PR's own figures exactly once its 13-vs-14 explanation (an already-merged #511 incidentally fixed one in between) is accounted for.

JarryShaw:

Reviewed at head `b1124264e913e5d8d83e6f55947c6f359ae191c1` in an isolated worktree.

**1. The one real behavioral fix — SACK length validation, independently reproduced.** `TCP._read_mode_sack` gained `if (schema.length - 2) % 8 != 0: raise ProtocolError(...)` before constructing `Data_SACK`, making the method's own docstring (`Raises: ProtocolError: If length is NOT multiply of 8 plus 2`) true for the first time. Siblings `_read_mode_sackpmt` and `_read_mode_echo` already validated their own lengths, confirming this was the outlier.
- Deleted just the 2-line check in my own worktree (reverting to `main`'s behavior) and reran the new test: **exit 1**, `6 failed, 3 passed`, explicit `AssertionError: (ProtocolError, FieldValueError) not raised` for all six invalid lengths (3, 11, 14, 17, 19, 25). Directly observed the pytest-9.1.1 mislabeling the PR warns about — the short summary printed `PASSED` for the parent test right next to six `SUBFAILED(length=N)` lines for the same test, confirming the instruction to check the real exit code rather than the summary.
- Restored the fix, confirmed `git diff --stat` empty (back to the exact PR commit), reran: exit 0, all green.

**2. Honesty of scope.** The PR states outright that the docstring corrections "cannot fail a test individually" and separately claims real fails-before coverage only for the SACK check, with its own reproduction numbers. It does not imply broader behavioral coverage than it has. Confirmed honest.

**3. Counts re-derived independently, then reconciled against the PR's own explanation.**
- **Phantom `Raises:` — 6 → 0.** 5 fixed by deleting stale clauses in `httpv2.py` (5 `_read_http_*` methods), 1 fixed by the SACK behavioral change making its existing clause true.
- **`Args:` naming a nonexistent parameter — 13 at base, 9 remaining (all in `KNOWN_DEFECTS`), 4 fixed.** 13 (not the issue's "11", and initially appears to contradict "14") — but the PR description itself explains this precisely: 14 was the true count when issue #519 was filed (the issue undercounted by 3), and 13 is the count at *this PR's actual base* `c8fd97bcd`, because the already-merged #511 incidentally fixed one `Args:` mismatch in between (IPX socket scrape's `soup`→`data` rename). Both figures are correct at their respective points in time; the PR discloses this itself rather than silently picking one. My independent count of 9 allowlisted entries + 4 diff fixes = 13 round-trips exactly.
- **Swallowed `Returns:` headers — 3 at base, 1 remaining (allowlisted), 2 fixed.** Fixed: `tcp.py:2979` and `tcp.py:3006` (dedented from inside `Args:` to top-level `Returns:`). Not fixed, and not claimed to be fixed: `ipv4.py:1550`, explicitly named in the PR's own "Reported, not fixed (other streams own the files)" section and present in `KNOWN_SWALLOWED`. Confirmed the mis-indentation genuinely exists there on `main`/this branch (file untouched by this diff).
- **Napoleon verification, literal not inferred.** Fed the actual mis-indented `ipv4.py:_make_opt_e_sec` docstring into `sphinx.ext.napoleon.GoogleDocstring` directly: renders `:param Returns: Constructured option schema.` with no `:returns:` field at all. Fed the fixed form (modeled on `tcp.py`'s post-fix layout): renders correctly as `:returns:`. Confirms the "renders as a parameter named Returns" claim literally.

**4. `tests/test_docstring_contract.py` — both properties confirmed.**
- **Import-independence:** its only imports are `ast`, `pathlib`, `unittest`, `typing`, and `tests._tiers.ROOT` (a plain path constant) — no `pcapkit` import anywhere in the chain, so it's immune to the editable-install shadowing that has corrupted other measurements in this environment.
- **Allowlist can't rot:** `test_known_defects_are_still_defects` exists and passes.
- **#527/#531 staleness check:** both confirmed still `OPEN` (unmerged) via `gh pr view`, and `c8fd97bcd` (this PR's base) is exactly `git merge-base origin/fix-519 origin/main`, confirming `main` hasn't moved past that point. Beyond that: neither #527 nor #531 touches the specific files behind any `KNOWN_DEFECTS`/`KNOWN_SWALLOWED` entry even prospectively (#527 touches package `__init__.py` files, not `foundation/registry/foundation.py` where the allowlisted docstring lives; #531 touches `vendor/__init__.py`/`vendor/mh/__init__.py`, not the individual `vendor/mh/*.py` files with allowlisted entries). No staleness risk currently, and this PR correctly did not need to touch the allowlist on their account. The one time the allowlist did rot (the `vendor/ipx/socket.py` entry, killed by the already-merged #511 during a rebase) was already caught and removed, per the PR's own account.

**5. #525 — confirmed present, not raised as a finding, per instruction.** The union assertion (`assertRaises((ProtocolError, FieldValueError))`) exists exactly as described, with the test's own docstring recording that a deliberate attempt to force the `FieldValueError` path did not reproduce it. This is a disclosed, deliberately-unresolved state-dependent issue tracked separately — not treated as a gap in this PR.

## What remains unverified
The broader `tests/test_tier_guard.py` and full `tests/protocols/transport/`+schema unit sweep numbers cited in the PR were not independently re-run (targeted tests directly falsifying the load-bearing claims were run and are clean; the broader counts are secondary and unverified by me, though I have no reason to doubt them given the targeted runs).

## Verdict
All load-bearing claims independently reproduce, including a full, honest reconciliation of the 13-vs-14 `Args:` count. Recommend merge.

Issue #491, opened 2026-09-19
IP address fields silently accept bool and corrupt the packet — unfixed sibling of #469/#481 at the shared root


JarryShaw:
Found by strand 3 of the post-wave-1 consistency sweep (`docs/source/pep.rst` — unaligned changes). Every claim below was reproduced by me on `5959a437c` before filing.

## This is the unfixed sibling of #469/#481

#481 fixed `bool` handling at **one call site**, `MH._make_opt_mn_id`, because `bool` is an `int` subclass and `True` was silently becoming `::1`. The same mechanism is still live at the **shared root**: `_IPAddressField.pre_process` in `pcapkit/corekit/fields/ipaddress.py:97-120` calls `ipaddress.ip_address(value)` with no `bool` exclusion.

`ipaddress.ip_address()` treats any `int` below `2**32` as IPv4, so on an **IPv4-typed field a `bool` is silently accepted and corrupts the packet**. On an IPv6-typed field it happens to raise, because the resulting `IPv4Address`'s version mismatches — which is why #481's fix looked sufficient. It was not; it was the IPv6 half masking the IPv4 half.

## Reproductions

**Silent corruption through the public construction API** — no exception, no warning:

```python
proto = object.__new__(IPv4)
raw = proto.make(src=True, dst=False).pack()
info = IPv4(io.BytesIO(raw), len(raw)).info
```

```
packed  -> 4500001400000000001100000000000100000000
decoded -> src=0.0.0.1  dst=0.0.0.0
```

**At the shared root**, showing the IPv4/IPv6 asymmetry that hid it:

```
IPv4AddressField.pre_process(True)  -> b'\x00\x00\x00\x01'
IPv4AddressField.pre_process(False) -> b'\x00\x00\x00\x00'
IPv6AddressField.pre_process(True)  -> raises FieldValueError (version mismatch)
```

**Outside the Field abstraction entirely** — `pcapkit/protocols/internet/esp.py:536,554` calls `ipaddress.ip_address()` directly with no version check at all:

```
SecurityAssociation(spi=1, destination=True).destination  ->  IPv4Address('0.0.0.1')
```

**`SCTP._make_param_ipv4`** (`pcapkit/protocols/transport/sctp.py:2504-2507`), `address=True` packs to `0005000800000001` — type 5, length 8, address `0.0.0.1`.

## One site the sweep flagged but did not execute — I ran it, and it corrupts too

[...]

PR #539, opened 2026-09-20
fix(corekit): reject a bool where a maker converts an address itself (#508)

### LINKED ISSUE #508: Address-typed SwitchField attributes bypass #491's bool guard: 3 mh.py makers still corrupt
Follow-up to #491 (fixed in PR #500). **The scope here is much narrower than I originally expected, and the correction is the useful part of this issue** — I had recorded roughly ten suspect maker sites, and re-measuring on `main` at `d2d9bb80c` shows the merged root guard covers almost all of them.

## What the merged guard already covers — measured, not assumed

#500 put the `bool` rejection in `_IPAddressField.pre_process` and `_IPInterfaceField.pre_process`, so every parameter routing through an address *field* is now guarded. I swept **36 address-ish maker parameters** across `mh.py`, `ipv4.py`, `sctp.py` and `hip.py` with `True` and `False`:

```
ipv4._make_opt_rr  (route=[True])  -> FieldValueError (guarded)
ipv4._make_opt_lsr (route=[True])  -> FieldValueError (guarded)
ipv4._make_opt_ssr (route=[True])  -> FieldValueError (guarded)
sctp._make_param_ipv4(address=True) -> FieldValueError (guarded)
esp  SecurityAssociation(destination=True) -> ProtocolError (guarded)
mh._make_opt_aca / _alt_ipv4_coa / _anchored / _dlif_lla / _hnp / _ipv4_ack ... -> guarded
```

**Every site I had listed as "not covered" is in fact covered**, including `SCTP._make_param_ipv4`, which I had specifically called out. That claim was wrong and is withdrawn here rather than left in a stale note.

## What genuinely still accepts a bool

Four maker parameters, all in `mh.py`:

```
mh._make_opt_bid(address=<bool>)     -> 000001000000000001
mh._make_opt_lma_up(address=<bool>)  -> 00000000000001
mh._make_opt_lmaa(address=<bool>)    -> 00010000000001
```

(`mh._make_opt_ni(home=<bool>)` also surfaced in the sweep and is **not** a defect — `NonceIndicesOption.home` is a `UInt16Field` at `pcapkit/protocols/schema/internet/mh.py:670`, a nonce *index* rather than an address. That one was a false positive of my own name-matching heuristic, and `home=True` becoming `1` in a uint16 field is correct behaviour.)

## The mechanism — one cause, not three

These parameters are not backed by `IPv4AddressField`/`IPv6AddressField`. They are backed by **`SwitchField`**, which selects the concrete field at runtime from the option length or address family:

```
mh.LMAAddressOption.address:            IPv4Address | IPv6Address | bytes   field=SwitchField
mh.LMAUserPlaneAddressOption.address:   IPv4Address | IPv6Address | bytes   field=SwitchField
mh.TargetCareofAddressSuboption.address: IPv4Address | IPv6Address | bytes  field=SwitchField
```

Because the dispatch happens inside `SwitchField`, the value never reaches `_IPAddressField.pre_process`, so #500's guard cannot fire. `BindingIdentifierOption` is the same shape — per :rfc:`5648#section-4.3` the option carries **no** address-family flag, so the family is selected from the option length alone, which is exactly why a switch is used.

## Every address-typed `SwitchField` in the schema tree

Enumerated rather than sampled — **7 attributes**, and the gap is not confined to `mh.py`:

| module | class.attribute | annotation |
|---|---|---|
| `internet.hip` | `Locator.value` | `IPv6Address \| LocatorData` |
| `internet.mh` | `DelegatedMNPOption.prefix` | `IPv4Address \| IPv6Address` |
| `internet.mh` | `LMAAddressOption.address` | `IPv4Address \| IPv6Address \| bytes` |
| `internet.mh` | `LMAUserPlaneAddressOption.address` | `IPv4Address \| IPv6Address \| bytes` |
| `internet.mh` | `MNIDOption.identifier` | `bytes \| str \| IPv6Address` |
| `internet.mh` | `TargetCareofAddressSuboption.address` | `IPv4Address \| IPv6Address \| bytes` |
| `transport.tcp` | `MPTCPAddAddress.address` | `IPv4Address \| IPv6Address` |

Two of those are already effectively handled: `MNIDOption.identifier` is where #469/#481 put a guard at the call site, and `DelegatedMNPOption.prefix` raised `ProtocolError` in my sweep. The rest are untested for this and at least three demonstrably corrupt.

## Suggested shape

The guard wants to sit where the switch resolves, not at each of the seven call sites — a call-site fix is what #481 did for one of them and is exactly how the root cause survived to become #491. Either `SwitchField` rejects a `bool` before dispatching, or each address-selecting switch routes through a wrapper that does.

`MPTCPAddAddress.address` is worth an explicit test either way: it is in `tcp.py` rather than `mh.py`, so a fix scoped to Mobility Header would miss it.

JarryShaw:
Closes #508.

## The issue's stated mechanism is wrong, and correcting it is most of this change

#508 reads the defect as a dispatch problem: because `SwitchField` picks the concrete field at runtime, a `bool` supposedly never reaches `_IPAddressField.pre_process`, so #500's guard cannot fire — which would put the fix in `SwitchField`. Measured, that is not what happens.

`SwitchField.pre_process` (`pcapkit/corekit/fields/misc.py:432-445`, a range I re-verified is still accurate) inspects only `self._field` and delegates straight to the resolved field:

```python
if self._field is None:
    return NoValue  # type: ignore[unreachable]
return self._field.pre_process(value, packet)
```

So a `bool` that genuinely arrives at an address-selecting switch is **already** rejected, on every address-typed branch of every such switch — #500 covers it. A guard in `SwitchField` would be dead code twice over, because `SwitchField.pack` delegates to `self._field.pack(...)`, which calls the resolved field's `pre_process` directly; `SwitchField.pre_process` is not even invoked on the pack path.

The real cause is upstream of the schema entirely. Seven `_make_*` methods have to know the address *family* before they can build the schema — to size an option whose length is the only thing on the wire carrying that family — so they convert the argument themselves with bare `ipaddress.ip_address()`. That conversion happens **before** the schema, so it launders `True` into a perfectly ordinary `IPv4Address('0.0.0.1')` that #500's guard can then only see as a legitimate address. Worse, the option length and family flag are then derived *from the laundered address*, so the selector picks the matching `IPv4AddressField` and no mismatch is left for anything to detect.

## What changed

- **`parse_ip_address(value, description, version=None)`** added to `pcapkit/corekit/fields/ipaddress.py`. It calls the existing `_reject_bool` as its first statement, then converts; `version` pins the family where the wire format fixes it, so an `int` widens correctly (`258` is `::102` for `version=6`, but `0.0.1.2` for family-agnostic `ip_address`). It raises `FieldValueError` — the same class `_IPAddressField.pre_process` raises for the identical value, so a caller sees one exception whether the `bool` arrived through the schema or through a maker. (`BoolError` would be wrong here: it means "must *be* a bool", and sits with `IntError`/`BytesError` in the `TypeError` block.)
- **`misc.py` is untouched**, for the reason above.
- The seven corrupting makers routed through it.
- The now-unused `import ipaddress` dropped from `tcp.py` (verified no remaining runtime use; the `TYPE_CHECKING` import is separate).
- `parse_ip_address` documented in `docs/source/pcapkit/corekit/fields/ipaddress.rst`, naming the defining module rather than a re-export.

## Re-derived counts

Both of the issue's counts are wrong, in opposite directions.

**Address-typed `SwitchField` attributes: 10, not 7.** Derived twice independently — statically, and at runtime from `Schema.__fields__`:

| module | class.attribute |
|---|---|
| `internet.hip` | `Locator.value` |
| `internet.hopopt` | `SMFIdentificationBasedDPDOption.tid` |
| `internet.ipv6_opts` | `SMFIdentificationBasedDPDOption.tid` |
| `internet.mh` | `BindingIdentifierOption.address` |
| `internet.mh` | `DelegatedMNPOption.prefix` |
| `internet.mh` | `LMAAddressOption.address` |
| `internet.mh` | `LMAUserPlaneAddressOption.address` |
| `internet.mh` | `MNIDOption.identifier` |
| `internet.mh` | `TargetCareofAddressSuboption.address` |
| `transport.tcp` | `MPTCPAddAddress.address` |

The three the issue's table misses are `hopopt`/`ipv6_opts` `SMFIdentificationBasedDPDOption.tid` and `mh` `BindingIdentifierOption.address` — all `ConditionalField`-wrapped, e.g.

```python
# pcapkit/protocols/schema/internet/mh.py:1294-1297
address: 'Optional[IPv4Address | IPv6Address | bytes]' = ConditionalField(
    SwitchField(selector=bid_address_selector),
    lambda pkt: pkt['length'] > 4,
)
```

so a grep anchored on `= SwitchField(` cannot see them. (The issue does mention `BindingIdentifierOption` in prose while omitting it from the table.) `test_the_address_typed_switch_table_is_complete` now pins this at 10 by rediscovering them from `Schema.__fields__`, so the next one added cannot go unnoticed — that "nobody listed it" failure mode is what #491 and #508 share.

**Corrupting call sites: 7 makers, not 3.** The issue lists 3, all in `mh.py`. The four it misses are `MH._make_fid_suboption`, `MH._make_opt_dmnp`, `HIP._make_param_locator_set` and `TCP._make_mptcp_addaddr`. Observed wire output before the fix:

| maker | `address=True` produced |
|---|---|
| `MH._make_opt_bid` | `23080001000000000001` |
| `MH._make_opt_lmaa` | `2906010000000001` (vs `2906010001020304` for `'1.2.3.4'`) |
| `MH._make_fid_suboption` | `0506000000000001` |
| `MH._make_opt_dmnp` | `3706801800000001` |
| `MH._make_opt_lma_up` | `3b06000000000001` |
| `HIP._make_param_locator_set` | byte-identical to `ip='::1'` |
| `TCP._make_mptcp_addaddr` | schema with `address=IPv4Address('0.0.0.1')`, `version=4` |

`_make_opt_dmnp` needs `prefix_length <= 32` to show it; the default `prefix_length=64` hides it behind the range check, which is why the issue's own sweep read that site as already guarded. `HIP._make_param_locator_set` is the most damning: `ip=True` and `ip='::1'` were literally indistinguishable on the wire. `TCP._make_mptcp_addaddr` cannot be constructed end to end at all (#541), so its corruption was only ever visible on the schema the maker returns — #508 was right that it needed its own test, and it has one.

## Fails-without evidence

`pytest-subtests` is not installed and pytest 9.1.1's native subtests **print a failing subtest's parent as `PASSED`**, so the exit code is the only honest signal. Read from a file, never through a pipe.

All three runs below were measured on this branch after it was rebased onto `691f12ab5`, rather than carried over from the pre-rebase measurement.

**Baseline, fix in place:** exit code **0** — 80 passed, 584 subtests passed.

**Proof 1 — revert all seven call sites, keep the helper:** exit code **1**. 18 failing subtests; only *one* parent produced a `FAILED` line. The other two tests (`test_switch_backed_address_makers_reject_a_bool`, `test_both_bool_guards_stay_catchable_as_value_error_and_as_base_error`) had failing subtests and were counted in "79 passed".

```
SUBFAILED(option='bid'|'lmaa'|'lma_up'|'tcoa'|'dmnp', value=True|False)   ... test_mh_unit.py
SUBFAILED(site='hip.Locator.value', value=True|False)                    ... test_fields_ipaddress.py
SUBFAILED(site='tcp.MPTCPAddAddress.address', value=True|False)          ... test_fields_ipaddress.py
18 failed, 79 passed, 567 subtests passed
```

Those seven distinct labels are the independent confirmation of the 7-maker count.

**Proof 2 — remove only `_reject_bool(value, description)` from inside `parse_ip_address`, leave all routing in place:** exit code **1**, 21 failing subtests, and **zero `FAILED` parent lines** — every parent reported as passing. This is the starkest form of the pytest hazard, and it shows the helper's guard is load-bearing rather than the routing alone.

After each revert the tree was restored and verified byte-identical to the committed one.

Every measurement was taken with `PYTHONSAFEPATH=1` and `PYTHONPATH` set to this worktree, with `pcapkit.__file__` printed and asserted to start with the worktree root first. That matters here: an editable install (`__editable___pypcapkit_1_4_1_post2_finder`) is present in the venv, and I confirmed it *appends* to `sys.meta_path` (`:76`), so `PathFinder` still wins and the measurements are of this tree rather than the main checkout.

## Coverage

`coverage run -m pytest` (no `pytest-cov`) over the three test files:

- `pcapkit/corekit/fields/ipaddress.py` — **93%**, missing lines `67, 294, 401, 439, 489, 534`. The new `parse_ip_address` spans **115–216**, so **none of its lines are missing**.
- `pcapkit/protocols/internet/mh.py` — 92%; the five changed makers all covered.
- The changed lines in `hip.py` (3081, 3086, 3091) and `tcp.py` (2907) are all covered.

`EXPECTED_FAILURES` in `tests/protocols/test_option_roundtrip_unit.py` is still **55** (imported, not grepped — it uses `**` unpacking), i.e. this change neither fixes nor breaks an option round-trip.

## Local suite

Rebased onto `691f12ab5`. The CI-equivalent unit selection (`--ignore=tests/integration --ignore-glob='*_runtime.py' --ignore-glob='*_regression.py'`) is **green: exit code 0 — 1058 passed, 8 skipped, 2602 subtests passed, zero failures**, with nothing to disclaim.

An earlier revision of this description reported one pre-existing failure on `main` — `tests/test_docstring_contract.py::DocstringParameterTests::test_known_defects_are_still_defects`, from a stale `KNOWN_DEFECTS` entry for `pcapkit/vendor/ipx/packet.py::process`. **#538 has since removed that entry**, so it is gone. Re-measured on this rebased branch: that file alone runs at exit code 0, 7 passed, 13 subtests passed.

GitHub Actions checks on this PR have not started — the queue is currently wedged rather than merely slow, so the run being absent says nothing about the change.

## Found but deliberately NOT fixed

Everything in this section is a real, verified defect that is out of scope for this PR. Each was reproduced on this branch's tree, i.e. these are *not* fixed by the change above. All are now tracked:

- **#540** — the four remaining bool-laundering make paths, filed as one issue: `ARP._make_proto_resolve` (`pcapkit/protocols/link/arp.py:400,402`, `True` → `00000001` for IPv4 and `…0001` for IPv6, against `01020304` for a real address); `IPv6_Route._make_data_type_rpl` (`pcapkit/protocols/internet/ipv6_route.py:737,745,753,764,768`, worse than a wrong address — the bool becomes a 4-byte `IPv4Address` which then corrupts the *derived compression metadata*: `dst=IPv6Address('::1'), ip=[True]` gave `cmpr_i=16 cmpr_e=3 addresses=['01']`); `IPv6_Route.make`'s `dst` (`ipv6_route.py:314`, same shape in `make` rather than a `_make_*`); and the latent `OSPF._make_id_numbers` (`pcapkit/protocols/link/ospf.py:318` — same conversion, but `grep -rn '_make_id_numbers' pcapkit/` returns only the `def`, so no live caller).
- **#541** — `TCP._make_mptcp_addaddr` cannot be constructed end to end (`KeyError: 'length'` from `pcapkit/protocols/schema/transport/tcp.py:790`). Independent of this fix and left alone; noted in a code comment at the call site.

They were left out of this PR because its file ownership does not extend to `arp.py`, `ipv6_route.py` or `ospf.py`, and because #481 fixing exactly one call site is how the root cause survived to become #491 — piecemeal is the failure mode, so the remaining sites deserve their own review rather than being tacked on here. `parse_ip_address` is now the sanctioned way to do this conversion, so #540 has somewhere to route them.


JarryShaw:
**✅ GOOD TO MERGE** — independently confirmed the issue's own proposed mechanism is wrong: on `main`, a `bool` reaching an address-typed `SwitchField` via either `.pack()` or `.pre_process()` is already rejected with `FieldValueError` (empirically reproduced), so the real defect is the seven `_make_*` call sites that convert addresses themselves *before* the schema — independently re-derived at exactly 10 address-typed switches and exactly 7 corrupting call sites, matching the PR's counts precisely.

JarryShaw:

Head sha reviewed: `6e2387ba896f4f7cf624031793ff0ce035e89699` (rebased from the original `9aac76c7c3874cb12298c3398f66b53d32a28dff` onto `main` at `691f12ab5` after #538 merged). Confirmed the rebase carries the identical patch content via `git patch-id --stable` on both diffs against their respective bases — both produce `a56f3a87252e341cee0d0a79359a4d156f781772` — so everything below, derived against the pre-rebase sha, applies unchanged to this head, and I re-ran the targeted verification directly against `6e2387ba8` as well (see "Final confirmation" below).

### The refuted mechanism — independently reproduced, not taken on the PR's word
The PR's central claim is that #508 misdiagnosed the defect: it says a `SwitchField` guard is unnecessary because `SwitchField.pre_process` already delegates to the resolved field's own `pre_process`, and `SwitchField.pack` calls `self._field.pack(...)` directly (bypassing `SwitchField.pre_process`, but not the resolved field's own guard, since `FieldBase.pack` calls `self.pre_process` internally). I read `pcapkit/corekit/fields/misc.py:432-460` and confirmed this delegation chain by inspection, then verified it empirically on `main` (`122d32795`), before this PR, with no code changes:

```python
sf = SwitchField(selector=lambda pkt: IPv4AddressField())
resolved = sf({})
resolved.pack(True, {})         # -> FieldValueError: invalid IP address: must not be a bool...
resolved.pre_process(True, {})  # -> FieldValueError: invalid IP address: must not be a bool...
```

Both raise on `main`, unmodified. This directly confirms the PR's refutation: a `SwitchField` guard proposed by #508 would be dead code, because #500's guard (inside `_IPAddressField.pre_process`) already fires on every address-typed switch. The real defect is upstream — `_make_*` methods that convert the address themselves before the schema ever sees it.

### Re-derived count: 10 address-typed `SwitchField` attributes, not 7
I did not accept the PR's count from its table. I grepped `SwitchField(` across `pcapkit/protocols/schema/` (25 hits total) and read the type annotation and any wrapping `ConditionalField` around each one by hand, rather than filtering by regex on the annotation (which is exactly what undercounts the three `ConditionalField`-wrapped sites). Independently arrived at exactly these 10:

`hip.Locator.value`, `hopopt.SMFIdentificationBasedDPDOption.tid`, `ipv6_opts.SMFIdentificationBasedDPDOption.tid`, `mh.BindingIdentifierOption.address`, `mh.DelegatedMNPOption.prefix`, `mh.LMAAddressOption.address`, `mh.LMAUserPlaneAddressOption.address`, `mh.MNIDOption.identifier`, `mh.TargetCareofAddressSuboption.address`, `tcp.MPTCPAddAddress.address`.

This matches the PR's table exactly, including the three it says a bare annotation-grep would miss (`hopopt`/`ipv6_opts` `SMFIdentificationBasedDPDOption.tid` and `mh.BindingIdentifierOption.address`, all three genuinely `ConditionalField`-wrapped — confirmed by reading `pcapkit/protocols/schema/internet/{hopopt,ipv6_opts}.py:504-508` and `mh.py:1293-1296` directly).

### Re-derived count: 7 corrupting call sites, not 3
Read the actual diff rather than trusting the PR's list: `pcapkit/protocols/internet/mh.py` changes exactly 5 makers (`_make_opt_bid`, `_make_opt_lmaa`, `_make_fid_suboption`, `_make_opt_dmnp`, `_make_opt_lma_up`), `hip.py` changes 1 (`_make_param_locator_set`), `tcp.py` changes 1 (`_make_mptcp_addaddr`) — 7 total, matching. I also checked the one address-typed switch site the PR does *not* list as corrupting, `mh.MNIDOption.identifier` — its maker `_make_opt_mn_id` (`mh.py:7769`) already has an explicit, pre-existing bool guard from #481/#469 (`Raises: ProtocolError: If identifier is a bool...`), so its exclusion from the 7 is correct, not an oversight.

### Exception type — confirmed correct, and confirmed why the PR's alternative would be wrong
`parse_ip_address` raises `FieldValueError` and calls `_reject_bool` as its first statement (`ipaddress.py:196`, verified by reading the source). I independently pulled up `pcapkit/utilities/exceptions.py`: `BoolError(BaseError, TypeError)` — docstring literally *"The argument(s) must be `bool` type"* — confirming the PR's claim that `BoolError` would be semantically backwards here (it means "should be a bool", not "must not be one"). `FieldValueError(BaseError, ValueError)` is the class actually used, matching `_IPAddressField.pre_process`'s exception for the identical value.

### `NonceIndicesOption.home` — confirmed untouched, not merely claimed so
Checked `MH._make_opt_ni` (`mh.py:7658`): `home: 'int' = 0`, not address-typed, not in the diff at all. This maker is completely outside the PR's changes, so `home=True` still coerces to `1` exactly as before — the PR does not touch this path, confirmed by the diff rather than assumed from the description.

### Falsification / fails-without proofs — both independently reproduced with exit codes read from a file
**Proof 1 (mh.py half):** reverted `pcapkit/protocols/internet/mh.py` to `main`, kept the new helper and all three test files. `PYTHONSAFEPATH=1 <repo>/.venv/bin/python -m pytest tests/protocols/internet/test_mh_unit.py -q`: **exit code 1**, `11 failed, 49 passed, 458 subtests passed`, with exactly the 10 `SUBFAILED(option=..., value=True|False)` combinations across the 5 reverted makers plus 1 parent `FAILED` line — a clean, reproducible fails-without result for the largest half of the fix.

**Proof 2 (the guard itself, not just the routing):** on the full PR head, removed only the `_reject_bool(value, description)` call from inside `parse_ip_address`, leaving every call site routed through it. `pytest tests/corekit/test_fields_ipaddress.py tests/corekit/test_fields_misc.py tests/protocols/internet/test_mh_unit.py`: **exit code 1**, `21 failed, 80 passed, 563 subtests passed`, and **zero** `^FAILED` parent lines (`grep -c "^FAILED"` → 0; `grep -c "SUBFAILED"` → 21). This is a stark, independently-reproduced demonstration of the pytest-subtests hazard this whole review programme is built around: every parent test's summary line would read as passing while the guard is completely gone. Both edits were reverted afterward; `git diff <head> --stat` empty before moving on each time.

### Coverage — independently measured, not copied from the PR
`coverage run -m pytest` (no `pytest-cov`) over the three test files: `pcapkit/corekit/fields/ipaddress.py` — **93%**, missing lines **67, 294, 401, 439, 489, 534** — matches the PR's numbers exactly, and confirmed none of those misses fall inside `parse_ip_address`'s own span (115-216).

### `EXPECTED_FAILURES` — confirmed unchanged
Imported (not grepped) `tests/protocols/test_option_roundtrip_unit.py::EXPECTED_FAILURES`: **55** entries, same as the programme's known baseline. This PR neither fixes nor breaks an option round-trip.

### `KNOWN_DEFECTS` rot check
Coordinator flagged that this PR touches `mh.py` and `tcp.py`, both of which have entries in `KNOWN_DEFECTS` (four `vendor/mh/*` entries are unrelated generated-code entries; the one live concern was whether #539 accidentally fixed any docstring defect it carries). Ran `tests/test_docstring_contract.py` on the (pre-rebase) PR head: the *only* rotted entry was the pre-existing, already-known `pcapkit/vendor/ipx/packet.py::process` one (fixed by #538, unrelated to this PR's files). No new rot from #539's own changes.

### ARP/IPv6_Route/OSPF "found but not fixed" — spot-checked, not accepted on faith
Read `pcapkit/protocols/link/arp.py:397-401` directly: `ipaddress.IPv4Address(addr)` / `IPv6Address(addr)` called bare, with no bool guard at all — confirms item 1 of the PR's "deliberately not fixed" list is a real, currently-unfixed defect and not a padding claim. Did not individually re-verify the `ipv6_route.py` and `ospf.py` items with the same depth, given time budget, but the ARP spot-check gives confidence the pattern-recognition behind the rest of that list is sound.

### Final confirmation on the rebased head
Checked out `6e2387ba896f4f7cf624031793ff0ce035e89699` directly (not inferred from patch-id alone) and re-ran `tests/test_docstring_contract.py tests/corekit/test_fields_ipaddress.py tests/corekit/test_fields_misc.py tests/protocols/internet/test_mh_unit.py`: **exit code 0**, `87 passed, 597 subtests passed`. The previously-inherited `test_known_defects_are_still_defects` failure is gone on this head, since it now sits on top of `main` post-#538-merge.

### CI status
Not run. GitHub Actions is backed up (35 queued, 3 pending, 0 in-progress at last check) with no declared incident; verdict is on local evidence only, per standing instruction not to wait for CI.

## What remains unverified
- The `ipv6_route.py` and `ospf.py` items in the "deliberately NOT fixed" list were not independently re-derived with the same rigor as the ARP spot-check.
- CI has not run on this head at all.

PR #551, opened 2026-09-20
perf: cache `struct.calcsize` in FieldBase length property to eliminate per-field parsing overhead

Ts-Boom:
## Summary

This PR optimizes the `length` property in `FieldBase` by caching the result of `struct.calcsize(self.template)`. By implementing a simple invalidation cache based on the `template` string, we eliminate redundant `struct.calcsize()` executions during high-throughput packet unpacking.

---

## Problem

In `pcapkit/corekit/fields/field.py`, the `length` property was dynamically computing the struct size on every access:

```python
# Before
@property
def length(self) -> 'int':
    """Field size."""
    return struct.calcsize(self.template)
```

The `unpack` method in the same class contains a comment explicitly acknowledging this bottleneck:
```python
# NOTE: ``length`` recomputes struct.calcsize() on every read, so the
# three reads this method used to make were three calcsize() calls for
# one value.
length = self.length
```

Because `FieldBase.unpack()` is called for *every single field* of *every single packet*, this resulted in millions of unnecessary `struct.calcsize()` calls in large PCAP files. While CPython's `struct.calcsize` does have an internal LRU cache, the Python-to-C function call overhead, dictionary lookups, and parameter parsing still consume measurable CPU cycles in tight loops.

---

## Fix

We introduced a lightweight caching mechanism directly in the `length` property:

```python
# After
@property
def length(self) -> 'int':
    """Field size."""
    # Cache calcsize to avoid massive CPU overhead during unpacking
    if getattr(self, '_length_cache_key', None) != self.template:
        self._length_cache = struct.calcsize(self.template)
        self._length_cache_key = self.template
    return self._length_cache
```

This ensures `struct.calcsize()` is executed exactly once per unique template string per field instance.

### Why not `@functools.cached_property`?
We avoided `@functools.cached_property` (introduced in Python 3.8) to maintain backward compatibility with Python 3.6/3.7, which PyPCAPKit supports. Additionally, some subclasses (like `NumberField` and `misc.py` fields) mutate `self._template` during their `pre_process` or `__call__` lifecycles. A static cached property would return stale lengths if the template changes. Our manual cache implementation inherently handles template invalidation safely.

---

## Performance Rationale

| Metric | Before | After |
|--------|--------|-------|
| `calcsize` executions | 1 per field unpacked (millions per large PCAP) | 1 per unique template per field instance |
| Property access overhead | High (Python function call + C extension call + internal cache lookup) | Minimal (attribute check + return) |
| Correctness | Guaranteed | Guaranteed — automatically invalidates if `_template` is modified |

By bypassing the C-extension overhead on repeated accesses, this optimization provides an immediate reduction in total CPU cycles required to ingest heavy network traffic files.


JarryShaw:
Thanks for this. I measured it, and unfortunately it makes `length` slower rather than faster.

`struct.calcsize` is already memoised inside CPython's `struct` module — it looks up a compiled `Struct` in an internal cache, so it is a dict lookup rather than a format-string parse. The added `getattr(..., None)` plus string comparison costs more than the call it avoids.

Measured on this branch, Python 3.14.7, 200,000 iterations per field:

| field | template | current | with cache |
|---|---|---|---|
| `UInt8Field` | `>B` | 68 ns | 87 ns |
| `UInt32Field` | `>I` | 68 ns | 89 ns |
| `BytesField(64)` | `64s` | 67 ns | 87 ns |

Consistent ~28% regression, not a win.

One thing the patch does get right, worth saying: keying the cache on `self.template` rather than caching unconditionally is the correct shape, because `_template` genuinely does mutate at runtime — `strings.py:64,93,150`, `numbers.py:108,167`, `collections.py:87`. A naive cache there would have returned stale lengths. So if a cache were wanted, this is how it would have to be written.

If you are chasing per-field unpacking overhead, `length` is probably not where it lives — `calcsize` was already cheap. Profiling a real `extract()` run would point at the actual hot spot.

Ts-Boom:
Thanks for the detailed benchmark and explanation! That makes total sense—I hadn't factored in CPython's internal struct caching overhead outweighing the extra attribute lookups for simple templates. I'll close this PR as suggested. Thanks again for your time and guidance! @JarryShaw

Issue #541, opened 2026-09-20
TCP._make_mptcp_addaddr cannot pack: kind/length rejected as fields, then the port predicate KeyErrors on length


JarryShaw:
`TCP._make_mptcp_addaddr` cannot build an `ADD_ADDR` option end to end. Found while fixing #508; unrelated to that defect, so filed separately.

Measured on `main` at `122d32795`, repo venv, `PYTHONSAFEPATH=1`, with `pcapkit.__file__` asserted to be the repository tree:

```
>>> t._make_mptcp_addaddr(MPTCPOption.ADD_ADDR, addr_id=1, addr='1.2.3.4')
UnknownFieldWarning: 'kind' is not a valid field name
UnknownFieldWarning: 'length' is not a valid field name
schema: addr=1.2.3.4 port=None
>>> .pack()
KeyError: 'length'
```

Same result with `port=443`. The schema object constructs; only `pack()` fails.

### The chain

`TCP._make_mptcp_addaddr` at `pcapkit/protocols/transport/tcp.py:2898-2901` passes `kind` and `length` to `Schema_MPTCPAddAddress`:

```python
return Schema_MPTCPAddAddress(
    kind=cast('Enum_Option', Enum_Option.Multipath_TCP),
    length=4 + (4 if version == 4 else 16) + (2 if port is not None else 0),
    ...
```

Both are rejected with `UnknownFieldWarning`, so neither lands in the packet mapping. Then `pack()` evaluates the `port` field's condition at `pcapkit/protocols/schema/transport/tcp.py:792-794`:

```python
port: 'int' = ConditionalField(
    UInt16Field(),
    lambda pkt: pkt['length'] in (10, 22),
)
```

`pkt['length']` is not there, so `KeyError: 'length'`.

Two things are wrong and it is worth deciding which is the root cause rather than patching the symptom:

1. The maker computes a correct `length` and the schema will not accept it. If `kind`/`length` are meant to be declared fields on this schema (siblings appear to carry them), the declaration is missing. If they are meant to be supplied by an enclosing option writer, the maker should not be passing them.
2. The `port` predicate depends on `length`, a field derived from whether `port` itself is present — a circularity that only works if `length` is materialised before the conditional is evaluated.

[...]

Issue #554, opened 2026-09-20
Unbounded allocation in FieldBase.unpack: a wire-declared length drives rjust() with no bound


JarryShaw:
`pcapkit/corekit/fields/field.py:240-244`:

```python
length = self.length

if not isinstance(buffer, bytes):
    buffer = buffer.read(length)
value = struct.unpack(self.template, buffer[:length].rjust(length, b'\x00'))[0]
```

`length` derives from the field's configured length, which for several schemas is a callable resolved against the packet under parse — so it can be attacker-influenced by a length written on the wire. `rjust(length, b'\x00')` then allocates `length` bytes regardless of how much data actually arrived.

The consequence is that a small crafted input can force a very large allocation. A roughly 40-octet PCAP-NG Decryption Secrets Block is enough to drive a multi-gigabyte `rjust`, because the block's own declared length feeds the field length while the file supplies almost no bytes.

This is the only memory-safety-shaped finding from a sweep of deferred work across this release's 68 merged PRs, and it was never raised in review.

### Why it is worth a real fix rather than a cap

`pcapkit` parses untrusted capture files by design — that is the whole use case — so "the input is hostile" is the normal condition rather than an edge case. A plain upper bound would stop the allocation but silently truncate legitimate large fields; better is to refuse when the declared length exceeds the bytes actually available, which is a genuine format error and already has an in-library exception for it.

Note only in-library exceptions from `pcapkit.utilities.exceptions` should be used, and `FieldValueError` is the existing precedent for an impossible field value.

### Coverage

A test that feeds a short buffer declaring a huge length and asserts a bounded in-library exception rather than an allocation. It must not actually attempt the allocation when run against the fixed code, and should be proven to fail — by raising `MemoryError` or hanging under a timeout — against the current code.

JarryShaw:
**Fixed, in two stages, and ready to close.**

- `4c0bcf9b7` (#569) added the per-field ceiling: `_MAX_ZERO_PAD_LENGTH = 0x40_000` in `pcapkit/corekit/fields/field.py`, bounding what a single wire-declared length can drive `rjust()` to synthesise.
- `13fd1860c` (#593) added the cumulative bound the per-field ceiling could not provide, since a packet holds many fields and their sum was unbounded: `_MAX_ZERO_PAD_SHORTFALL = 0x10_000` plus a budgeted band above it. Measured, the 200-block scenario went from 10,922x amplification to **54.6x**.

[...]

PR #558, opened 2026-09-20
fix(schema): read the MPTCP option length from bit 8, not bit 1 (#553)

### LINKED ISSUE #553: MPTCP option header decodes length from bit offset 1 instead of 8, reading MP_CAPABLE as 60 not 12
`pcapkit/protocols/schema/transport/tcp.py:576` declares the generic MPTCP option header as:

```python
test: 'MPTCPSubtypeTest' = ForwardMatchField(BitField(length=3, namespace={
    'length': (1, 8),
    'subtype': (16, 4),
}))
```

A TCP option header is `kind` (8 bits), `length` (8 bits), then the MPTCP `subtype` (4 bits) and flags. So `length` sits at **bit offset 8**, and the entry should be `(8, 8)`. At `(1, 8)` it reads eight bits starting one bit in, straddling `kind` and `length` — for MP_CAPABLE it decodes the length as **60 instead of 12**.

This is currently masked by an unrelated `eool=0` check further along, which is why no test catches it. Same bug class as the SMF_DPD bit offset fixed in #432.

Found by a sweep of deferred work across this release's 68 merged PRs; the offset was never flagged in review.

### Related, and worth fixing in the same pass

**#541 is narrower than its title suggests.** It covers `TCP._make_mptcp_addaddr`, one of **eight** MPTCP makers. `EXPECTED_FAILURES` in `tests/protocols/test_option_roundtrip_unit.py` holds 8 `tcp-mptcp/*` entries sharing the same `kind=`/`length=` rejection, and #430's review had already established the family-wide scope five days before #541 asked whether other `_make_mptcp_*` helpers did the same. Either widen #541 or fix the family here and close it.

### Coverage

A test that decodes a real MP_CAPABLE option header and asserts `length == 12`. Prove it fails with `(1, 8)` and passes with `(8, 8)`, reading the exit code from a file — `pytest-subtests` is not installed and pytest 9.1.1 prints a failing subtest's parent as `PASSED`.

JarryShaw:
Closes #553.

## The defect

`pcapkit/protocols/schema/transport/tcp.py` declared `_MPTCP`'s option-header
forward match as:

```python
test: 'MPTCPSubtypeTest' = ForwardMatchField(BitField(length=3, namespace={
    'length': (1, 8),
    'subtype': (16, 4),
}))
```

`BitField`'s namespace entries are `(start_bit, bit_width)`, sliced out of one
big-endian bit string over the whole 3-octet window. `length` read eight bits
starting **one** bit in, straddling the low seven bits of `kind` and the high bit
of `length`.

## The offset arithmetic, checked against RFC 8684

RFC 8684 section 3 lays every Multipath TCP option out as `Kind` (8 bits),
`Length` (8 bits), `Subtype` (4 bits), then subtype-specific content. Numbering
from the option's first octet that puts `Length` at bits 8-15.

Three things confirm the window really is anchored at `kind`, rather than after
the option header:

1. The sibling entry `'subtype': (16, 4)` is already written against that
   numbering, and decodes correctly today. Two entries in one namespace cannot be
   anchored differently — `BitField.post_process` slices both from the same bit
   string — so `subtype` at bit 16 fixes the origin at `kind` and leaves bit 8 as
   the only place `length` can begin.
2. `_MPTCP` is registered as the schema for the option itself
   (`Option.register(Enum_Option.Multipath_TCP, _MPTCP)`) and, alone among the
   option schemas, does **not** inherit `Option` — so it is handed the option's
   first octet and nothing has consumed `kind`/`length` before it.
3. Measured, on the 12-octet MP_CAPABLE of RFC 8684 figure 4, whose first three
   octets are `1e 0c 01` → bits `000111100000110000000001`:

   | slice | value |
   |---|---|
   | `[1:9]` — old `(1, 8)` | **60** |
   | `[8:16]` — new `(8, 8)` | **12** |
   | `[16:20]` — `subtype` | `0` = MP_CAPABLE |

So `(8, 8)`, and the issue's reported "60 instead of 12" reproduces exactly.

## Why nothing caught it

The decoded length is used for one thing — `SchemaField(length=pkt['test']['length'])`
in `mptcp_data_selector`, sizing the nested subtype schema. Over-declaring that
size fails silently: the enclosing `OptionField` stops at the end-of-option-list
marker or the declared options length, so the over-read is absorbed rather than
reported.

## Coverage

New `tests/protocols/transport/test_tcp_mptcp_length_unit.py`, five cases read off
the schema's own declared field (`_MPTCP.__fields__['test']`) rather than a copy,
so a regression in the shipped declaration is what fails. Includes the exhaustive
case: all 256 lengths a single octet can declare must decode back to themselves.

Proven to fail without the fix — pytest exit code read from a file, since a
wrapper's exit code is not pytest's:

```
# with 'length': (1, 8)
PYTEST_EXIT=1
3 failed, 2 passed
AssertionError: 60 != 12
# with 'length': (8, 8)
PYTEST_EXIT=0
5 passed
```

The two cases that pass either way are the `subtype` and anchor invariants, which
are correct before and after — they exist to rule out "fixing" the length by
moving the window instead of the entry.

Measured against `/local/home/jarryx/GitHub/PyPCAPKit/.venv/bin/python` with
`PYTHONSAFEPATH=1` and `pcapkit.__file__` asserted into this worktree.

## Scope: #541 and the other seven MPTCP makers

#553 asks whether the eight `tcp-mptcp/*` entries in `EXPECTED_FAILURES` share
#541's root cause, so that #541 could close here. **They share one structural
cause but need three different fixes, and one of them is unrelated** — so this PR
deliberately does not attempt the family, and #541 should stay open. Measured, by
calling each of the eight makers and reading the exception:

The structural cause: every MPTCP subtype schema inherits `MPTCP`
(`schema/transport/tcp.py:608`), not `Option`, and `MPTCP` declares `kind`,
`length` and `subtype` **only under `if TYPE_CHECKING:`** (lines 613-619) — pure
annotations, no field descriptors. Their per-class `TYPE_CHECKING __init__` stubs
nonetheless advertise `kind`/`length`, which is what leads every
`_make_mptcp_*` helper to pass them; both are then dropped with
`UnknownFieldWarning`.

It surfaces three ways:

| Group | Subtypes | Failure |
|---|---|---|
| A | MP_CAPABLE, ADD_ADDR (#541), REMOVE_ADDR, MP_PRIO | `KeyError: 'length'` in `pack()`, from each schema's own `pkt['length']` predicate |
| B | DSS, MP_FAIL, MP_FASTCLOSE | `AttributeError: no attribute 'kind'`, deferred to `_read_tcp_options` — `pack()` *succeeds* |
| C | MP_JOIN | `AttributeError: 'TCP' object has no attribute '_flags'` — never reaches schema construction at all |

Group C is a genuinely separate defect: `_make_mptcp_join` branches on
`self._flags`, which only exists while parsing. It masks whether the kind/length
cause even applies to MP_JOIN.

Fixing the family means giving the subtype schemas real `kind`/`length` fields (or
adding a construct-path wrapper mirroring `_MPTCP`), which changes ten schema
classes and both directions of the MPTCP path — a much larger change than a bit
offset, and one that would swamp this diff. Worth noting it is needed for parsing
too, not just construction: a real 12-octet MP_CAPABLE does **not** parse on
`main` either, raising `KeyError: 'length'` from `MPTCPCapable.rkey`'s
`pkt['length'] != 32` predicate, independently of which bit offset the header
used. That is why this PR's assertion is on the decoded namespace rather than on
an end-to-end parse.

`EXPECTED_FAILURES` is **untouched** — all eight entries still fail in the
recorded way, and `tests/protocols/test_option_roundtrip_unit.py` passes
unchanged.

## Verification

- Unit tier (`--ignore=tests/integration --ignore-glob='*_runtime.py' --ignore-glob='*_regression.py'`), the selection CI runs: **1104 passed, 8 skipped**, identical to the pre-change baseline on `main`.
- The only failures anywhere are in the fixture-dependent tier, which needs `make samples` and fails identically on `main` (134 failed / 1165 passed both before and after).


JarryShaw:
**✅ GOOD TO MERGE** — independently derived the correct bit offset from RFC 8684 section 3's own layout (Kind at bits 0-7, Length at bits 8-15, Subtype at bits 16-19) and hand-decoded the worked example's raw bytes (`1e 0c 01`) bit-by-bit, confirming both the old offset's wrong answer (60) and the new offset's right answer (12) exactly as claimed. Confirmed the new test genuinely exercises the wire format via `.unpack()` on real bytes rather than a self-consistent round-trip, reproduced the fails-without proof exactly (3 failed/exit 1 reverted, 5 passed/exit 0 fixed), and additionally falsified with a plausible off-by-one wrong fix (`(9, 8)`) to confirm the test suite is genuinely sensitive rather than coincidentally satisfied. Confirmed the eight `tcp-mptcp/*` `EXPECTED_FAILURES` entries and `tests/protocols/test_option_roundtrip_unit.py` are correctly left untouched — nothing in the current suite exercises the buggy decode path on a currently-passing route, since every other MPTCP-touching test constructs its packet dict with a hand-supplied `length=` value rather than decoding real bytes through the fixed field.

JarryShaw:

Head sha reviewed: `6c9eab6c61d00dcc542273a8ee6f6bd9fe6f3d85`.

### Is `(8, 8)` actually right, or merely different from `(1, 8)`?
Derived the offset from RFC 8684 section 3 myself, independent of the PR's own citation: every Multipath TCP option is laid out as `Kind` (8 bits), `Length` (8 bits), `Subtype` (4 bits) plus subtype-specific content, numbered from the option's own first octet. That puts `Length` at bits 8-15, matching the fix.

Then hand-decoded the actual worked example rather than trusting the claimed numbers: RFC 8684 figure 4's 12-octet MP_CAPABLE begins `1e 0c 01`, i.e. the 24-bit string `000111100000110000000001`. Slicing it myself:

- Old `[1:9]` (8 bits from position 1): `00111100` = **60**.
- New `[8:16]` (8 bits from position 8): `00001100` = **12**.
- `subtype` `[16:20]`: `0000` = 0 = MP_CAPABLE.

Both match the PR's claim exactly, and independently confirm the fix is *correct*, not merely *different*. I also checked the "eleven extra lines" the diff ratio implied might be padding: they are entirely a source-code comment on the `test` field explaining the anchor and the RFC citation, in the same style used elsewhere in this file — not hidden logic, not padding.

### Does the fix change decoded length for other MPTCP subtypes, and did anything depend on the old value?
The same `test` field fronts every subtype, so the fix changes the decoded length uniformly across all of them. Checked whether anything *currently passing* depends on the old (wrong) value: `mptcp_data_selector` (`pcapkit/protocols/schema/transport/tcp.py:188-213`) is the only consumer of the decoded length, using it solely to size the nested `SchemaField`. Grepped every other test file that touches MPTCP (`test_fields_ipaddress.py`, `test_fields_misc.py`, `test_tcp_udp_unit.py`, `test_protocols.py`) and found that every one of them constructs its packet dict with a **hand-supplied** `'length': 12` (or similar) rather than decoding real wire bytes through `_MPTCP`'s `BitField` -- so none of them exercises the buggy offset at all, before or after this fix. Combined with the PR's own Group A/B/C analysis (independently spot-checked: `test_option_roundtrip_unit.py`'s 8 `tcp-mptcp/*` entries all fail before `pack()` ever reaches the length-consuming code), there is no currently-passing route that depended on the wrong offset, and none that could regress.

### Does the test exercise the wire format, or just round-trip?
Read `tests/protocols/transport/test_tcp_mptcp_length_unit.py` in full. `decode_header()` calls `field({}).unpack(option, {})` directly on real bytes -- genuine wire-format decoding, not a construct-then-repack round-trip that a consistently-wrong offset could satisfy in both directions. The field itself is read off the shipped schema (`_MPTCP.__fields__['test']`), not copied, so a future regression in the real declaration is what the test would catch. The exhaustive 256-case sweep specifically targets the failure mode the coordinator was worried about: with `kind` fixed at `0x1e` (binary `00011110`), any wrong offset that happens to give the right answer for one or two hand-picked lengths would still have to survive all 256 -- which is exactly the coincidence the old bug exploited on the two RFC-figure examples this file's own docstring names.

**Fails-without, reproduced exactly, exit code read from a file:**
- Reverted `pcapkit/protocols/schema/transport/tcp.py` to `main`, ran the new test file: **exit 1, 3 failed, 2 passed**, `AssertionError: 60 != 12` -- matches the PR's claimed table precisely.
- Restored, clean run: **exit 0, 5 passed**.

**Additional falsification beyond what the PR itself did:** tried a plausible off-by-one wrong fix, `'length': (9, 8)` instead of `(8, 8)` -- a mistake someone might make by miscounting from 0 vs 1. Result: **exit 1, 3 failed**, `AssertionError: 24 != 12`. The test suite is genuinely sensitive to more than just the one specific bug being fixed, not merely tuned to flip on this exact change. Reverted; `git diff <head> --stat` empty before moving on.

### Did it stay inside its lane?
File list confirms only `pcapkit/protocols/schema/transport/tcp.py` and the new test file were touched -- `tests/protocols/test_option_roundtrip_unit.py` (owned by nobody) is untouched, confirmed by `git diff origin/main 6c9eab6c6 -- tests/protocols/test_option_roundtrip_unit.py` producing no output. The PR does not stay silent about whether it *should* have affected an `EXPECTED_FAILURES` entry -- it explicitly states "`EXPECTED_FAILURES` is untouched -- all eight entries still fail in the recorded way," which I independently confirmed by running `tests/protocols/test_option_roundtrip_unit.py` on the PR head (6 passed, 358 subtests, exit 0) and by importing `EXPECTED_FAILURES` directly to count the `tcp-mptcp/*` keys (8, matching the PR's table exactly: ADD_ADDR, DSS, MP_CAPABLE, MP_FAIL, MP_FASTCLOSE, MP_JOIN, MP_PRIO, REMOVE_ADDR).

Also confirmed #554 was not folded into this PR -- the diff is exactly the 2 files stated (`pcapkit/protocols/schema/transport/tcp.py`, the new test), nothing touching whatever #554's scope covers.

### CI status
Per the coordinator, CI on this head is running (4 SUCCESS/0 failing/2 SKIPPED/17 running of 23 at last check) and not yet finished; not waited on, per standing instruction. Re-tallying a CI number is not attempted here beyond what the coordinator already reported, per the recent correction about `__typename`-filtering CheckRuns vs. StatusContexts.

## What remains unverified
- The full unit-tier suite (`1104 passed, 8 skipped` claimed) is running in the background as I post this and had not completed; I will post the exact number as a follow-up once it finishes, having learned from an earlier mistake in this session not to touch the worktree checkout while a background test run depends on it.
- The Group A/B/C analysis of the other seven MPTCP makers (why #541 should stay open) was read and spot-checked against `mptcp_data_selector` and the roundtrip test's pass/fail state, but I did not independently call each of the eight `_make_mptcp_*` helpers myself to re-derive the exact exception type table.

PR #571, opened 2026-09-21
corekit: reject short dynamic field buffers

### LINKED ISSUE #554: Unbounded allocation in FieldBase.unpack: a wire-declared length drives rjust() with no bound

lux-liang:
Fixes #554.

## Summary

- reject short buffers when a field's length is resolved from packet context
- preserve fixed-width short-read padding used by higher-level parser diagnostics
- cover both byte buffers and streams, plus the existing compatibility behavior

## Why

`FieldBase.unpack()` padded every short read to the declared field length. For a
callable length derived from packet data, that could create bytes that never
arrived. The parser now raises the existing `FieldValueError` before padding a
short dynamically sized field.

## Verification

- regression-first check on the original code: 2 failed, 1 passed
- focused tests: 4 passed
- corekit and schema tests: 128 passed, 125 subtests passed
- unit tier: 1014 passed, 114 skipped, 2588 subtests passed
- `python -m compileall` on both changed files
- `git diff --check`


JarryShaw:
Thanks for this, and for spotting that a naive `len(buffer) < length` guard would break the library — you're right that fixed-width short reads have to keep padding, and building the discriminator around that is the correct instinct. Unfortunately the specific discriminator doesn't hold, and I don't think it can be made to.

### The blocker: a truncated option stops parsing

Discriminating on whether the length came from packet context doesn't separate the attack from the legitimate short read, because **both are wire-derived**. A capture cut short by the snapshot length declares an over-long option length in exactly the same way a hostile one does.

`OptionField.unpack` rewinds and calls `schema.unpack(...)` at `collections.py:420`, which unpacks the option's *data* field — and that field is callable-length: `BytesField(length=lambda pkt: pkt['length'] - 2)` at `schema/transport/tcp.py:321` and `schema/internet/ipv4.py:259`. So the guard fires on it.

A TCP segment with data offset 7, unassigned option kind `0x4f` declaring length 12 with 6 data octets present — on `main`:

```
PARSED ok; options= [(<Option.Reserved_79: 79>, 12)]
SchemaWarning: packet length < 0: -4
```

That warning is the library noting the shortfall and carrying on, which is what #431 established. On this branch the same segment raises `FieldValueError: Field data requires 10 octets, but only 6 are available.`, escaping `TCP(...)` unwrapped. Same at declared length 32.

The full suite stays green either way, which is why this wasn't obvious — that path has no test at all. I've filed that gap as #572, independently of this PR.

### What's right

Your exception choice is correct: `FieldValueError` is the house pick for length mismatches in `corekit/fields/`. And your message spelling `{self.name}` is better than what I had — plain is 8/8 across `pcapkit/corekit/`, including five lines away in the same file. I'm taking that.

### Where this leaves it

I'm going with #569, which bounds on magnitude instead: it rejects only when a declared length exceeds both the buffer *and* 262,144 octets (libpcap's `MAXIMUM_SNAPLEN`, and this package's own `snaplen` default), so every short read stays under the ceiling and keeps padding. So I'm closing this one — but the reasoning in it was sound and the short-read constraint you identified is exactly the thing that makes this fix difficult.

Two smaller notes for next time: tests for a `field.py` change belong in `tests/corekit/test_fields_field.py` rather than `test_fields_strings.py`, and `CHANGELOG.md` is generated from `docs/source/changelog/1.5.0.rst` by `util/changelog_md.py`. Also, your run showing 114 skipped suggests the sample captures weren't generated — `python examples/generators/make_samples.py` first, then it's 17.

Please do send more.

lux-liang:
Thanks for the detailed analysis. I agree that snapshot truncation and hostile lengths are indistinguishable at that point in the wire-derived input, so the discriminator in #571 cannot be made sound without breaking the established short-read padding behavior.

I’ll leave the bounded-magnitude approach in #569 to address the original issue, and I’ll keep an eye on #572 for the missing truncated-option coverage. I’ve also noted the test-file and generated-changelog guidance for future field changes. Thanks again for the careful review — and I’ll send more focused fixes.

JarryShaw:
Following up now that #554 is actually fixed: **the defect you reported is closed, by a different mechanism than this PR proposed.**

Two commits did it:

- `4c0bcf9b7` (#569) — the per-field ceiling, `_MAX_ZERO_PAD_LENGTH = 0x40_000` in `pcapkit/corekit/fields/field.py`.
- `13fd1860c` (#593) — the cumulative bound, `_MAX_ZERO_PAD_SHORTFALL = 0x10_000` plus a budgeted band above it, because a per-field ceiling does not bound the *sum* and a packet holds many fields. That took a measured 200-block amplification from 10,922x to **54.6x**.

**Your instinct was right and your specific guard was not, and it is worth being precise about the difference.** A `len(buffer) < length` rejection cannot be used, because at the point of the read a snapshot-truncated capture and a hostile declared length are indistinguishable, and the library deliberately accommodates short reads (#431). You had already accepted that when we discussed it.

What we did not know then, and now do, is that the obvious next design — a single running budget over all padding — **also fails, and for a subtler reason.** Measured: with one budget, the same legitimate 54-octet frame declaring an IPv4 total length of 65,535 parsed to **four different results across 40 byte-identical calls**, refusing on calls 26, 33 and 39 and differing only by position in the run, with `beholder` swallowing the refusal into a silently different `Raw` layer. A guard whose answer depends on parse history is worse than the amplification it prevents. That is why the shipped fix has two bands rather than one.

The part your PR was reaching for that remains genuinely open is now **#594**: a 16-bit shortfall band is still padded unconditionally, so a crafted capture amplifies 1,637x — and a legitimately truncated offload frame amplifies by *the same* ratio, 1637.375x versus 1637.393x, so nothing at the field layer can tell them apart. Separating them needs the frame's own `incl_len`/`orig_len` from the PCAP-NG layer. If that problem interests you, #594 has the measurements and the constraints.

So this PR can be closed as superseded. Thank you for raising it — it is the reason the family got looked at properly, and four issues came out of it.

PR #630, opened 2026-09-22
fix(license): start the copyright term at 2017, and drop the end year

JarryShaw:
`LICENSE:3` read `Copyright (c) 2018-2026, Jarry Shaw`. It now reads `Copyright (c) 2017, Jarry Shaw`. That is two changes in one line, and both were asked for by the repository owner — there is no open issue for this, so there is no `Fixes` line to give.

## The start year is 2017, not 2018

Established from the repository's own history rather than from memory:

```console
$ git log --reverse --format='%ad %h %s' --date=format:'%Y-%m-%d' origin/main | head -1
2017-11-07 c57f7d0b7 Initial commit
```

`c57f7d0b7` is genuinely the only root of `main` (`git rev-list --max-parents=0 origin/main` returns it alone), so the work the notice covers began on **2017-11-07** and `2018` understated the term by a year.

**That start year has been wrong since the moment it was first written.** The file's history:

| commit | date | line 3 |
|---|---|---|
| — | 2017-11 → 2018 | stock GPL text, no author notice |
| — | 2018 → 2020 | stock MPL 2.0 text, no author notice |
| `bc836cfa2` | 2020-05-31 | BSD-3-Clause text arrives: `Copyright (c) 2018-2020` |
| `4f07f4ee4` | 2022-05-31 | `2018-2022` |
| `f1950035a` | 2023-05-08 | `2018-2023` |
| `15189abff` (#615) | 2026-09-21 | `2018-2026` |

So `2018` is the first start year the project ever asserted, and it was already a year late on the day it was written. The three hand bumps that followed each corrected the *end* year and copied `2018` forward untouched — #615, the most recent, edited this very line, got the end right, and left the start wrong.

It also disagreed with the only other copyright site in the tree. `docs/source/conf.py:51` reads

```python
copyright = f'2017-{datetime.date.today().year}, Jarry Shaw'  # pylint: disable=redefined-builtin
```

so the rendered docs footer has been saying 2017 for as long as it has existed while `LICENSE` said 2018. After this change the two agree on where the term starts, which resolves the discrepancy.

## The end year is dropped rather than automated

A range in this file is maintenance that buys nothing, for three reasons that all point the same way:

- **It ships inside a published artifact.** `LICENSE` is included in the sdist and the wheel, so its text is fixed at build time in every copy anyone has already downloaded. It cannot be computed the way `docs/source/conf.py:51` computes its own footer, because there is nothing evaluating it at read time.
- **Copyright subsists from creation regardless of the notice.** Under Berne, protection does not depend on a notice at all, let alone on the notice naming the current year. A stale end year costs nothing legally; maintaining it buys nothing legally.
- **BSD-3-Clause's canonical form carries a single year.** The upstream template is `Copyright (c) <year> <owner>`, singular.

Against that, the cost is real and measured: six years of hand maintenance on one line, three separate edits, every one of them late, and every one of them a chance to get the start year wrong — which is exactly what happened, three times running.

So **no workflow, no script hook, and no other automation is added for this.** That option was considered explicitly and rejected: automating a value that does not need to be current is strictly worse than not having the value.

`docs/source/conf.py:51` is deliberately **not touched**. It is already self-maintaining and already starts at 2017, and a docs footer showing a range is conventional and correct. Only the static legal text loses its range.

## Nothing else in the tree needed changing

- `git grep -nEi "copyright.*20[0-9]{2}"` over the tree, excluding `CHANGELOG.md` and `docs/source/changelog/` (historical prose, which should keep saying what it said at the time), returns exactly two hits: `LICENSE:3` and `docs/source/conf.py:51`. Two broader sweeps — every case-insensitive `copyright` mention, and every `20xx-20xx` range anywhere — turn up no third site, and sweep two confirms no year range now remains outside changelog prose.
- `CITATION.cff` has no `copyright` field and no year field. Its only date-shaped value is `date-released: '2026-09-20'`, which the file's own header describes as the newest *published release* rather than anything to do with the copyright term. Unaffected.

## The licence body is untouched

Checked in both directions, because a licence file is the wrong place to be casual:

- The whole-file diff is `1 insertion(+), 1 deletion(-)` in a single hunk at line 3. No other line differs.
- The remaining 28 lines were compared word-for-word against the canonical BSD-3-Clause text and are identical.
- Three things distinguish this file from SPDX's bare `licenseText`, and **all three predate this change**: the `BSD 3-Clause License` title line, which the opensource.org/choosealicense template carries and SPDX omits; the three conditions being `*` bullets rather than `1.`/`2.`/`3.`; and the extra `All rights reserved.` line. Each is present unchanged at `ead73b204`.
- `LICENSE:29` still ends `... POSSIBILITY OF SUCH DAMAGE.` — the stray `DAMAGE.s` that #615 removed has not come back.
- No CRLF, no non-ASCII, trailing newline intact, file mode unchanged (`100755`, pre-existing).

## Changelog

A `**Fixed**` bullet in `docs/source/changelog/1.5.0.rst`, with `CHANGELOG.md` regenerated by `util/changelog_md.py` and `--check` confirmed to exit 0. `CHANGELOG.md` was not hand-edited.

No tests were run: this changes one line of legal prose and one changelog entry, and touches nothing under `pcapkit/`.

## Note for whoever merges

This PR and #631 both insert a bullet at the same point in `docs/source/changelog/1.5.0.rst`, so they conflict with each other (and with the other changelog-touching PRs in flight). Whichever lands second needs a rebase that keeps every bullet, then a `util/changelog_md.py` regeneration rather than a hand merge of `CHANGELOG.md`.


JarryShaw:
## Cross-review: **NEEDS CHANGES** — arrived after merge, and the fix is #638

Independent cross-review by a subagent on a **different model (Claude Haiku)**, run read-only on its own clones in `/tmp`, briefed to falsify rather than to bless.

**This PR merged before the review completed.** Posting the verdict anyway, because it found a real defect in what landed. **The code change in this PR is correct** — `LICENSE:3` reading `Copyright (c) 2017, Jarry Shaw` is right, and the licence body is intact. The defect is in the **changelog prose**, and it is corrected by **#638**.

### The change itself: confirmed correct

- **Licence body byte-identical apart from line 3.** Whole-file diff is one hunk, one line: `LICENSE | 2 +-`. The remaining 28 lines were normalised and compared token-by-token against the SPDX `licenseText` fetched from `spdx.org/licenses/BSD-3-Clause.json` — 213 tokens each side, `diff` exit 0 for this-PR-vs-SPDX, base-vs-SPDX and this-PR-vs-base. Nothing altered, dropped or reordered.
- **`LICENSE:29` clean.** `cat -A` gives `OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.$` — no trailing `s`; `grep -n 'DAMAGE\.s'` exit 1. No CRLF, no non-ASCII, trailing newline intact. 29 lines / 1510 bytes, against the base's 1515, the 5-byte delta being `-2026` dropped.
- **First-commit date confirmed, including a check for hidden roots.** `git rev-list --max-parents=0 origin/main` returns exactly one root, `c57f7d0b7`, authored `2017-11-07 13:39:45 +0800`. `--all` does reveal a second root, `7392a4b71`, but it is `Initial gh-pages commit` dated 2021-01-10, on `origin/gh-pages` only and not an ancestor of `main`. No earlier commit exists.
- **No third copyright-year site.** Re-derived independently with a plain recursive grep over the whole worktree (reaching `.github/`, `conda/`, `.idea/`, `.vscode/`, docs templates, `setup.py`, `pyproject.toml`) plus a second sweep for any `20(1[7-9]|2[0-9])` across `*.cfg *.toml *.yml *.yaml *.in *.cff setup.py *.json`: exactly `LICENSE:3` and `docs/source/conf.py:51`, nothing else copyright-related.
- **`docs/source/conf.py` genuinely untouched** — `cmp` against the base reports IDENTICAL, and line 51 still reads `copyright = f'2017-{datetime.date.today().year}, Jarry Shaw'`.
- **`CITATION.cff` has no copyright or copyright-year field** — `grep -niE 'copyright'` exit 1; the only date-shaped value in the file is `date-released: '2026-09-20'`.
- One commit on `ead73b204`, no merges, author and committer `Jarry Shaw <jarryshaw@icloud.com>`, and only `LICENSE` plus the two changelog files touched.

### The defect: two false claims in the changelog entry

**1. The pre-BSD history.** The entry asserts the file "until then was stock MPL text carrying no author notice at all, so 2018 is the first start year the project ever asserted and it was already a year late when it was written". Every clause of that is wrong as a description of the era. Enumerating all 14 commits that ever touched `LICENSE`: it was MIT (`c57f7d0b7`, 2017-11-07), then GPL v3, then Apache 2.0, then MPL 2.0 — four licences before BSD. **`c57f7d0b7:LICENSE:3` is `Copyright (c) 2017 Jarry Shaw`**, so the project asserted 2017 in its very first commit and kept doing so for seven months; `2018` first appeared in the Apache notice of 2018-12-08 as a single year, current for the year it was written. The passage also contradicted its own opening sentence, which dates the first commit to 2017-11-07.

The true history is a **better** argument than the one written: this change *restores* the year the project originally asserted rather than asserting it anew.

**2. The title-line attribution.** The entry credits the `BSD 3-Clause License` title line to "the opensource.org template". That template's licence text begins at the copyright line; the one whose body opens with that exact title is GitHub's choosealicense.com, which also carries the `Copyright (c) [year], [fullname]` comma this file uses.

### How it got past me

I checked `bc836cfa2^:LICENSE` and `1c69341dc:LICENSE`, found no author notice in either, and generalised. But `git rev-parse bc836cfa2^` is `8f937d7af`, whose tree holds `1c69341dc`'s blob — **the two things I checked were the same file**, so it was one data point covering only the last 18 months of a 2.5-year period. Enumerating every commit that touched `LICENSE`, rather than sampling the one before the relicence, is what found it.

### Also confirmed by the review, for the record

The `2017` start year was right, and the reasoning behind dropping the end year holds: `LICENSE` ships inside the sdist and the wheel so it cannot be computed at read time, copyright subsists from creation regardless of the notice, and BSD-3-Clause's canonical form carries a single year. No workflow or hook was added, deliberately.

**Corrected in #638.** No tests were run for either PR, deliberately — a `coverage run -m pytest` in this repo reached 41.4 GB RSS today and had to be killed, and neither change needs the suite.

PR #631, opened 2026-09-22
fix(packaging): ship CITATION.cff in the source distribution

JarryShaw:
`MANIFEST.in` carries `include README.md` (line 18), `include LICENSE` (19) and `include CHANGELOG.md` (37), but had no line for `CITATION.cff`. Its two `global-include` patterns are `*.rst` and `*.py`, and neither can match a `.cff`, so `CITATION.cff` was in the repository and in no source distribution at all. This adds `include CITATION.cff`.

Requested by the repository owner; no issue is open for it, so there is no `Fixes` line.

## Why this matters now

#625 has just taught `util/bump_version.py` to keep `CITATION.cff`'s `version` and `date-released` in step with the version bump, and the owner intends a beta release specifically to exercise that path. Releasing with this gap would publish an sdist that omits the very artefact the release is meant to be testing — the bump would update a file that then does not ship.

The omission was flagged when the file landed: #615 added `CITATION.cff` and noted it ("**`CITATION.cff` is not in the sdist.** … Left alone because `MANIFEST.in` is owned by a concurrent branch this wave"), and #619 touched `MANIFEST.in` only for `README.md`, so it did not address the gap.

It also matters independently of the release. GitHub renders the "Cite this repository" button from the *repository*, so the button works either way and the gap is invisible from the web UI. Citation managers, Zenodo and dependency inventories read the *published artifact*, which is exactly the surface where the file was missing — so credit stopped travelling with the code at precisely the point where nobody can see the repository.

## There was no default to fall back on

Worth establishing rather than assuming, because two of the three neighbouring `include` lines turn out to be redundant. Removing `include README.md`, `include LICENSE` and `include CHANGELOG.md` all at once from a clean checkout at `ead73b204` and rebuilding gives an sdist whose root is:

```
pypcapkit-1.5.0b4/LICENSE
pypcapkit-1.5.0b4/MANIFEST.in
pypcapkit-1.5.0b4/PKG-INFO
pypcapkit-1.5.0b4/pyproject.toml
pypcapkit-1.5.0b4/README.md
pypcapkit-1.5.0b4/setup.cfg
pypcapkit-1.5.0b4/setup.py
```

`README.md` and `LICENSE` still ship — setuptools adds the licence from its `license_files` default, logging `adding license file 'LICENSE'` and recording `License-File: LICENSE` in `PKG-INFO`. `CHANGELOG.md` is gone. So of the three, only `include CHANGELOG.md` is genuinely load-bearing, and a `.cff` is in the same position as it: nothing outside `MANIFEST.in` knows a citation file exists. That is what the new comment says, and it is why it anchors on `CHANGELOG.md` rather than on its immediate neighbour.

## Proof, both directions

Same command before and after the one-line change:

```console
$ python -m build --sdist --outdir /tmp/cffsdist .
$ tar tzf /tmp/cffsdist/pypcapkit-*.tar.gz | grep -i citation
```

**Before** — no output, exit 1. The root of the sdist held `CHANGELOG.md`, `LICENSE`, `MANIFEST.in`, `PKG-INFO`, `pyproject.toml`, `README.md`, `setup.cfg`, `setup.py`.

**After** — exit 0:

```
pypcapkit-1.5.0b4/CITATION.cff
```

Nothing else moved. Diffing the two full archive listings (sorted under `LC_ALL=C`) gives exactly one added line and nothing removed — 860 entries before, 861 after:

```console
$ comm -13 b.list f.list
pypcapkit-1.5.0b4/CITATION.cff
$ comm -23 b.list f.list        # removed: none
```

The shipped copy is the repository's copy, not a rewritten one — `sha256` of both is `9d231a0bb7ab5ceece390497b2f9c42818cd6dbc47b88bb6d52d276cef488a25`.

## `twine check` still passes

```console
$ python -m twine check --strict /tmp/cffsdist/*
Checking /tmp/cffsdist/pypcapkit-1.5.0b4.tar.gz: PASSED
```

`PASSED`, exit 0, on both the before and the after archive. Worth stating explicitly, since #497 records a beta that had to be reshipped as `1.5.0b2` after PyPI rejected its wheel over a `twine check` failure.

## The comment

The two commented `include` blocks in this file each explain why the line is load-bearing rather than belt-and-braces, so this one does too, in the same voice: which `global-include` fails to reach a `.cff`, that the absence was measured by building one each way rather than assumed, that no setuptools default covers it, and why the published artifact rather than the repository is the surface that matters.

## Changelog

A `**Fixed**` bullet in `docs/source/changelog/1.5.0.rst`, with `CHANGELOG.md` regenerated by `util/changelog_md.py` and `--check` confirmed to exit 0. `CHANGELOG.md` was not hand-edited.

No tests were run: this is one line of packaging configuration plus a changelog entry, and touches nothing under `pcapkit/`. `python -m build` was the relevant verification and it is quoted above.

## Notes for whoever merges

- This PR and #630 both insert a bullet at the same point in `docs/source/changelog/1.5.0.rst`, so they conflict with each other (and with the other changelog-touching PRs in flight). Whichever lands second needs a rebase that keeps every bullet, then a `util/changelog_md.py` regeneration rather than a hand merge of `CHANGELOG.md`.
- Found but deliberately **not** fixed here, to keep this PR to one line: the pre-existing comment at `MANIFEST.in:14-17` (from #619) claims `include README.md` is "the only thing that puts it in an sdist" and that "an sdist without it cannot be installed at all". The build above shows `README.md` shipping with that line removed, so the claim is wrong. Left alone as out of scope — happy to send it as its own PR if wanted.


JarryShaw:
@copilot resolve the merge conflicts on this branch.

JarryShaw:

Independent cross-review by a subagent on a **different model (Claude Haiku)**, run read-only on its own clones in `/tmp`, briefed to falsify rather than to bless. It took **three passes**; the first two returned **NEEDS CHANGES** and both found real defects, which is the point of doing it.

**Verdict on this PR: GOOD TO GO.**

> No false clause remains. The `MANIFEST.in` comment now reads "setuptools ships `LICENSE` from its `license_files` whether or not this file mentions it, but no packaging default covers a citation file" — accurate on both halves, and the `include CHANGELOG.md` anchor is the genuinely load-bearing one.

### What it caught, and what changed as a result

**Pass 1 — the comment's central analogy was inverted.** The first version of the new comment opened "Load-bearing for the same reason as `include LICENSE` above". Both halves of that were wrong: `include LICENSE` carries no comment of its own (the block above it is about `README.md`), and more importantly it is **not load-bearing**. Verified independently by removing all three `include` lines from a clean checkout and rebuilding — `README.md` and `LICENSE` still ship, `CHANGELOG.md` disappears. The build log says `adding license file 'LICENSE'` and `PKG-INFO` carries `License-File: LICENSE`. The comment was rewritten to anchor on `include CHANGELOG.md`, the one genuinely load-bearing neighbour, and that measurement is now stated in the changelog entry too.

**Pass 2 — "nothing outside this file knows a citation file exists" was false.** `git grep -nl CITATION` returns `util/bump_version.py` and `tests/project/test_bump_version.py`. Narrowed to "no packaging default covers a citation file".

**Pass 2 also surfaced a second, independent motivation for this PR**, now recorded in the entry: `RepositoryCitationTests` in `tests/project/test_bump_version.py` — the gate #625 added for the hand-authored bumps that never run the script — currently skips itself with `'CITATION.cff is not shipped in the source distribution'`. This PR is what makes that gate live in the sdist context. The reviewer then checked its own suggestion rather than assuming it was safe, confirming that the test file ships (entry 753 of 861) and that its `ROOT` resolves to the sdist root, so the skip genuinely fires today and genuinely stops.

**Two PR-body claims were also overstated and were corrected:** that the surrounding `include` lines "each carry a comment" (two do not), and a conflation of #497 with #498 — #497 is the `twine check` rejection, #498 was a network timeout, and only #497 is cited now.

### Independently re-measured on the final head, against the rebased base `cfb81d3f6`

- 860 archive entries before, **861** after; `diff` output is exactly `2a3` / `> pypcapkit-1.5.0b4/CITATION.cff`; nothing removed
- `grep -i citation` exit 1 before, exit 0 after
- `twine check --strict` → `PASSED`, exit 0, on **both** archives
- shipped `CITATION.cff` sha256 `9d231a0bb7ab5ceece390497b2f9c42818cd6dbc47b88bb6d52d276cef488a25`, identical to the repository copy
- `util/changelog_md.py --check` exit 0; both changelog bullets present in order with this PR's last, before the closing paragraph
- one commit on `cfb81d3f6`, no merges, author and committer `Jarry Shaw <jarryshaw@icloud.com>`, and only `MANIFEST.in` plus the two changelog files touched

### Advisory, non-blocking

- **Merge #638 before this one** if there is a choice. They conflict on the changelog files; #638 edits the #630 passage that this PR does not touch, so that order makes this one a clean append. Resolve the `.rst` keeping every bullet and regenerate `CHANGELOG.md` rather than merging it by hand.
- The bullet list in `1.5.0.rst` is inconsistent about blank lines between items — pre-existing, not made worse here, and free to tidy while the file is being hand-resolved anyway.
- **Left deliberately unfixed:** the pre-existing comment at `MANIFEST.in:14-17` (from #619) claims `include README.md` is "the only thing that puts it in an sdist" and that "an sdist without it cannot be installed at all". The build above shows `README.md` shipping with that line removed, so both halves are wrong. Kept out of this PR to hold it to one line of behaviour change; happy to send it separately.

### Caveats the reviewer stated rather than glossed

Every measurement resolved `setuptools==84.0.0` in the isolated build env, so behaviour at the declared floor of `setuptools>=61.0.0` is **not** verified. No tests were run at all (deliberately — a `coverage run -m pytest` in this repo reached 41.4 GB RSS today and had to be killed), so the `RepositoryCitationTests` conclusion is read from source at lines 482-497 rather than from a run.

PR #656, opened 2026-09-22
docs(contributing): correct four statements the recent merges falsified

JarryShaw:
## What is the purpose of this pull request?

- [x] `docs` — documentation only

`CONTRIBUTING.md` had accumulated four statements that the tree no longer supports. One was the
reported defect; three came out of checking the rest of the file against `da381f259` while in there.

## 1. The reported defect — "not run by the pull-request workflows"

`CONTRIBUTING.md:104` read:

> None of these is run by the pull-request workflows, so they are a local gate rather than something a
> pull request will fail on. Running them anyway saves a review round.

#626 made that false when it added `.github/workflows/lint.yml`. The correction states, from the
workflow rather than from memory:

| linter | in `lint.yml`? | triggers | blocking? |
|---|---|---|---|
| `bandit` | yes | `pull_request` → `main`, `schedule` Sat 06:00 UTC, `workflow_dispatch` | no — `continue-on-error: true` |
| `mypy` | yes | same | no — `continue-on-error: true` |
| `vermin` | yes (`vermin-ci`) | same | no — `continue-on-error: true` |
| `pylint` | yes | same | no — `continue-on-error: true` |
| `isort` | **no** | only `cron-vendor.yml`, as a formatter | n/a |

Three things the new prose is careful about, because getting any of them wrong would mislead a
contributor in a new direction:

- **Advisory, not blocking.** All four steps carry `continue-on-error: true`
  (`lint.yml:117,134,163,183`), so a red linter cannot fail a pull request. "Linting runs in CI"
  without that qualifier would be worse than the stale sentence it replaces.
- **`isort` is excluded.** It is not in `lint.yml` at all; its only workflow appearance is inside
  `cron-vendor.yml`, re-sorting regenerated constants as a formatter rather than checking anything.
  Import ordering is still unverified on a pull request.
- **`make vermin` does not fail on violations.** `Makefile:147-150` redirects into `temp/vermin.txt`
  and hands the file to `code`, falling back to `cat`, so the status returned is the viewer's.
  `make vermin-ci` (`Makefile:158-159`) is the target CI actually runs.

The text also records that CI invokes the same `Makefile` targets with `RUN=`, which is what keeps
local and CI from drifting, and points at the workflow header for the current counts rather than
copying numbers that will go stale.

## 2. Three further corrections

- **`CONTRIBUTING.md:11`** claimed the README has a *Testing* section holding the test commands. It
  does not — the README's headings are `Installation`, `Usage` and `Documentation`, and its only
[...]

PR #659, opened 2026-09-22
docs(contributing): the `make vermin` trap is the opposite of what was recorded

JarryShaw:
## What is the purpose of this pull request?

- [x] `docs` — documentation only

Follow-up to #656, which merged while the cross-review of it was still running. Four corrections to
the section #656 rewrote: one is a claim that is **actively wrong and inverts what a contributor
should do**, three are precision fixes the cross-review raised.

## 1. `make vermin` fails on violations — the opposite of what #656 said

#656 shipped this, and it is false:

> The status you get back is that viewer's rather than vermin's, so a run that found violations still
> looks like a success.

That sentence was not invented for #656 — it is what `Makefile:152-157` and `lint.yml:150-152` both
assert, and it was carried across in good faith. The reasoning behind it looks at the shell semantics
of `command -v code >/dev/null && code temp/vermin.txt || cat temp/vermin.txt` and correctly concludes
that line always exits 0. What it misses is that **the redirect is a separate recipe line**, and make
abandons a target at the first line that exits non-zero, so the viewer line is never reached at all.

Measured on this tree, not reasoned about:

```console
$ PATH=.venv/bin:$PATH make vermin RUN=
mkdir -p temp
vermin pcapkit --backport argparse ... pcapkit > temp/vermin.txt
make: *** [vermin] Error 1
MAKE_EXIT=2
```

```console
$ vermin pcapkit --backport argparse ... pcapkit      # the same invocation, un-redirected
Minimum required versions: 3.11
Incompatible versions:     2.x
Target versions not met:   3.6
VERMIN_EXIT=1
```

`temp/vermin.txt` is written (365 lines) but never displayed, and `REACHED_VIEWER_LINE` in a minimal
reproduction of the same three-line recipe is never printed. A control where the tool succeeds *does*
reach the viewer line and exits 0 — which is what isolates the redirect, rather than the `A && B || C`
chain, as the deciding factor. GNU Make 3.82, no `.ONESHELL:` anywhere in the `Makefile`.

vermin exits 1 today because `vermin.ini` sets `targets = 3.6` while the code's real floor is 3.11.

**Why this matters more than the wording:** a contributor who believed the old text would see a red
`make vermin`, remember being told it cannot fail on violations, and dismiss a genuine failure as a
known false pass. The trap is real but inverted — the failure propagates and the *findings* are what
get hidden, in a file the target never opens.

## 2. Three precision fixes from the cross-review of #656

- **The file contradicted itself four lines apart.** #656's CI paragraph claimed the job runs "the
  same `Makefile` targets listed above", but `lint.yml:166` runs `make vermin-ci`, which is not in
  that block — and the paragraph further down admitted it. It now says which target each tool takes,
  and rests the no-drift claim on `$(VERMIN_FLAGS)` and friends being shared between `vermin` and
  `vermin-ci`, which is the thing that actually holds.
- **"every step writes its count there"** is wrong for vermin: `lint.yml:173` emits
  `Minimum required versions:` / `Incompatible versions:` / `Target versions not met:`, which are
  version verdicts, not counts. Now "each step writes its own verdict there".
- **`Changelog drift` has a third trigger.** The job carries no `gate-only` guard, unlike its sibling
  jobs, so it also runs on every `workflow_call` into `unit-tests.yml` — `cron-vendor.yml:27`,
  `cron-conda.yml:29`, `deploy-pages.yml:37` and `create-release.yml:40` all gate on it.

## Notes

- **`Makefile:152-157` and `lint.yml:150-152` still carry the wrong claim.** Deliberately not fixed
  here: `.github/workflows/**` should not be touched ahead of the `1.5.0b5` release, and the two
  comments want their own change. Worth doing, because they are what this documentation was derived
  from in the first place — the error will propagate again otherwise.
- **No changelog entry, deliberately** — not user-visible, and a bullet would collide with the
  changelog consolidation in flight.
- No file outside `CONTRIBUTING.md` is modified. The test suite was not run; nothing here executes
  library code.
- One commit, authored `Jarry Shaw <jarryshaw@icloud.com>`, on top of `7b0df4a9c`.


JarryShaw:
The cross-review verdict that produced this PR is on #656: https://github.com/JarryShaw/PyPCAPKit/pull/656#issuecomment-5780026695

It opened **NEEDS CHANGES** against #656 and raised items 1-3 of the section above. Item 1 here — the `make vermin` inversion — is *not* from the review: the review marked that claim CONFIRMED, agreeing with the false version. It was found separately by testing the target instead of reading it.

Unpublished-equivalent status: this PR is open and awaiting your review. Nothing here has been merged, tagged or released.

Issue #655, opened 2026-09-22
Neither the HIP PUZZLE nor SOLUTION builder consults version when sizing, so HIPv1 rejects what it builds


JarryShaw:
`_make_param_puzzle` and `_make_param_solution` both take a `version` keyword and neither reads it when sizing the parameter. Both derive `Length` from `int.bit_length()` alone. Under HIPv1 that is not allowed to vary: RFC 5201 fixes `Random #I` and `Puzzle solution #J` at 8 octets each and therefore fixes `Length` at 12 for PUZZLE and 20 for SOLUTION. So any value narrower than the full field builds a parameter that **this library's own reader then rejects** — the same shape as #608, in a second guise, and present in the PUZZLE builder that #629 deliberately did not touch.

## The RFC text, fetched

Fetched from `https://www.rfc-editor.org/rfc/rfc5201.txt` (240,492 octets, sha256 `8b42d181a8e239713eb8d608c11e8e75829561d994adbd3caa96c6db6e69cef6`). Unlike RFC 7401, HIPv1 states both widths as literal constants. RFC 5201 §5.2.4, PUZZLE, verbatim:

```
      |                      Random #I, 8 bytes                       |
      |                                                               |
      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

      Type           257
      Length         12
```

and RFC 5201 §5.2.5, SOLUTION:

```
      |                      Random #I, 8 bytes                       |
      |                                                               |
      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
      |                 Puzzle solution #J, 8 bytes                   |
      |                                                               |
      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

      Type               321
      Length             20
```

[...]

Issue #653, opened 2026-09-22
Re-serialising a parsed HIP PUZZLE or SOLUTION loses leading zero octets: a len=20 parameter rebuilds as len=6


JarryShaw:
`_make_param_puzzle` and `_make_param_solution` recompute the width of `Random #I` and `Puzzle solution #J` from `int.bit_length()`, and the data model carries no width to recompute it from. So a parameter that arrived on the wire with an 8-octet `Random #I` of `0x0000000000000001` re-serialises into a single octet: the value survives, the field width does not. A SOLUTION parsed with `Length = 20` rebuilds as `Length = 6`.

This matters because the field width is not decoration in this parameter. `RHASH_len` is the output length of the puzzle hash, and both fields are defined as exactly that wide — RFC 7401 §5.2.5 gives `Random #I` as a "random number of size RHASH_len bits" and `Length` as `4 + RHASH_len / 4`. Narrowing the field changes the implied `RHASH_len`, so the rebuilt parameter describes a different puzzle from the one that was received.

## The sites

`pcapkit/protocols/internet/hip.py:3195-3196`, in `_make_param_solution`'s `param is not None` branch — the path taken when re-serialising something that was parsed:

```python
random = param.random
solution = param.solution
```

`pcapkit/protocols/internet/hip.py:3212`, sizing from those ints and nothing else:

```python
len=4 + 2 * math.ceil(max(random.bit_length(), solution.bit_length()) / 8),
```

The sibling `_make_param_puzzle` has the same shape at `:3152` and `:3160`:

```python
len=4 + math.ceil(random.bit_length() / 8),
```

There is nothing wrong with `bit_length()` as such — the problem is that it is the *only* source of width available. `pcapkit/protocols/data/internet/hip.py:196` and `:213` define the data models, and neither carries a width field:

```
Data_SolutionParameter fields = ['type', 'critical', 'length', 'index', 'lifetime', 'opaque', 'random', 'solution']
```

`length` is the record length, not the field width, and it is not consulted by either builder. The reader discards the width on the way in: `pcapkit/protocols/schema/internet/hip.py:454` and `:456` size the two fields at `(pkt['len'] - 4) // 2` each and hand back plain ints, so by the time `_read_param_solution` builds the data object the octet count is gone.

## Measured

On `origin/main` (`a62aed134`), CPython 3.14.7, importing from an immutable `git archive` snapshot with `pcapkit.__file__` asserted against it:

```
[...]

PR #664, opened 2026-09-22
fix(hip): pad HIP parameters to the record length RFC 7401 gives, not the contents (#651)

### LINKED ISSUE #651: HIP parameter padding aligns the contents, not the record, so every parameter pcapkit emits is 4 (mod 8) octets
Every padding site in the HIP schemas computes the pad from the contents length alone — `(8 - (pkt['len'] % 8)) % 8` — which aligns `len` rather than `len + 4`. RFC 7401 §5.2.1 requires the *total* length of a TLV parameter, type and length fields included, to be a multiple of 8 octets. Aligning the contents instead puts every parameter pcapkit writes at 4 (mod 8), never 0 (mod 8), for every possible `Length` value.

**The sharp part: pcapkit round-trips these records perfectly, because its writer and its reader share the same error.** A real peer does not share it. So no round-trip test can detect this, however many parameters it covers — the comparison has to be against the RFC, not against pcapkit. That is why this survived a suite that already round-trips 29 of the 49 HIP parameter codes.

## The RFC text, fetched

Fetched from `https://www.rfc-editor.org/rfc/rfc7401.txt` (309,319 octets, sha256 `09366b9f83dc80593172304ffcf05f57afd186aacb468ac6170a8fe26944c4be`). RFC 7401 §5.2.1, "TLV Format", verbatim:

```
   All of the encoded TLV parameters have a length (that includes the
   Type and Length fields), which is a multiple of 8 bytes.  When
   needed, padding MUST be added to the end of the parameter so that the
   total length is a multiple of 8 bytes.  This rule ensures proper
   alignment of data.  Any added padding bytes MUST be zeroed by the
   sender, and their values SHOULD NOT be checked by the receiver.

   The Length field indicates the length of the Contents field (in
   bytes).  Consequently, the total length of the TLV parameter
   (including Type, Length, Contents, and Padding) is related to the
   Length field according to the following formula:

   Total Length = 11 + Length - (Length + 3) % 8;

   where % is the modulo operator.
```

The RFC states the arithmetic explicitly, so there is nothing here to infer from a diagram. `Length` is the contents length — "Length of the Contents, in bytes, excluding Type, Length, and Padding" — and the *total* is what must be a multiple of eight.

## The sites

`git grep -c '% 8'` over the two HIP modules — the trailing number is a match count, not a line number — and every one of those 95 matches is a padding computation:

```
$ git grep -c '% 8' -- pcapkit/protocols/internet/hip.py pcapkit/protocols/schema/internet/hip.py
pcapkit/protocols/internet/hip.py:49
pcapkit/protocols/schema/internet/hip.py:46
```

All 46 in the schema module are the same expression, once per parameter class, e.g. `pcapkit/protocols/schema/internet/hip.py:325`:

```python
padding: 'bytes' = PaddingField(length=lambda pkt: (8 - (pkt['len'] % 8)) % 8)
```

All 49 in the protocol module are the reported record length in the data model, e.g. `pcapkit/protocols/internet/hip.py:796`:

```python
length=4 + schema.len + (8 - schema.len % 8) % 8,
```

Neither expression has a `+ 4` inside the modulus. Grepping for anything else: `grep -n '% 8' … | grep -v <that expression>` returns nothing in either file, so there are no other `% 8` uses and no site that already gets it right.

## Measured

On `origin/main` (`a62aed134`), CPython 3.14.7, importing from an immutable `git archive` snapshot whose `pcapkit.__file__` is printed and asserted rather than trusting the editable install:

```
MEASURED pcapkit.__file__ = /tmp/hipissues/tA62/pcapkit/__init__.py
CPython 3.14.7

SOLUTION len=8 -> pcapkit packs 12 octets: 01 41 00 08 01 21 6f 70 00 01 00 01
SOLUTION len=8 -> RFC 7401 5.2.1 requires 16 octets (11 + 8 - (8+3)%8 = 16)
  pcapkit padding = (8 - (8 % 8)) % 8 = 0 ; RFC padding = 4
```

A `len=8` SOLUTION emits twelve octets where sixteen are required, exactly as reported. The defect is not confined to that case, and it is not a consistent offset in one direction — pcapkit both under-pads and over-pads depending on the residue:

```
   L | L%8 | pcapkit pad | pcapkit total | RFC pad | RFC total | agree?
   0 |  0  |      0      |        4      |    4    |      8    | no
   1 |  1  |      7      |       12      |    3    |      8    | no
   4 |  4  |      4      |       12      |    0    |      8    | no
   8 |  0  |      0      |       12      |    4    |     16    | no
  12 |  4  |      4      |       20      |    0    |     16    | no
  16 |  0  |      0      |       20      |    4    |     24    | no
  20 |  4  |      4      |       28      |    0    |     24    | no

pcapkit total is a multiple of 8 for any L?  False
pcapkit agrees with RFC for any L in 0..63?  False
```

Those two `False` results are the whole issue in one line. Across every contents length from 0 to 63 there is **not one** for which pcapkit's record length is a multiple of eight, and not one for which it matches the RFC.

`L = 4` is the case that shows it is not merely a shortfall. SEQ carries a single 4-octet Update ID, so it is complete in eight octets and needs no padding at all — and pcapkit appends four:

```
SEQ update_id=0x01020304 -> len=4, packs 12 octets: 01 81 00 04 01 02 03 04 00 00 00 00
RFC 7401 5.2.1: Total Length = 11 + 4 - (4+3) % 8 = 8 octets
pcapkit appends 4 octets of padding; RFC requires 0
```

SEQ is not a corner of the parameter space. RFC 7401 §5.3.5 states that "An UPDATE packet without either a SEQ or an ACK parameter is invalid", so one of the two is on every UPDATE, and both carry contents that are a multiple of four octets — precisely the residue at which pcapkit adds four octets that must not be there.

The consequence on pcapkit's own reader, same run:

```
One SOLUTION per packet, built then re-parsed by pcapkit:
  REJECTED: ProtocolError: HIPv2: invalid format
Two SOLUTIONs per packet:
  built 96 octets
  parsed OK  <- writer and reader share the same padding error
```

One parameter is 4 (mod 8) and the header length field cannot represent it; two are 0 (mod 8) and it can. The error cancels in pairs, which is why the symptom looks like an odd restriction on packet shape rather than like a padding bug.

## The codebase already documents this

`examples/generators/options.py:949-962` describes the defect in terms and works around it, which is why the round-trip suite is green:

```
#: How many copies of the parameter under test go in one packet.
#:
#: Two, not one, and this is not padding for its own sake. ``HIP.make``
#: computes the header's ``len`` field as ``total_length // 8 + 4``, which is
#: only lossless when the parameter octets are a multiple of eight -- and the
#: parameter padding rule pads the *contents* to eight, ignoring the four-octet
#: type-and-length header, so a single parameter is always ``4 (mod 8)`` and
#: always loses those four octets. ``_read_hip_param`` then checks the value
#: exactly and rejects the packet. Two copies sum to a multiple of eight, so
#: the arithmetic is exact and the parameter constructors become reachable:
#: measured, this is the difference between 4 and 29 of the 49 codes
#: round-tripping. The single-parameter case is not lost -- it is what
#: ``hip-parameter-single`` in the test's expected-failure table records.
HIP_COPIES = 2
```

"the parameter padding rule pads the *contents* to eight, ignoring the four-octet type-and-length header" is this defect, already written down. It was read as a constraint to be accommodated rather than as a bug, and `HIP_COPIES = 2` is the accommodation — it is the difference between 4 and 29 codes round-tripping. The `hip-parameter-single` expected-failure entry is the same defect recorded as a known limitation.

## Consequence

Two directions, both interoperability failures rather than crashes.

Writing: every HIP packet pcapkit produces carries parameters whose total length violates §5.2.1. A conformant receiver reads `Length`, computes `11 + Length - (Length + 3) % 8`, and consumes that many octets — for a `len=8` SOLUTION it consumes sixteen where pcapkit wrote twelve, taking four octets of the following parameter as padding and then reading the rest of the parameter area at a wrong offset.

Reading: given a conformant packet, pcapkit consumes `4 + Length + (8 - Length % 8) % 8` and lands in the wrong place by the same amount. At `Length = 20` it consumes 28 octets of a 24-octet record, so the next parameter's type is read four octets into the wrong field.

Neither shows up in this repository's fixtures, because every HIP fixture was produced by this library and is self-consistent.

## Fixing it rewrites every HIP fixture

The padding is on the wire, so correcting the formula changes the octets of every HIP parameter in every synthesised capture, and the header `len` field along with them. `examples/generators/make_samples.py` regenerates them wholesale, but the diff is the entire HIP fixture set, and `HIP_COPIES = 2` and the `hip-parameter-single` expected-failure entry both become removable at the same time. That is a wide change and wants to be its own, which is why #629 measured it and deliberately left it alone.

## Notes

- Found by the worker fixing #608 while measuring that defect, and reported in PR #629's body rather than folded in. #608 was the SOLUTION `len` formula; this is the padding, in all 95 sites.
- Re-verified here from scratch against `a62aed134` before filing: the RFC fetched rather than paraphrased, the grep counts re-run, and the arithmetic executed rather than reasoned about.
- Also reproduces identically on `375e9d411`, `a62aed134`'s parent, where #608 is still unfixed — the schema module is byte-identical between the two commits (sha256 `b56203d2a044e69a91743c812bf801af619494b7323e89f34b358135ec74814d`), so #629 did not touch any padding site.
- Related: #608 and PR #629 (SOLUTION `len` arithmetic, where this was found). The other three side-findings from that work are #653 (re-serialising loses leading zero octets), #654 (SOLUTION's `Reserved` octet written as a `Lifetime`) and #655 (neither builder consults `version`). This one is the only one of the four that affects every HIP parameter rather than PUZZLE and SOLUTION alone, and the only one whose fix rewrites the fixtures.

JarryShaw:
Fixes #651.

## The RFC text, fetched

`https://www.rfc-editor.org/rfc/rfc7401.txt`, 309,319 octets, sha256
`09366b9f83dc80593172304ffcf05f57afd186aacb468ac6170a8fe26944c4be`. Section
5.2.1, "TLV Format", verbatim:

```
   All of the encoded TLV parameters have a length (that includes the
   Type and Length fields), which is a multiple of 8 bytes.  When
   needed, padding MUST be added to the end of the parameter so that the
   total length is a multiple of 8 bytes.  This rule ensures proper
   alignment of data.  Any added padding bytes MUST be zeroed by the
   sender, and their values SHOULD NOT be checked by the receiver.

   The Length field indicates the length of the Contents field (in
   bytes).  Consequently, the total length of the TLV parameter
   (including Type, Length, Contents, and Padding) is related to the
   Length field according to the following formula:

   Total Length = 11 + Length - (Length + 3) % 8;

   where % is the modulo operator.
```

and, from the same section's field list, `Length` is "Length of the Contents, in
bytes, excluding Type, Length, and Padding" with `Padding` "0-7 bytes, added if
needed". So it is the **total** that must be 8-aligned, and `Length` is only the
contents.

## Where the padding changed

One site, not 95. Both expressions were byte-identical at every occurrence
(`git grep -c '% 8'`: 46 in `schema/internet/hip.py`, 49 in `internet/hip.py`,
and `sort -u` over the matched lines returns exactly one distinct line in each
file), so they collapse into:

```python
def parameter_total_len(length: 'int') -> 'int':
    ...
    return 11 + length - (length + 3) % 8


def parameter_padding_len(pkt: 'dict[str, Any]') -> 'int':
    ...
    length = pkt['len']
    return parameter_total_len(length) - 4 - length
```

`grep '% 8'` over the two files now returns two lines: the RFC quote in
`parameter_total_len`'s docstring, and its `return`.

### The helper takes no `version`, and RFC 5201 is why

Worth stating because HIPv1 and HIPv2 differ on several parameter field widths,
and #655 is about the builders not consulting `version` — so a reader could
reasonably ask whether padding needs it too. It does not. RFC 5201 §5.2.1, fetched
(`240,492` octets, sha256 `8b42d181…9cef6`) and quoted rather than paraphrased:

```
   All of the TLV parameters have a length (including Type and Length
   fields), which is a multiple of 8 bytes.  When needed, padding MUST
   be added to the end of the parameter so that the total length becomes
   a multiple of 8 bytes.  This rule ensures proper alignment of data.
   ...
   Total Length = 11 + Length - (Length + 3) % 8;
```

The same formula as RFC 7401 §5.2.1, to the character. So `parameter_total_len`
is correctly version-independent, and nothing here needs to change when #655
lands.

## `HIP.make`'s `len` computation was **not** the second defect

`total_length // 8 + 4` is correct, and RFC 7401 section 5.1.3 says why:

```
   The Header Length field contains the combined length of the HIP
   Header and HIP parameters in 8-byte units, excluding the first
   8 bytes.
```

The fixed header is 40 octets, so `Header Length = (40 - 8 + params) / 8 =
4 + params / 8`. The repo's own note called it "lossless only when the parameter
octets are a multiple of eight" — which is exactly the precondition the padding
fix now establishes, so the floor division is exact rather than exact-in-pairs.
`_make_hip_param` sums real `pack()` lengths, every one of which is now a
multiple of 8. Left alone deliberately.

## One coupled fix, forced rather than opportunistic

`EncryptedParameter.data`'s length callback subtracted the 16-octet `iv` but not
the four `reserved` octets, while `_make_param_encrypted` writes
`len = 4 + len(iv) + len(data)`. The old record total was therefore
`8 + Length + (-Length % 8)` against the RFC's `4 + Length + pad`.

**Correction to an earlier revision of this description**, which said those
"coincide **only** when `Length % 8 == 0`" — that is wrong, and the earlier table
only sampled residues 0 and 4, which is how it went unnoticed. Measured by
packing through `_make_param_encrypted` at every residue on `0c7f2b7c9`:

| `Length` | `Length % 8` | pre-PR octets | RFC octets | agree? |
|---|---|---|---|---|
| 4 | 4 | 16 | 8 | no |
| 5 | 5 | 16 | 16 | **yes** |
| 6 | 6 | 16 | 16 | **yes** |
| 7 | 7 | 16 | 16 | **yes** |
| 8 | 0 | 16 | 16 | **yes** |
| 9 | 1 | 24 | 16 | no |
| 10 | 2 | 24 | 16 | no |
| 11 | 3 | 24 | 16 | no |
| 12 | 4 | 24 | 16 | no |

So they agreed at `Length % 8` in **{0, 5, 6, 7}** and differed at **{1, 2, 3, 4}**
— four residues, not one. `Length = 8`, which the unit suite happened to use, is
one of the four, which is why the module emitted RFC-conformant `ENCRYPTED`
octets there while getting both halves wrong.

Fixing the padding alone would still have been a regression, and for a stronger
reason than the original wording gave: with only the padding corrected the total
becomes `8 + Length + pad` against a correct `4 + Length + pad`, a uniform
four-octet surplus with **no** residue left where it cancels. So `ENCRYPTED`
would have gone from right at four of the eight residues to wrong at all eight.
The two go together; `encrypted_data_len` is the third new helper, and the
residue sets above are now asserted in
`test_hip_encrypted_data_length_excludes_reserved_and_iv` rather than only
claimed in prose.

## Widths, and why they discriminate

The defect was exactly `4 (mod 8)`, so a convenient spread of multiples of 8
would prove nothing. `test_hip_parameter_total_length_matches_the_rfc_7401_formula`
sweeps **every** `Length` in 0..63 — all eight residues — against a `rfc_total`
transcribed inline from the RFC rather than imported from the implementation
(comparing an implementation with itself asserts nothing). What each residue
class rules out:

- **`Length % 8 == 4`** (`4` — a whole `SEQ`; `20` — a whole `SOLUTION`): the
  RFC wants **zero** padding. The old rule appended 4; a rule that dropped the
  outer `% 8` (`8 - (Length + 4) % 8`) would append 8. Only this residue
  separates the right answer from both.
- **`Length % 8 == 0`** (`0`, `8`, `16`): contents are 8-aligned alone, so the
  old rule appended nothing and left the record 4 octets **short**. This is the
  under-padding residue, invisible to any "result is at least as long as the
  contents" check.
- **`Length` not a multiple of 4** (`1`,`2`,`3`,`5`,`6`,`7` → pad 3,2,1,7,6,5):
  catches a formula that only ever moves in steps of four, which both the old
  and the fixed one resemble at a glance.

The test also asserts `total != pre_651_total(length)` at every `Length`, i.e.
there is no residue at which it would have passed before this change.

**A gap the cross-review found, now closed.** Eight residues are enough for any
formula that is periodic in `Length % 8` — but not for one that is not. A
`parameter_total_len` masking its argument (`length & 0xFF`) agrees on 0..63 and
diverges at 256, and nothing in the generator builds a parameter that long, so it
passed the *entire* suite. The test now also checks the whole domain of the
16-bit `len` field in one loop. Verified by planting exactly that formula in a
scratch copy of this branch:

```
421:    length &= 0xFF  # deliberately wrong: masks to 8 bits
RESOLVED pcapkit.__file__ = /tmp/hip651/masked/pcapkit/__init__.py
=== masked formula vs the arithmetic test: exit 1 ===
E  - [256, 257, 258, ...]
E  + [] : parameter_total_len diverges from RFC 7401 5.2.1 at 65280 of the
         65536 representable Length values, first at [256, 257, ...]
1 failed, 2 warnings, 64 subtests passed in 0.73s
```

## Failing, then passing

Both runs import from a tree that is **printed and asserted** first, with
`PYTHONPATH` set and the editable finder accounted for (see the trap below).

New tests against the pristine `0c7f2b7c9` library. **An earlier revision of this
description gave one figure without saying which files it covered**, which
invited reading it as both modules; here are both, from a run whose library was
checked by md5 against `git show 0c7f2b7c9:` and whose test files were checked by
md5 against this branch:

```
RESOLVED pcapkit.__file__ = /tmp/hip651-owner-only/before/pcapkit/__init__.py
CPython 3.14.7
OK
--- tests/protocols/internet/test_hip_unit.py only:  exit 1 ---
76 failed, 22 passed, 2 warnings, 74 subtests passed in 28.31s
--- ... plus tests/protocols/test_option_roundtrip_unit.py: exit 1 ---
79 failed, 28 passed, 2 warnings, 431 subtests passed in 28.96s
```

Both are real and differ only in file selection. Broken down, because
`pytest-subtests` counts a failing *subtest* as its own "failed" entry:

| run | SUBFAILED | FAILED methods | "failed" | passed methods | subtests passed |
|---|---|---|---|---|---|
| HIP unit only | 71 | 5 | 76 | 22 | 74 |
| both modules | 74 | 5 | 79 | 28 | 431 |

The round-trip module adds 3 failures — the two subtests of
`test_a_hip_packet_carrying_one_parameter_round_trips`, plus the ENCRYPTED case in
`test_round_trip_is_identity_or_a_recorded_gap`, which on the old library still
`MISMATCH`es while this branch's table no longer records it.

Worth flagging because it tripped the cross-review: **`74` is the number of
subtests that *passed* in the one-module run and, coincidentally, the number that
*failed* in the two-module run.** Same integer, different quantity.

The failures are real octet counts rather than an `AttributeError` — the
octet-level test deliberately imports no new helper:

```
(maker='_make_param_seq',        length=4)  AssertionError: 12 != 8
(maker='_make_param_esp_info',   length=12) AssertionError: 20 != 16
(maker='_make_param_solution',   length=20) AssertionError: 28 != 24
(maker='_make_param_unassigned', length=0)  AssertionError: 4 != 8
(maker='_make_param_unassigned', length=1)  AssertionError: 12 != 8
(maker='_make_param_unassigned', length=5)  AssertionError: 12 != 16
```

Over-padding at `4`, `1`; under-padding at `0`, `5`. Same tests on this branch:

```
RESOLVED pcapkit.__file__ = .../worktrees/agent-a5e1c50562a5aab83/pcapkit/__init__.py
OK
=== after_hip2: pytest exit code 0 ===
33 passed, 1 warning, 507 subtests passed in 25.77s
```

against `30 passed, 434 subtests` on `0c7f2b7c9` for the same two modules.

## Fixture evidence that does not trust pcapkit at all

`examples/generators/make_samples.py` regenerates the set (never `rm -rf
examples/captures` — `dhcp.pcapng` and `in.pcap` are committed). A standalone
walker then parses `options-internet.pcap` with **no pcapkit**, taking its stride
solely from `11 + Length - (Length + 3) % 8`, so a record written to any other
rule desynchronises it immediately:

```
0c7f2b7c9   15006 octets  145 frames  46 HIP frames  98 records  43 violations
this branch 14974 octets  145 frames  46 HIP frames  92 records   0 violations
```

It also checks that each record's padding is zeroed, as 5.2.1 requires of the
sender, and that the header's `len` accounts for exactly the octets walked.

**Read "0 violations" narrowly, though — the cross-review found that it conceals
one record, and it is worth spelling out because it is the same
two-defects-cancelling shape as everything else here.** `R1_COUNTER` declares
`Length = 12` but its schema packs only 12 octets in total, four short of the
RFC's 16, because of the separate `counter` width defect described below. A
Length-driven reader therefore advances 16 and lands four octets inside the
*second* copy, where it reads a phantom `Type = 0, Length = 0` record out of that
copy's own zeroed contents — whose "padding" is zero, so the zero-padding check
passes and the walk ends tidily on the area boundary. The generator builds
`R1_COUNTER` with `counter = 0` (it has no entry in `_hip_overrides()`), which is
exactly what makes the desynced octets indistinguishable from real padding.

Measured, by patching only the second copy's counter in a scratch copy of the
file and re-walking:

```
as generated            : 92 records, 0 violations
with a non-zero counter : 92 records, 1 violations
    frame 101: type 0 padding not zeroed: aabbccdd
```

So the honest reading is **"no padding violations among the records a conformant
reader can find"**, not "the capture is conformant". The padding fix is not
implicated — the concealment is entirely the `counter` width defect — and
`43 → 0` remains a real difference, since all 43 pre-fix violations were padding
ones. Two follow-ups this suggests, neither taken here: give `R1_COUNTER` a
non-zero `counter` in the generator's `_hip_overrides()` so the fixture stops
hiding it (outside this PR's remit for that file, which is `HIP_COPIES` and its
note), and fix the width itself.

## `HIP_COPIES = 1`, measured — and why it stays at 2

Measured over the generator's 49 HIP codes, at both settings, on both trees:

| tree | one copy | two copies |
|---|---|---|
| `0c7f2b7c9` | **4 OK** | 45 OK |
| this branch | **45 OK** | 46 OK |

So 41 codes that could not survive alone now can, which is the strongest
available statement that the padding was what made a lone parameter
unrepresentable — and better evidence than any assertion in this PR.

It still stays at **2**, because the four codes that fail at one copy fail for
reasons unrelated to padding: `R1_COUNTER` and `R1_Counter` (the `counter` width),
`HOST_ID` (declares `len=8`, packs 18) and `HIP_TRANSFORM` (HIPv1-only, built here
at version 2). Flipping the constant would trade this fix's workaround for one new
expected-failure entry, one changed status, and the loss of `R1_COUNTER` from the
round-tripping set — a change about *those* defects.

`R1_COUNTER` deserves a sentence, because it is the one code that round-trips at
one copy *before* this PR and not after, which looks like a regression and is not.
At `Length = 12` the old contents-aligning rule appended exactly four surplus
octets, and those happened to fill this parameter's own four-octet shortfall,
bringing the record to 16. Two defects cancelling. Correcting the padding removes
the compensation and leaves the shortfall visible, which is the right outcome; at
two copies the shortfall sums to 8 and the code round-trips either way.

The single-parameter case is not lost meanwhile — it is asserted directly, and
now positively, by `test_a_hip_packet_carrying_one_parameter_round_trips`.

## `EXPECTED_FAILURES`: exactly one entry moved

**Correction to an earlier revision of this description:** it said the table is
"`**`-free", which is wrong — 30 of its entries are produced by two
`**{f'pcapng-option/{name}': Gap(...) for name in (...)}` comprehensions
(`test_option_roundtrip_unit.py:460` and `:504`), which is exactly why grepping
it cannot enumerate it. It was therefore checked two independent ways instead:

1. All 322 cases run on both trees and the statuses diffed — one line differs,
   `hip-parameter/ENCRYPTED  MISMATCH` → `OK`.
2. The table **imported** from each tree by file path and compared structurally:

```
entries: 45 -> 44
removed : ['hip-parameter/ENCRYPTED']
added   : []
changed : []
```

So:

- **`hip-parameter/ENCRYPTED`: `MISMATCH` → `OK`, entry deleted.** Not deleted
  merely because it passes: its two recorded causes are both genuinely gone (the
  `cipher=` half by #556, the `data`-length half by this PR), and the reasoning
  is kept as prose where the entry was, in this module's existing style.
- Nothing else moved — `changed : []` is the whole check. `hip-parameter/R1_Counter`,
  `hip-parameter/HIP_TRANSFORM` and `hip-parameter/HOST_ID` keep their recorded
  status, fragment and defect string byte for byte.
- **The record of the 6-octet `UInt32Field` pack asymmetry was not touched.**
  Note it is not a `Gap` entry at all but the prose block at
  `test_option_roundtrip_unit.py:205-230`, explaining that `SIDOption.sid` was a
  `UInt32Field` where RFC 791 §3.1 gives the Stream ID 16 bits, so the option
  re-emitted as `880400000037` where the wire holds `88040037`. `git diff
  0c7f2b7c9` over that file matches none of `UInt32Field`, `UInt16Field`,
  `SIDOption` or `880400000037`, i.e. the block is untouched.

`test_a_single_hip_parameter_cannot_be_constructed` is renamed to
`test_a_hip_packet_carrying_one_parameter_round_trips` and inverted — it required
the library to reject its own single-parameter packets, which is the defect.

## Found and deliberately **not** fixed

**`R1_COUNTER`'s counter is four octets where RFC 7401 section 5.2.3 requires
eight.** The RFC diagram reads "R1 generation counter, 8 bytes" with
`Length 12`, and the prose says "a 64-bit unsigned integer in network byte
order"; `R1CounterParameter.counter` is a `UInt32Field`, so the schema declares
`len=12` and packs 12 octets where the RFC total is 16. `HIP_COPIES = 2` hid it,
because two identical 4-octet misalignments sum to 8 — and, before this PR, the
old padding rule hid it a second way by appending exactly the four octets it was
missing. This is the same shape of defect as #608 and wants its own issue and
review — not folded in here. Its consequences show up twice above: it is why
`R1_COUNTER` no longer round-trips at one copy, and why the fixture walker's
"0 violations" conceals one record. **Now filed as #672**, carrying the RFC §5.2.3
quote, the `file:line`, the measured octet counts for both codes 128 and 129, the
three separate things that were hiding it, and the walker reproduction.

**Also not fixed, and also pre-existing: `_make_hip_param`'s `bytes` and `Schema`
branches are unreachable.** Both append *packed octets* to a list that
`Schema_HIP.param` packs as a `ListField` of `SchemaField`s, so
`HIP(parameters=[b'...'])` and `HIP(parameters=[some_schema])` both raise
`AttributeError: 'bytes' object has no attribute 'type'` — identically on
`0c7f2b7c9` and on this branch. Worth recording because it is what makes the
`total_length // 8 + 4` truncation hole unreachable in practice: only the
`(code, kwargs)` branch works, and every parameter it produces is 8-aligned after
this fix. On `0c7f2b7c9` that same one-parameter call raised
`ProtocolError: HIPv2: invalid format`; on this branch it yields a 48-octet packet
with `Header Length = 5`, which is the whole fix in one line.

Also recorded, for the sibling #653/#654/#655 work rather than for this PR:
`hip-parameter/HIP_TRANSFORM`'s `defect` string in `EXPECTED_FAILURES` blames
"the make-side arithmetic", but measured on this branch the parameter packs 8
octets for `len=0`, exactly the RFC total. Its actual cause is that
`_read_param_hip_transform` raises for any `version != 1` while the generator
builds it at version 2 — which sounds like #655's territory, so the entry is
left untouched here.

## Not touched

`protocols/protocol.py`, `application/http.py`, `corekit/io.py`,
`dumpkit/common.py`, `transport/tcp.py`, `tests/_support.py`, `misc/pcapng.py`,
`.github/**`, `docs/source/changelog/1.5.0.rst`, `CHANGELOG.md`. **No changelog
bullet on this branch** — that goes to #657 / `docs/changelog-1.5.0` separately.

## Traps hit while measuring, for the record

`PYTHONSAFEPATH=1` is what *causes* the editable-install trap rather than
avoiding it. `__editable__.pypcapkit-1.4.1.post2.pth` installs an
`_EditableFinder` hard-wired to `MAPPING = {'pcapkit':
'/local/home/jarryx/GitHub/PyPCAPKit/pcapkit'}` — the main checkout — and
`PYTHONSAFEPATH=1` removes the script-directory entry that would otherwise shadow
it. Measured: the first `make_samples.py` run in this session produced an
`options-internet.pcap` carrying pre-fix 20-octet `ESP_INFO` records while the
worktree's own library packed 16, because the generator had silently imported the
main checkout. Every measurement above was redone with `PYTHONPATH=<tree>` and a
printed, asserted `pcapkit.__file__`.

Second, the finder is appended to `sys.meta_path` as a **class**, so
`type(f).__module__` reads `builtins` and a filter on that misses it entirely;
`f.__module__` is what carries the `__editable__` prefix.


JarryShaw:

Independent cross-review of `32b6b65a4` (then at `749fb9a47`) by an agent on a
**different model — Sonnet**, where the change itself was authored on Opus. It was
briefed to falsify rather than to bless: to re-derive RFC 7401 §5.2.1's arithmetic
from the RFC text before reading any code, to hunt for a wrong formula that still
passes the new tests, and to try to find a real HIP capture to check against.

Verdict: **GOOD TO GO**, with four follow-ups, none of them blocking. All four
have been acted on and are folded into `32b6b65a4` and the description above.

### What it confirmed independently

- **The RFC formula and the implementation agree.** It fetched
  `rfc7401.txt` itself (same sha256, `09366b9f…44c4be`, 309,319 octets), derived
  the padding arithmetic from §5.2.1's text *before* reading the code, and then
  compared its own derivation against `parameter_total_len` /
  `parameter_padding_len` for every `Length` 0–255 with zero disagreements,
  spot-checked to 65535. Same formula, not a different one.
- **All 95 sites, and no others.** 46 + 49 identical expressions on the base,
  collapsing to two `% 8` occurrences on the branch (the docstring quote and the
  one `return`). It then swept the whole `pcapkit/` tree, found 18 other `%8`
  hits, and inspected each: all structurally different, already-correct mechanisms
  in IPv6/MH/TCP/pcapng. None shares the HIP pattern.
- **`HIP.make`'s `total_length // 8 + 4` is correct**, §5.1.3 quote checked.
- **The `EncryptedParameter.data` coupling is necessary**, derived independently
  from §5.2.18 and confirmed against real packed bytes on both trees.
- **Coverage, the `HIP_COPIES` table, and the `EXPECTED_FAILURES` diff** all
  reproduced exactly, the last by importing the table from each tree and comparing
  structurally rather than grepping it.
- **House rules**: one commit, `Fixes #651` present, in-library exceptions only,
  `.rst` not `.md`, no `GH-nnn`, no changelog file touched, none of the nine
  off-limits files in the diff.

### What it disputed, and what changed as a result

1. **A prose overstatement in this PR, now corrected.** The description claimed
   the old `ENCRYPTED` total and the RFC's "coincide **only** when
   `Length % 8 == 0`". False — they also coincide at residues 5, 6 and 7. The
   earlier table only sampled residues 0 and 4, which is how it went unnoticed.
   Re-measured at every residue: they agreed at `{0, 5, 6, 7}` and differed at
   `{1, 2, 3, 4}`. Corrected in the description, in `encrypted_data_len`'s
   docstring, in the test docstring and in the `EXPECTED_FAILURES` deletion note —
   and the residue sets are now **asserted** in
   `test_hip_encrypted_data_length_excludes_reserved_and_iv` rather than only
   claimed. The coupling argument is unaffected and in fact stronger: with only
   the padding corrected there is no residue left where the errors cancel.

2. **A real gap in the new tests, now closed.** Eight residues pin any formula
   periodic in `Length % 8`, but the sweep stopped at 63, so a
   `parameter_total_len` masking its argument (`length & 0xFF`) agreed on 0–63,
   diverged at 256, and **passed the entire suite**. The test now also checks the
   whole 16-bit domain of the `len` field. Verified by planting exactly that
   formula in a scratch copy: it now fails naming the first divergence at
   `Length = 256` and the count 65280.

3. **The fixture walker's "0 violations" conceals one record.** This is the most
   substantive finding and I reproduced it myself. `R1_COUNTER` declares
   `Length = 12` but packs 12 octets rather than 16, so a Length-driven reader
   advances 16 and reads a phantom `Type = 0, Length = 0` record out of the second
   copy's zeroed contents — whose "padding" is zero, so the check passes. Patching
   only that copy's counter to `aabbccdd` turns 0 violations into 1
   (`frame 101: type 0 padding not zeroed: aabbccdd`). The padding fix is not
   implicated; the concealment is entirely the `counter` width defect. The
   description now states the caveat explicitly instead of citing "0 violations"
   bare.

4. **File the `R1_COUNTER` 64-bit-width issue.** Agreed and unchanged in
   substance: it is real, pre-existing, out of scope here, and now cited in three
   places above. I have not opened the issue because that was not in my remit;
   it needs one line of go-ahead.

### Two corrections in the other direction

Reported for the record, because a cross-review is only worth having if it is
also checked:

- The review reported the description's `76 failed, 22 passed, 74 subtests
  passed` as wrong, giving `79 failed, 28 passed, 431 subtests passed` and reading
  `74` as a *failed* count. Re-measured with the library md5-checked against
  `git show 0c7f2b7c9:` and the test files md5-checked against this branch: **both
  figures are right and differ only in file selection** — 76/22/74 for
  `test_hip_unit.py` alone, 79/28/431 with the round-trip module added. And `74`
  really is the *passed*-subtest count in the one-module run; it is also,
  coincidentally, the *failed*-subtest count in the two-module run, which is what
  crossed the wires. The description now gives both with a breakdown. The original
  fault was mine: it quoted one figure without saying what it covered.
- One of the review's own sub-agents reported an `R1_COUNTER` regression where a
  standalone packet "succeeds pre-fix, raises post-fix". The lead reviewer
  re-ran it and it did not reproduce, and my own measurement agrees: it raises
  identically on both trees. The real effect is the round-trip-at-one-copy change
  explained above, which is two defects ceasing to cancel rather than a
  regression.

### Could not verify

**No real third-party HIP capture was found.** The Wireshark SampleCaptures wiki
returned HTTP 403, the automated-captures mirror is a fuzzer corpus rather than
curated samples, and the HIPL and OpenHIP source trees carry no capture files. So
the conformance evidence here is the RFC arithmetic and the independent
RFC-only walk over the generated capture, not a packet from a real peer. Stated
plainly rather than glossed: if anyone has a HIP capture, it is the one check this
PR cannot make.

### On the `breaking` label

`breaking` is right, and for a reason independent of the "the old bytes were
plainly wrong anyway" counter-argument: the fix changes `Data_*Parameter.length`,
a **documented public attribute of the parse result**, for every HIP parameter —
a `SEQ` now reports `length = 8` where it reported `12`. Code asserting on that
attribute breaks, and unlike the emitted octets there is no sense in which the old
value was "obviously" wrong to a caller reading the data model. The wire-output
change qualifies on its own too, since this repo's label contract defines
`breaking` as "changes public API **or wire output**".

JarryShaw:
## Interaction with #665 — **not order-independent**, and the reason is in #665's fixtures

Recorded here rather than only in a handoff note, because whoever merges these two
needs it and the conflict resolution is where the fix belongs.

`git merge-tree` between the two heads reports exactly **one** conflict, in
`tests/protocols/internet/test_hip_unit.py`. No library file conflicts: this PR's
46 `PaddingField` lines and 49 `length=` lines sit on different lines from #665's
`lifetime`→`reserved` rename, its `rhash_len` additions and its rewritten
`_make_param_puzzle` / `_make_param_solution` bodies. The textual collision is only
that both PRs append new test methods after the same existing one.

**But the two are coupled on correctness, in both directions, and the coupling is
not visible in the conflict.** A real three-way merge of the two heads was built in
a disposable clone and the suites run against it: `test_option_roundtrip_unit.py`
is clean (6 passed, 360 subtests) and this PR's octet assertions all hold, but
**two of #665's own tests fail** — `test_hip_puzzle_and_solution_keep_the_on_wire_field_width`
and `test_hip_solution_second_octet_is_reserved_not_a_lifetime` — each on
`bytes(rebuilt) == wire`, with `rebuilt` four octets shorter than the fixture. Both
pass on #665 alone.

The cause, verified directly against #665's diff: its hand-written wire literals
encode the padding this PR removes. For example

```python
wire = bytes.fromhex('0141' '0014' '01' '20' '6f70'
                     '0000000000000001' '0000000000000001' '00000000')
self.assertEqual(len(wire), 28)
```

That is a `SOLUTION` with `Length = 0x0014 = 20`, and RFC 7401 §5.2.1 gives
`11 + 20 - (20 + 3) % 8 = 24` octets, not 28 — the trailing `00000000` is the
contents-aligned padding this PR corrects. Same shape in the `PUZZLE` literal
(`Length = 12`, asserted at 20 octets where the RFC gives 16) and in both the
`conformant` and `received` literals of the `Reserved` test. So **#665's fixtures
currently assert non-RFC-conformant octets**, which is not a criticism of #665 —
they are faithful to the tree it was written against — but it does mean the trim
is a correctness improvement there and not merely a merge chore.

Four literals need the trailing `00000000` dropped and their length assertions
moved `28 → 24` and `20 → 16`. A fifth, in
`test_hip_puzzle_lifetime_guard_raises_an_in_library_error`, carries the same stale
padding but is never byte-compared, so it passes either way; worth trimming for
consistency.

**The coupling runs the other way too, and one line of this PR is on the hook.**
`test_hip_parameter_records_are_eight_octet_aligned_on_the_wire` builds its
`SOLUTION` case with `_make_param_solution(..., lifetime=1, ...)`, and it must —
on `main` a `lifetime` of 0 reaches `math.log2(0)`, which is #654. #665 removes
that keyword, after which it is absorbed by `**kwargs` and does nothing. Measured:
the test still passes in the combined tree. But it is a dead keyword at that point
and should be dropped in the same rebase, rather than left depending on `**kwargs`
tolerance.

**Recommended order: this PR first, then rebase #665** and make both edits as part
of that rebase — the four literals and the dead `lifetime=1`. The dependency exists
either way round; taking it in #665's rebase puts it where the failure is
self-explanatory, instead of surfacing later as an unexplained regression blamed on
this PR.

One thing that is **not** an interaction, checked because it looked like one: HIPv1
and HIPv2 state the *same* padding rule. RFC 5201 §5.2.1 gives
`Total Length = 11 + Length - (Length + 3) % 8`, character for character identical
to RFC 7401 §5.2.1, so `parameter_total_len` correctly takes no `version` and
#655's version work does not reach it. Both RFCs were fetched and quoted (RFC 5201
sha256 `8b42d181…9cef6`).

JarryShaw:
## Correction: #664 and #665 **are** order-independent. My earlier comment was wrong.

Retracting the load-bearing claim of
[the interaction note above](https://github.com/JarryShaw/PyPCAPKit/pull/664#issuecomment-5782100395).
Left in place rather than edited away, because what it asserted and on what basis
is worth keeping on the record. #665 carries a mirror-image claim on its own PR
that needs the same correction.

### What I got wrong, and how

**I did not build or run the merge before posting that comment.** I verified the
*premise* — that #665's wire literals carry the pre-#651 padding — directly, by
reading `gh pr diff 665`, and that part is true and still true: the `SOLUTION`
literal is 28 octets for `Length = 20` where RFC 7401 §5.2.1 gives 24, the `PUZZLE`
literal is 20 for `Length = 12` where it gives 16. But I took the *conclusion* — that
this makes two of #665's tests fail in a combined tree — from the interaction
agent's report without running it myself. Verifying the premise and inheriting the
conclusion is the whole of the error, and the arithmetic in that comment being
independently right is exactly what made it look checked.

**The conclusion is false, because #665's author already made those tests
padding-agnostic and said so.** In
`test_hip_puzzle_and_solution_keep_the_on_wire_field_width`:

```python
# Compared without the trailing padding, and then against the re-packed
# source schema rather than against the literal. Both are deliberate: the
# padding rule is itself in flight (#651/#664) ...
self.assertEqual(bytes(rebuilt)[:4 + rebuilt.len], wire[:4 + unpacked.len])
self.assertEqual(bytes(rebuilt), bytes(unpacked))
```

The first assertion slices the padding off; the second compares two schemas
re-packed under whatever rule is in force. Same shape in
`test_hip_solution_second_octet_is_reserved_not_a_lifetime`
(`bytes(rebuilt)[:4 + rebuilt.len]`, `bytes(rebuilt) == bytes(received_schema)`,
`bytes(sanitised) == bytes(conformant_schema)`). So the trailing `00000000` in
those literals is only ever a source buffer to `unpack`, never a comparand, and
#664 cannot move them. **The four literals do not need trimming, and keep-both is
the right resolution.**

### The run I should have done first

Two disposable clones under `/tmp/hip651-mrg2` and `/tmp/hip651-mrg3` (private
paths — an earlier shared one got raced by a concurrent agent), both at the current
pushed heads:

```
pr664 32b6b65a4ab12b887941afc48f2c719795c34d9c
pr665 ebc929cd27d0d2111cf365a03f0b384db1a9f20b
merge-base 0c7f2b7c9c57eaae91704fad5a166ca778c662c8
```

```bash
git clone --shared --no-checkout <worktree> repo
git branch -f pr664 32b6b65a4 && git branch -f pr665 ebc929cd2
git checkout pr665 && git merge --no-ff --no-edit pr664     # and the reverse
```

Both library files auto-merge; one conflict, `tests/protocols/internet/test_hip_unit.py`.
**Resolution: keep-both**, applied by script so it is reproducible — the regex
`^<<<<<<< HEAD\n(.*?)^=======\n(.*?)^>>>>>>> pr664\n` replaced by `group(1) +
group(2)`, i.e. HEAD's block then the incoming block, in that order, nothing else
touched. Union verified rather than assumed:

```
test counts:  merge-base 24   pr664 27   pr665 29   merged 32   (= 24 + 5 + 3)
conflict markers remaining: 0
each of the 8 new test names: exactly 1 occurrence
py_compile: exit 0
```

Provenance asserted in-process before any other import, with `PYTHONPATH` set and
no `PYTHONSAFEPATH` (which is what *causes* the editable-finder trap here, since it
strips the entry that shadows a finder hard-wired to the main checkout):

```
RESOLVED pcapkit.__file__ = /tmp/hip651-mrg2/repo/pcapkit/__init__.py
CPython 3.14.7
OK
```

Exit codes read from a file, not off a pipeline:

```
=== pr665 + pr664 ===
test_hip_unit.py                 exit 0   32 passed, 168 subtests passed
test_option_roundtrip_unit.py    exit 0    6 passed, 360 subtests passed
the four tests in question       exit 0    4 passed,  73 subtests passed
FAILED+SUBFAILED lines: 0 in all three

=== pr664 + pr665 (reverse order) ===
both modules                     exit 0   38 passed, 528 subtests passed
FAILED+SUBFAILED lines: 0
```

`32 passed, 168 subtests` reproduces the coordinator's figure exactly. **Both
orders are clean, so the two PRs are order-independent on correctness and
keep-both is safe in either direction.**

### What survives from the original note

- **The library files do not conflict.** This PR's 46 `PaddingField` lines and 49
  `length=` lines sit on different lines from #665's `lifetime`→`reserved` rename,
  its `rhash_len` additions and its rewritten `_make_param_*` bodies. Confirmed
  again here: both auto-merge.
- **RFC 5201 §5.2.1 states the padding rule identically to RFC 7401 §5.2.1**, so
  `parameter_total_len` is correctly version-independent and #655's `version` work
  does not reach it. Verified from the fetched text (sha256 `8b42d181…9cef6`).
- **One dead keyword, now a tidy-up rather than a coupling.**
  `test_hip_parameter_records_are_eight_octet_aligned_on_the_wire` passes
  `_make_param_solution(..., lifetime=1, ...)`, which it must on this branch —
  `lifetime = 0` reaches `math.log2(0)`, which is #654. In the merged tree
  `lifetime` is gone from the signature and absorbed by `**kwargs`; the test passes
  and draws no warning. Worth deleting in whichever rebase happens, but it breaks
  nothing.

### Why the earlier finding was ever true

Not stale by sha on my side — `32b6b65a4` predates my comment by 18 minutes — but
stale by provenance: the interaction agent measured #664 at `749fb9a47`, and
whatever #665 head it used, the two tests it saw fail were comparing
`bytes(rebuilt) == wire` against the literal. At `ebc929cd2` they do not. So the
finding was real when it was made and had already been fixed at the source by the
time I repeated it. Repeating someone else's failing run as though it were mine is
the thing to not do again.

JarryShaw:
## Cross-review of the narrowing: **GOOD TO GO**

Second cross-review, on a **different model — Sonnet** — against `9f17491ce`, the
first revision to carry the #679 narrowing. The earlier `GOOD TO GO` on this PR
predates the narrowing and does not cover it; the reviewer said so unprompted,
having checked the commit and decision timestamps itself. Briefed to attack the
narrowing specifically: whether `LOCATOR_SET` really is byte-identical to `main`,
and whether the exclusion leaks.

Verdict **GOOD TO GO**, with three findings. All three were real, I reproduced each
independently, and all three are now folded into `45a3e1d68`.

### What it confirmed, by methods other than mine

- **Byte-identity, over 15 construction cases and 3 parse cases** rather than the
  three lengths this PR quoted: plain n=1..6, the SPI/`LocatorData` variant n=1..3,
  a mixed traffic/type/preferred/timedelta case, a mixed SPI+plain case, and the
  empty set — plus `_read_param_locator_set` on a 2-locator mix, a `len`-mismatched
  single locator, and an empty one, comparing the reconstructed data model including
  `.length`. `diff` exit 0, zero differences.
- **The exclusion is structurally incapable of leaking**, which is a better argument
  than my count. `grep -n "item_type=" schema/internet/hip.py` finds 17 `ListField`s,
  and `LocatorSetParameter`'s is the **only** one whose `item_type` is a
  `SchemaField` — i.e. a nested `Schema`, whose own `pack()` does
  `packet.update(self.__dict__)` and so can overwrite the parent's `len`. Every other
  `ListField` uses `EnumField` / `UInt32Field` / `IPv6AddressField`, none of which
  touch the shared context. So no other parameter *can* have this defect.
- It also called the raw `LocatorSetParameter.padding` callback directly over
  `len` 0..39 and confirmed it disagrees with the RFC formula at all 40 values —
  proving the line really is the pre-#651 expression and not a disguised helper.
- **A genuine gap in my evidence, closed.** `HOST_ID` and `HIP_TRANSFORM` never
  appear in `options-internet.pcap` (their builders fail in the generator, `[left
  out]` in both trees' logs), so the walker's "1 violation" never exercised
  `HOST_ID`'s padding at all. It built `HostIDParameter` directly over five
  `hi_len`/`di_len` combinations: all five match the RFC on this branch, all five
  mismatch on `main`. It independently reproduced the `ENCRYPTED` residue sets
  `{0,5,6,7}` / `{1,2,3,4}` by sweeping `dlen` 0..19, and checked the five `% 8`
  lines against the line numbers in this description byte-for-byte.

### The three findings, all reproduced and fixed

1. **My frame-102 explanation was wrong.** I wrote that the walker "advances 8 octets
   into a 32-octet record". It does not: frame 102's parameter area is
   `00 c1 00 00 00 c1 00 00` — **two *empty* LOCATOR_SETs**, `type=193, len=0`, four
   octets each, where `11 + 0 - 3 = 8` is required. The walker advances 8 and reads
   the second copy's header as padding. Corrected here and in the commit message.
2. **"Identical for every locator count" overstated the cancellation.** It holds for
   homogeneous plain-IPv6 sets only. Measured by me on this branch and on
   `a18846c8f`, identically on both:

   | case | declared `len` | packed | mult of 8? | RFC for byte-count `Length` |
   |---|---|---|---|---|
   | empty, n=0 | 0 | **4** | no | 8 |
   | plain, n=1 | 4 | 32 | yes | 32 |
   | plain, n=2 | 8 | 56 | yes | 56 |
   | SPI, n=1 | 5 | **35** | no | 32 |
   | SPI, n=2 | 10 | **63** | no | 64 |
   | plain then SPI | 9 | **59** | no | 56 |
   | SPI then plain | 9 | **60** | no | 56 |

   An SPI locator is 28 octets, not 24, so `24n ≡ 0 (mod 8)` does not apply; and an
   empty set has no locator to shadow `len` at all. Every one of those shapes is
   non-conformant **before and after**, byte-identically — which is precisely why
   leaving the line alone is the *safe* choice rather than the *correct* one. The
   code comment now says exactly that, and #679 has been told it must account for
   all these shapes and not just the homogeneous one.
3. **The suite count was stale.** I quoted "33 passed"; it is 34 on the sha reviewed
   and 39 now (the guard test, plus #665's five arriving in `main`). Re-measured:
   **39 passed / 528 subtests** here against `a18846c8f`'s **35 / 455**.

### One correction in the other direction

The reviewer flagged its 34-vs-my-33 as possibly a plugin difference. It was not:
34 was simply right and my figure predated the guard test I had added. Noting it
because the reviewer was inclined to excuse a discrepancy that was mine.

### Not verified, and said so

No CRUX/AutoSDE-equivalent gate exists on this repo, so there was nothing of that
kind to check, and the reviewer did not re-derive the #665 interaction analysis —
which has since been settled by `main` itself: #665 merged as `c0e9af933`, this
branch is rebased over it, and the append-append conflict in
`tests/protocols/internet/test_hip_unit.py` resolved **keep-both** exactly as
predicted. Union verified: 24 on the old base, 29 with #665, 33 with mine, no name
collisions, and the combined suite is green at 62 passed / 586 subtests.

Issue #652, opened 2026-09-22
_make_http_data never reads frame.flags, so END_STREAM is silently dropped on a parse-reconstruct round trip


JarryShaw:
`_make_http_data` is the only one of the six `_make_http_*` methods that does not restore its flag from the frame object it is handed. It reads back `pad_len` and `data` and silently drops `END_STREAM`, so a DATA frame parsed with that bit set rebuilds with the bit clear. A parse → reconstruct round trip loses it.

Measured on `375e9d411`, CPython 3.14.7, tree asserted against the worktree rather than the editable install. `pcapkit/protocols/application/httpv2.py` is byte-identical on current `origin/main` (`da381f259`).

## The site

`pcapkit/protocols/application/httpv2.py:945-947`:

```python
        if frame is not None:
            pad_len = frame.pad_len
            data = frame.data
```

There is no `end_stream = frame.flags.END_STREAM`. Two lines later the accumulator is seeded and the bit is tested against the *caller's parameter*, which is still at its default because nothing restored it:

```python
        flags = Schema_DataFrame.Flags(0)          # 949
        if end_stream:                             # 950
            flags |= Schema_DataFrame.Flags.END_STREAM
```

Every sibling does restore it — `_make_http_headers:986-988` (`priority`, `end_headers`, `end_stream`), `_make_http_settings:1094` (`ack = frame.flags.ACK`), `_make_http_push_promise:1150`, `_make_http_ping:1186`, `_make_http_continuation:1278`.

## Measured

Parsing a real DATA frame and feeding the resulting data object straight back to `_make_http_data`:

```
parsed flags=0x00 (END_STREAM clear)  info.flags.END_STREAM=False -> rebuilt flags=<Flags: 0>
parsed flags=0x01 (END_STREAM SET)    info.flags.END_STREAM=True  -> rebuilt flags=<Flags: 0>
```

The bit is present on the parse and gone on the rebuild. Control, with the same shape of frame object handed to the siblings:

```
_make_http_settings(frame ACK=True)        -> flags=<Flags.ACK: 1>          ACK kept? True
_make_http_ping(frame ACK=True)            -> flags=<Flags.ACK: 1>          ACK kept? True
_make_http_continuation(END_HEADERS=True)  -> flags=<Flags.END_HEADERS: 4>  kept? True
_make_http_data(frame END_STREAM=True)     -> flags=<Flags: 0>              kept? False
```

[...]

Issue #650, opened 2026-09-22
open question: FrameType.post_process seeds __flags__ with a bare 0, so a flagless HTTP/2 frame leaves it a plain int


JarryShaw:
**Filed as an open question, not an asserted defect.** `FrameType.post_process` seeds its flag accumulator with a bare `0` and promotes it only as a side effect of `|=`, so an HTTP/2 frame with no flag bits set leaves `__flags__` a plain `int` where both the schema and the data model declare it a `Flags`. That is the #616 *shape* exactly, and it reproduces end to end. What it is missing is #616's *impact*: no consumer anywhere in the library, its tests, its docs or its dump output observes the difference. Whether that makes it a defect or pre-emptive hardening is the open question.

Measured on `375e9d411`, CPython 3.14.7, tree asserted against the worktree rather than the editable install. Both files below are byte-identical on current `origin/main` (`da381f259`).

## Correction to where this lives

It is **not** in `pcapkit/protocols/application/httpv2.py` — that file has no `post_process`. It is in the schema:

`pcapkit/protocols/schema/application/httpv2.py:129-147`

```python
    def post_process(self, packet: 'dict[str, Any]') -> 'Schema':
        ...
        flags = 0                                                    # 139
        for key, val in filter(lambda kv: kv[0].startswith('BIT_'),
                               self.Flags.__members__.items()):
            name = key.lower()
            if packet['flags'][name]:
                flags |= val                                         # 144

        self.__flags__ = flags                                       # 146
        return self
```

One accumulator, on `FrameType` (the base schema), inherited by all 11 frame schemas.

## The declared type says `Flags`, and the sibling make path already agrees

- `pcapkit/protocols/schema/application/httpv2.py:122` — `__flags__: 'Flags'` (under `if TYPE_CHECKING:`)
- `pcapkit/protocols/data/application/httpv2.py:39` — `__value__: 'FrameType.Flags'`
- the enum is nested in the schema classes: `FrameType.Flags` at `:126`, per-frame subclasses at `:165`, `:192`, `:276`, `:295`, `:328`, `:374`. Nothing lives under `pcapkit/const/http/` — that package has no flags module.

The decisive asymmetry: **the construct side in the same package already seeds it correctly, six times** — `pcapkit/protocols/application/httpv2.py`:

```
 949        flags = Schema_DataFrame.Flags(0)
 999        flags = Schema_HeadersFrame.Flags(0)
[...]

Issue #668, opened 2026-09-22
Three HTTP/2 payload length callbacks are mis-parenthesised, so an unpadded DATA/HEADERS/PUSH_PROMISE frame parses with its whole payload dropped


JarryShaw:
Three payload length callbacks in the HTTP/2 frame schemas put the conditional expression in the wrong place, so the `else` arm returns `0` — "read no payload at all" — where it was meant to subtract `0`. The result is that an **unpadded** DATA, HEADERS or PUSH_PROMISE frame parses with its entire payload silently discarded. Padded frames are unaffected, and HTTP/2 padding is rare, so this is the common case rather than the edge case.

Measured on `0c7f2b7c9`, CPython 3.14.7, tree asserted against the worktree rather than the editable install (`pcapkit.__file__ = …/.claude/worktrees/agent-a2db5abe44de872e5/pcapkit/__init__.py`).

## The sites

`pcapkit/protocols/schema/application/httpv2.py:195` — `DataFrame.data`:

```python
data: 'bytes' = BytesField(length=lambda pkt: pkt['__length__'] - pkt['pad_len'] if pkt['flags']['bit_3'] else 0)
```

`:237` — `HeadersFrame.fragment`, and `:329` — `PushPromiseFrame.fragment`, both in the parenthesised-but-equivalent form:

```python
fragment: 'bytes' = BytesField(length=lambda pkt: (
    pkt['__length__'] - pkt['pad_len'] if pkt['flags']['bit_3'] else 0
))
```

The outer parentheses in the latter two are line-continuation only; they do not change the grouping.

## Why it parses the way it does

A conditional expression binds looser than `-`, so all three read as:

```python
(pkt['__length__'] - pkt['pad_len']) if pkt['flags']['bit_3'] else 0
```

Confirmed against the AST rather than by eye — the top-level node is the `IfExp`, whose `body` is the subtraction and whose `orelse` is the bare `0`:

```
top-level node: IfExp
 test  = pkt['flags']['bit_3']
 body  = BinOp(pkt['__length__'] - pkt['pad_len'])
 orelse= Constant(value=0)
```

The intent was almost certainly the other grouping — subtract the padding when there is padding, and subtract nothing when there is not:

```python
pkt['__length__'] - (pkt['pad_len'] if pkt['flags']['bit_3'] else 0)
```

`__length__` is the *remaining* declared length at the field, so the `else` arm wants `__length__`, not `0`.

## Measured

[...]

PR #669, opened 2026-09-22
fix(httpv2): restore END_STREAM on rebuild, and seed the parsed flag accumulator as `Flags` (#652, #650)

JarryShaw:
Two adjacent flag defects in the HTTP/2 tier, filed separately and fixed together because they share the two files.

`Fixes #652` — `_make_http_data` was the only one of the six `_make_http_*` methods that never read `frame.flags`, so `END_STREAM` was silently dropped on a parse → reconstruct round trip.

`Fixes #650` — `FrameType.post_process` seeded its flag accumulator with a bare `0`, leaving `__flags__` a plain `int` for a flagless frame where both the schema and the data model declare a `Flags`. #650 was filed as an **open question**, not an asserted defect, on the grounds that no consumer observed the difference. **That turned out to be false, and the measurement below is what settles it.**

Measured on `0c7f2b7c9` (the merge base), CPython 3.14.7, tree asserted against the worktree rather than the editable install:

```
pcapkit.__file__ = …/.claude/worktrees/agent-a2db5abe44de872e5/pcapkit/__init__.py
```

## #652 — before and after

Parsing a DATA frame and feeding the resulting data object straight back to `_make_http_data`:

```
BEFORE
  parsed flags=0x00 (END_STREAM clear) info.flags.END_STREAM=False -> rebuilt flags=<Flags: 0>
  parsed flags=0x01 (END_STREAM SET  ) info.flags.END_STREAM=True  -> rebuilt flags=<Flags: 0>

AFTER
  parsed flags=0x00 (END_STREAM clear) info.flags.END_STREAM=False -> rebuilt flags=<Flags: 0>
  parsed flags=0x01 (END_STREAM SET  ) info.flags.END_STREAM=True  -> rebuilt flags=<Flags.END_STREAM: 1>
```

And through `make`, on the constructed header's flags octet — i.e. what a round trip writes to the wire:

```
BEFORE  original flags octet 0x01 END_STREAM -> rebuilt flags octet 0x00
AFTER   original flags octet 0x01 END_STREAM -> rebuilt flags octet 0x01
```

All five siblings restore theirs on both sides, which is the specification this follows rather than a mechanism invented here:

```
_make_http_settings(ACK=True)     -> <Flags.ACK: 1>
_make_http_ping(ACK=True)         -> <Flags.ACK: 1>
_make_http_continuation(EH=True)  -> <Flags.END_HEADERS: 4>
_make_http_headers(ES|EH=True)    -> <Flags.END_STREAM|END_HEADERS: 5>
_make_http_push_promise(EH=True)  -> <Flags.END_HEADERS: 4>
```

`PADDED` was never affected, being re-derived from `pad_len` rather than read from `frame.flags`.

## #650 — the dump output *does* change, so it is not unobservable

#650's "not established" section says no consumer observes the plain `int`, having found dump output identical on both paths. The cross-review on #634 had found the equivalent TCP fix *did* change dump output, from int `0` to the string `"Flags::None [0]"`. **The same thing happens here.** Real parsed frames, rendered through this library's own `make_dumper`:

```
BEFORE
  flags=0x00 flagless (the COMMON case):  type(__value__) = int
    json  "__value__": 0
    tree  |-- __value__ -> 0
  flags=0x01 END_STREAM set:              type(__value__) = Flags
    json  "__value__": "Flags::END_STREAM [1]"
    tree  |-- __value__ -> Flags::END_STREAM [1]

AFTER
  flags=0x00 flagless:                    type(__value__) = Flags
    json  "__value__": "Flags::None [0]"
    tree  |-- __value__ -> Flags::None [0]
  flags=0x01 END_STREAM set:              unchanged
```

So the observable defect is a **type inconsistency in the dump**, not merely an internal type: `__value__` was a JSON *number* for a flagless frame and a JSON *string* for every other frame, within the same capture. It is consistently a string now. That is the same argument #634 made for TCP's `connection`, and #634 shipped with `fix` + `breaking`.

The `TypeError` goes away too — #616's message verbatim:

```
BEFORE  END_STREAM in __value__ -> TypeError: argument of type 'int' is not a container or iterable
AFTER   END_STREAM in __value__ -> no error
```

`Flags(0) == 0` is `True`, so every int-safe consumer the issue enumerated is unaffected.

The literal `None` in `"Flags::None [0]"` is the pre-existing `pcapkit/dumpkit/common.py:216` defect that #634 already identified and left to its own change; `Flags(0).name is None` for a bitless pseudo-member. Not touched here — that file belongs to another change in flight.

## The one-token fix in the issue would have crashed

#650 proposes "the fix is one token — `flags = self.Flags(0)`". **On Python 3.11 and later that breaks parsing outright for six of the twelve frame schemas.** `FrameType.Flags` declares no members, and from the 3.11 enum rewrite a memberless `enum.Flag` subclass refuses instantiation (measured on 3.14.7):

```
FrameType         Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
UnassignedFrame   Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
PriorityFrame     Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
RSTStreamFrame    Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
GoawayFrame       Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
WindowUpdateFrame Flags is FrameType.Flags         BIT_ members= 0  RAISED TypeError: <flag 'Flags'> has no members
DataFrame         Flags is DataFrame.Flags         BIT_ members= 2  OK  repr=<Flags: 0>
HeadersFrame      Flags is HeadersFrame.Flags      BIT_ members= 4  OK  repr=<Flags: 0>
SettingsFrame     Flags is SettingsFrame.Flags     BIT_ members= 1  OK  repr=<Flags: 0>
PushPromiseFrame  Flags is PushPromiseFrame.Flags  BIT_ members= 2  OK  repr=<Flags: 0>
PingFrame         Flags is PingFrame.Flags         BIT_ members= 1  OK  repr=<Flags: 0>
ContinuationFrame Flags is ContinuationFrame.Flags BIT_ members= 1  OK  repr=<Flags: 0>
```

So the seed is guarded — `self.Flags(0) if self.Flags.__members__ else 0`. The five inheriting schemas keep the plain `int`, which nothing observes because all five pass `flags=None` into their data objects. The test pins both the guard and the reason for it, so it cannot be "simplified" back into a crash.

**That version dependence is real, and locating it took two tries.** CI runs the unit tier on 3.10 through 3.15 (`python -m pytest -q --ignore=tests/integration`), so an unconditional `assertRaises(TypeError)` would fail on any interpreter that does not refuse. The boundary is **3.11**, measured across every interpreter available rather than inferred from a message string:

```
3.8.20   Flags(0) -> OK <Flags.0: 0>
3.9.25   Flags(0) -> OK <Flags.0: 0>
3.10.21  Flags(0) -> OK <Flags.0: 0>
3.11.15  Flags(0) -> TypeError: <flag 'Flags'> has no members defined
3.12.13  Flags(0) -> TypeError: ... has no members; specify `names=()` ...
3.14.7   Flags(0) -> TypeError: ... has no members; specify `names=()` ...
```

Confirmed in CPython's source too. 3.11's `enum.py:1117`, inside `Enum.__new__`, raises `TypeError("%r has no members defined" % cls)` when `not cls._member_map_`, and it runs *before* the `_missing_` hook that manufactured the pseudo-member on 3.10; 3.10's `enum.py` has no such raise, its only "has no members" occurrence being a comment at `:616`. 3.11 is also where the metaclass was renamed — `class EnumType(type)` at `:479` with `EnumMeta = EnumType` at `:1052`, against 3.10's `class EnumMeta(type)` at `:161` — so "the enum rewrite" is the 3.11 release and 3.12 only reworded the message.

**A first pass of this PR gated on `(3, 12)` and was wrong**, having inferred the version from the message text measured on 3.14 rather than from the boundary itself. The cross-review caught it; it is corrected here, in the gate and in both comments. The effect of the error was under-assertion rather than a red build — on 3.11 the block was skipped though 3.11 does raise — but it was also a false statement in a shipped source comment.

The test asserts the memberless-ness unconditionally, that being the predicate the guard actually keys on, and gates only the `TypeError` behind `sys.version_info >= (3, 11)`. The guard in the source is correct on every supported interpreter either way: necessary from 3.11, harmless before.

## These are stdlib `enum`, not `aenum` — verified, not assumed

```
Schema_DataFrame.Flags.__mro__ = (<flag 'Flags'>, <flag 'Flags'>, <flag 'IntFlag'>, <class 'int'>,
                                  <enum 'ReprEnum'>, <flag 'Flag'>, <enum 'Enum'>, <class 'object'>)
issubclass(…, enum.IntFlag)  = True
issubclass(…, aenum.IntFlag) = False
TCP Flags (contrast)         = (<aenum 'Flags'>, <aenum 'IntFlag'>, …)   aenum: True
```

`pcapkit/const/tcp/flags.py` was read, not edited — it is `aenum`, it declares no `_missing_`, and #661 owns const changes.

## Failing, then passing

Both sides run from immutable `git archive` snapshots of `0c7f2b7c9`, with each file hashed. The test file is **byte-identical** across the two runs (`6bfb2b2f…`, the head's own); only the two source files differ.

**Before** — `schema/…/httpv2.py` `6a155ba5…`, `application/httpv2.py` `9e50fe92…`:

```
SUBFAILED(flags='none set')  …::test_a_flagless_frame_seeds_its_flags_as_an_enum
SUBFAILED(end_stream=True)   …::test_make_http_data_restores_end_stream_from_the_frame
FAILED …::test_httpv2_frame_constructors_cover_all_frame_types_and_branches
FAILED …::test_httpv2_schema_selector_and_frame_post_process_flags
4 failed, 2 passed, 27 deselected, 2 warnings, 2 subtests passed in 2.98s
### EXIT CODE FROM FILE: 1
```

with, at the #652 subtest:

```
>       self.assertEqual(bool(rebuilt & DataFrame.Flags.END_STREAM), end_stream)
E       AssertionError: False != True
```

Each new test's *other* subtest passes on that side (`end_stream=False`, `flags='END_STREAM'`), so both fail for the reason claimed rather than incidentally.

**After** — `schema/…/httpv2.py` `745c5aaa…`, `application/httpv2.py` `3a404359…`:

```
4 passed, 27 deselected, 2 warnings, 4 subtests passed in 2.91s
### EXIT CODE FROM FILE: 0
```

Exit codes read from a file, not a pipeline.

## Coverage does not go backwards

The `end_stream` site and the `flags = 0` seed both already executed, so fixing them moves no coverage number; the subtest count is the real evidence. `coverage run --source=pcapkit` over `tests/protocols/application/`:

| | base `0c7f2b7c9` | this PR |
|---|---|---|
| `protocols/application/httpv2.py` | 312 stmts / 122 br, 100% | 313 stmts / 122 br, **100%** |
| `protocols/schema/application/httpv2.py` | 101 stmts / 4 br, 100% | 101 stmts / 4 br, **100%** |
| `protocols/data/application/httpv2.py` | 41 stmts, 100% | 41 stmts, **100%** |
| tests passed | 65 | **67** |
| subtests passed | 35 | **39** |

All three modules stay at 100%. The schema statement count is unchanged because a conditional *expression* is still one statement; both arms are exercised (`DataFrame` and `UnassignedFrame`).

Full tiers, in the worktree: `tests/protocols/application/` **67 passed, 39 subtests, exit 0**; `tests/foundation/registry/test_protocols.py` + `tests/test_docstring_contract.py` **11 passed, 82 subtests, exit 0**.

## `EXPECTED_FAILURES` — nothing moved

`tests/protocols/test_option_roundtrip_unit.py` still passes: **6 passed, 358 subtests, exit 0**, matching #634's baseline exactly. `EXPECTED_FAILURES` was **imported, not grepped** (`**` unpacking hides it) — 45 entries, of which exactly one is HTTP/2:

```
httpv2-frame/PRIORITY -> Gap(status='CONSTRUCT', fragment='HTTP/2: [Type 2] invalid format',
    defect='pcapkit/protocols/application/httpv2.py:572 -- reads length != 9 for a frame
            make() always builds with length 14')
```

Unrelated to flags, still failing in the declared way, **not removed**. The test passing means every entry still fails as declared and nothing else regressed, so no entry became stale.

## Found and deliberately NOT fixed — filed as #668

While measuring #652's round trip the payload came back empty, which turned out to be a **third and larger defect in the same schema file**: three payload length callbacks are mis-parenthesised, so an **unpadded** DATA, HEADERS or PUSH_PROMISE frame parses with its **entire payload silently dropped**.

`pcapkit/protocols/schema/application/httpv2.py:195`, `:237`, `:329` all read

```python
pkt['__length__'] - pkt['pad_len'] if pkt['flags']['bit_3'] else 0
```

A conditional expression binds looser than `-` — confirmed against the AST, the top-level node is the `IfExp` with `orelse=Constant(value=0)` — so the unpadded arm asks for **zero** octets instead of subtracting zero. The intended grouping is `pkt['__length__'] - (pkt['pad_len'] if … else 0)`.

```
AS SHIPPED                            WITH THE CONDITIONAL PARENTHESISED
  LOSS DATA         unpadded  b''       OK DATA         unpadded  b'hello'
  OK   DATA         padded    b'hello'  OK DATA         padded    b'hello'
  LOSS HEADERS      unpadded  b''       OK HEADERS      unpadded  b'frag!'
  LOSS PUSH_PROMISE unpadded  b''       OK PUSH_PROMISE unpadded  b'frag!'
```

`ContinuationFrame.fragment`, `UnassignedFrame.data` and `GoawayFrame.debug` use the plain `pkt['__length__']` with no conditional and are the controls that isolate the shape — CONTINUATION returns `b'frag!'` correctly.

**Filed as #668 and deliberately left out of this PR:** its fix changes parse output for essentially every HTTP/2 capture, so it wants its own review and its own regression tests rather than riding along on a flags change. It is also why the #652 test asserts the **flags octet** rather than whole-frame byte equality — a DATA round trip stays lossy for that separate reason until #668 lands, and claiming byte-level round-trip fidelity here would have been false.

## Labels, and why `breaking`

`fix` + `breaking`. `breaking` is additive and its description is "Alters public API or wire output", and this alters output twice over: the reconstructed DATA frame's **flags octet** changes (`0x00` → `0x01` where END_STREAM was set), and a flagless frame's **dumped `__value__`** changes from the number `0` to the string `"Flags::None [0]"`. Both are corrections of wrong output rather than API changes, but both are visible to anyone diffing dumps or reconstructed bytes — and #634 set the precedent for exactly this second change by carrying both labels.

## Housekeeping

Sample captures were regenerated with `examples/generators/make_samples.py` — a fresh worktree has none, and without them 5 runtime-tier tests in `tests/protocols/application/` fail on `FileNotFoundError` for `http.pcap`, unrelated to this change. `examples/captures/` was **not** deleted; `dhcp.pcapng` and `in.pcap` are committed fixtures. No changelog bullet on this branch — it goes to #657.

No file owned by another open PR or live worker was touched: `protocol.py`, `http.py`, `extraction.py`, `interface/core.py`, `corekit/io.py`, `hip.py`, `logging.py`, `ipv6_route.py`, `dumpkit/common.py`, `transport/tcp.py` and `.github/**` are all untouched.


JarryShaw:
## Cross-review: GOOD TO GO (after one round of NEEDS CHANGES, now fixed)

Independent cross-review on a **different model (Sonnet)**, briefed to falsify rather than bless — a verdict per load-bearing claim, on evidence the reviewer obtained itself rather than from this PR's transcript. Read-only throughout; it never touched the branch. No model substitution was needed.

It ran in two rounds, and **the second round found a real error**, so the honest headline is: GOOD TO GO on the first pass's ten claims, NEEDS CHANGES on the amendment, and the change it asked for has landed.

### Round 1 — GOOD TO GO on all ten claims

| # | Claim | Verdict |
|---|---|---|
| 1 | #652 is real — `END_STREAM` clear on rebuild before the fix | **CONFIRMED** |
| 2 | The fix matches the five siblings' mechanism and placement; nothing else left unrestored | **CONFIRMED** |
| 3 | #650's fix changes dump output `0` → `"Flags::None [0]"` in JSON and Tree | **CONFIRMED** |
| 4 | The naive `self.Flags(0)` crashes for the six memberless schemas; guard correctly shaped | **CONFIRMED** |
| 5 | stdlib `enum.IntFlag`, not `aenum` | **CONFIRMED** |
| 6 | Tests fail without the fix, pass with it, for the claimed reason | **CONFIRMED** |
| 7 | Coverage does not go backwards; tiers green | **CONFIRMED** |
| 8 | No `EXPECTED_FAILURES` entry moved | **CONFIRMED** |
| 9 | #668 correctly diagnosed and correctly scoped out | **CONFIRMED** |
| 10 | `fix` + `breaking` justified | **CONFIRMED** |

It derived #652 from its own `git archive` snapshots rather than reusing anything here:

```
BASE: parsed flags=0x01 info.flags.END_STREAM=True -> rebuilt <Flags: 0>,             wire octet 0x01 -> 0x00
PR:   parsed flags=0x01 info.flags.END_STREAM=True -> rebuilt <Flags.END_STREAM: 1>,  wire octet 0x01 -> 0x01
```

It built its own dumper instances for claim 3 and reached `"__value__": 0` → `"__value__": "Flags::None [0]"` on a real flagless DATA frame. It reproduced the coverage table exactly, including the +1 statement, by running the base snapshot's *own original* test file for a true baseline. It imported `EXPECTED_FAILURES` rather than grepping it, and matched the single `httpv2-frame/PRIORITY` entry verbatim. It AST-parsed #668's three sites itself and confirmed this diff touches none of them.

On the guard-inconsistency question it traced every use site and reported that **the six frame types declaring `Flags` members are exactly the six that surface `'__value__': schema.__flags__`, and the five memberless ones are exactly the five that pass `flags=None` and never read `__flags__` at all** — before and after. So the `Flags`/`int` split the guard produces was already unobservable for those five. It found no house-rule violation and no defect this PR does not already mention.

### Round 2 — NEEDS CHANGES: the version boundary was wrong

Round 1 ran against `f823c57a1`. It did not see the later amendment that gated the empty-enum assertion behind `sys.version_info >= (3, 12)`, so the delta went back for a second pass — and the reviewer found that **the boundary is 3.11, not 3.12.** It was right and this PR was wrong. Re-measured here across every interpreter available:

```
3.8.20   Flags(0) -> OK <Flags.0: 0>
3.9.25   Flags(0) -> OK <Flags.0: 0>
3.10.21  Flags(0) -> OK <Flags.0: 0>
3.11.15  Flags(0) -> TypeError: <flag 'Flags'> has no members defined
3.12.13  Flags(0) -> TypeError: ... has no members; specify `names=()` ...
3.14.7   Flags(0) -> TypeError: ... has no members; specify `names=()` ...
```

and confirmed in CPython's source rather than only behaviourally: 3.11's `enum.py:1117`, inside `Enum.__new__`, raises `TypeError("%r has no members defined" % cls)` when `not cls._member_map_`, *before* the `_missing_` hook that manufactured the pseudo-member on 3.10 — whose `enum.py` has no such raise, its only "has no members" occurrence being a comment at `:616`. 3.11 is also where the metaclass was renamed (`class EnumType(type)` at `:479`, `EnumMeta = EnumType` at `:1052`, against 3.10's `class EnumMeta(type)` at `:161`). So "the enum rewrite" is 3.11; 3.12 only reworded the message, which is precisely what misled the first attempt — the version was inferred from message text measured on 3.14 rather than from the boundary itself.

Consequence of the error, stated plainly: not a red build, since under-gating only skips an assertion that would have passed. But on 3.11 — a version the unit-test matrix runs — the test silently failed to exercise the thing its own comment existed to pin, and the source comment shipped a false statement about CPython. Both are corrected at the head (`1b9388d43`): `(3, 12)` → `(3, 11)` in the gate and in both comments, with the six-interpreter measurement written into the test comment so the boundary is pinned by evidence rather than by memory. `tests/protocols/application/` is **67 passed, 39 subtests, exit 0** after the correction, and the failing-then-passing evidence in the PR body was re-run against the head's own test file (`6bfb2b2f…`).

The reviewer independently re-ran the before/after on the amended head and reproduced it, and confirmed both tiers green there.

### The `NONE = 0` alternative — a genuine disagreement, left unresolved and non-blocking

Round 1 raised an alternative this PR had not discussed: give `FrameType.Flags` a canonical `NONE = 0` member so `self.Flags(0)` succeeds for all twelve classes and the guard disappears. We converged on the mechanism and disagreed on the verdict, so both readings are recorded rather than one folded away.

Agreed facts. A zero-valued member of a `Flag` is not canonical, which is exactly why it would not trip the "cannot extend" check:

```
Flag base     _member_names_ : []          <- NONE = 0 is not canonical
IntEnum       _member_names_ : ['ZERO']    <- and subclassing an IntEnum does raise:
IntEnum with a member, subclassed -> TypeError: <enum 'ChildOfPlain'> cannot extend <enum 'Plain'>
```

And the reviewer confirmed against the real classes that the five memberless schemas do not subclass `Flags` at all — `UnassignedFrame.Flags is FrameType.Flags` is `True` — so a base `NONE = 0` would reach them, while the six with their own nested subclass would keep `.name is None`:

```
BaseWithZero(0) = <BaseWithZero.NONE: 0>   name = 'NONE'
Child(0)        = <Child: 0>               name = None
```

**My reading:** that is a *new* rendering split across frame types — `"Flags::NONE [0]"` for five, `"Flags::None [0]"` for six — replacing the flagless/flagged split this PR closes, and it invents a flag name RFC 9113 does not define.

**The reviewer's push-back, which I accept as a correction:** that split is not observable through the dump path the `breaking` label is actually about, because all five memberless types discard `__flags__` into `flags=None` before any dumper sees it. So on the surface that matters, the two approaches are identical, and it called my "worse rather than partial" framing an overstatement. It also found a wrinkle neither of us had flagged: `proto.__header__`'s raw schema `repr()` — not the `Data`/`Info` dump — already shows an int-vs-`Flags` split across frame types under the *shipped* fix (`DataFrame.__flags__` → `<Flags: 0>`, `UnassignedFrame.__flags__` → `0`), so neither approach is clean on that internal surface. Its verdict there is "a wash", not a point against the alternative.

Both of us land on not taking the alternative in this PR either way, since it would touch the `dumpkit`-adjacent `.name` semantics deliberately scoped out here. Recording the disagreement because the reviewer's narrowing is the more accurate claim and my original framing was too strong.

PR #671, opened 2026-09-22
fix(tcp): drop the doubled separator from four MPTCP error messages (#649)

### LINKED ISSUE #649: four MPTCP ProtocolError messages carry a doubled separator, rendering as 'TCP: : [OptNo 30] 1: ...'
Four `ProtocolError` messages in the MPTCP option handlers carry a doubled `: ` separator — `f'{self.alias}: : [OptNo …]'` — so they render as `TCP: : [OptNo 30] 1: invalid flags combination` with an empty field between the protocol alias and the option number. Cosmetic, but it is the string a user sees when an MP_JOIN or DSS option is rejected.

Measured on `375e9d411`; `pcapkit/protocols/transport/tcp.py` is byte-identical on current `origin/main` (`da381f259`).

## The sites — all four, all in `pcapkit/protocols/transport/tcp.py`

| `file:line` | Enclosing method | Message |
|---|---|---|
| `tcp.py:1564` | `_read_mptcp_join` (def `:1544`) | `invalid flags combination` |
| `tcp.py:2818` | `_make_mptcp_join` (def `:2800`) | `invalid flags combination` |
| `tcp.py:2984` | `_make_mptcp_dss` (def `:2944`) | `missing required fields` |
| `tcp.py:2986` | `_make_mptcp_dss` (def `:2944`) | `missing required fields` |

Verbatim, `tcp.py:1564`:

```python
        raise ProtocolError(f'{self.alias}: : [OptNo {schema.kind}] {schema.subtype}: invalid flags combination')
```

and `tcp.py:2818`:

```python
        raise ProtocolError(f'{self.alias}: : [OptNo {Enum_Option.Multipath_TCP}] {subtype}: invalid flags combination')
```

`grep -rn "alias}: : " pcapkit/` returns nothing outside this file, so the defect is confined to these four lines.

## Measured

Rendering the f-strings with the real values they are given — `self.alias == 'TCP'`, `kind == Option.Multipath_TCP` (30), `subtype == MPTCPOption.MP_JOIN` (1):

```
tcp.py:1564 -> 'TCP: : [OptNo 30] 1: invalid flags combination'
tcp.py:2818 -> 'TCP: : [OptNo 30] 1: invalid flags combination'
tcp.py:2984 -> 'TCP: : [OptNo 30] 1: missing required fields'
```

## Why it is a defect rather than a style preference

The rest of the file is consistent and does not do this. Counting the alias-prefixed messages in `tcp.py`:

```
doubled  '{self.alias}: : '  :  4
single   '{self.alias}: '    : 28
```

The 28 all read `f'{self.alias}: [OptNo {schema.kind}] …'` — e.g. `tcp.py:1527`, rendering as:

```
'TCP: [OptNo 30] invalid format'
```

So the four are the outliers against a 28-site house form in the same file.

## Consequence

Purely presentational: the exception type, the option number and the subtype are all correct, and nothing downstream parses these strings. The cost is that a user reading `TCP: : [OptNo 30] 1:` sees an empty field and may reasonably wonder which value failed to interpolate.

Note the test that exercises the read-path message does not pin the prefix — `tests/protocols/transport/test_tcp_mptcp_join_flag_ordering_unit.py:404` asserts only `self.assertIn('invalid flags combination', str(caught.exception))` — so fixing the four lines needs no test change. Other tests quote the *correct* single-separator form in their docstrings (e.g. `tests/protocols/transport/test_tcp_mptcp_subtype_unit.py:279`, `tests/protocols/transport/test_tcp_mptcp_capable_length_unit.py:195`), which is further evidence of the intended shape.

## Notes

- All four lines were introduced together in `3bba8a1748` (2023-04-10, *"revised TCP schemas with OptionField/etc. & redesigned read funcs"*), per `git blame`. Pre-existing since then and untouched by any later change.
- Reported as a side-finding while working on #616/#634 in this file; filed separately because it is unrelated to the flag accumulator's type and wants a different (trivial) fix.
- Originally reported against `_read_mptcp_join` alone; the other three were found by grepping for the same pattern and should be fixed in the same pass.

JarryShaw:
Fixes #649.

**Cosmetic.** No behavioural change: the exception type, the option number and the subtype were always correct, and nothing downstream parses these strings. What changes is the rendered text a user sees when an MP_JOIN or DSS option is rejected.

## The defect

Four `ProtocolError` messages in the MPTCP option handlers were spelled `f'{self.alias}: : [OptNo …]'`, so they rendered with an empty field between the protocol alias and the option number:

```
TCP: : [OptNo 30] 1: invalid flags combination
```

An empty field reads as a value that failed to interpolate, which sends a reader looking for a missing variable that was never there.

## All four sites

All in `pcapkit/protocols/transport/tcp.py`. Line numbers had shifted from the issue's by 13, so they are given as found:

| line | method | before | after |
|---|---|---|---|
| `:1577` | `_read_mptcp_join` | `TCP: : [OptNo 30] 1: invalid flags combination` | `TCP: [OptNo 30] 1: invalid flags combination` |
| `:2831` | `_make_mptcp_join` | `TCP: : [OptNo 30] 1: invalid flags combination` | `TCP: [OptNo 30] 1: invalid flags combination` |
| `:2997` | `_make_mptcp_dss` | `TCP: : [OptNo 30] 2: missing required fields` | `TCP: [OptNo 30] 2: missing required fields` |
| `:2999` | `_make_mptcp_dss` | `TCP: : [OptNo 30] 2: missing required fields` | `TCP: [OptNo 30] 2: missing required fields` |

`30` is `Option.Multipath_TCP`, `1` is `MPTCPOption.MP_JOIN` and `2` is `MPTCPOption.DSS`; all three format as their integer value. The before/after strings above are rendered by execution, not read off the f-strings.

## Why this form, and not a preference

The correct form is the local majority in the very same file:

```
doubled  f'{self.alias}: : '   :  4  ->  0
single   f'{self.alias}: [OptNo':  28  -> 32
```

The four were outliers against a 28-site house form in their own module, and the fix moves them into it. `grep -rn "alias}: : " pcapkit/` returns nothing outside this file, so the defect is confined to these four lines and is now gone from the package entirely. Two sibling test modules already quote the *correct* single-separator form in their docstrings (`test_tcp_mptcp_subtype_unit.py:279`, `test_tcp_mptcp_capable_length_unit.py:195`), which is further evidence of the intended shape.

## The test, and why a substring assertion would not have done

New module `tests/protocols/transport/test_tcp_mptcp_error_message_unit.py`, 4 tests / 2 subtests. Every message is asserted **in full**, and that is the point of the change. **Two** existing assertions already exercise these messages and neither could have caught this, because both match a substring that straddles the defect:

```python
# test_tcp_mptcp_join_flag_ordering_unit.py:404  (_make_mptcp_join)
self.assertIn('invalid flags combination', str(caught.exception))

# test_tcp_udp_unit.py:1629  (_read_mptcp_join)
with self.assertRaisesRegex(ProtocolError, 'invalid flags combination'):
```

Both pass identically before and after this fix. **Both modules are left untouched** — a new module rather than an edit to either, given both have been repaired recently (#627, #628, #634).

*(Corrected after cross-review: an earlier draft of this description named only the first of the two.)*

Four things the module does deliberately:

- **Reaches each of the two `_make_mptcp_dss` guards separately** — `dsn` without its data-level fields, and a data-level field without a `dsn`. They carry the identical message, so asserting the string once would not show that both lines were fixed.
- **Goes through the public constructor** for the two `_make_*` sites, `TCP(syn=…, ack=…, options=[(Option.Multipath_TCP, …)], …)`, rather than poking a bare `object.__new__(TCP)`.
- **Parses a real flagless segment** for `_read_mptcp_join` rather than hand-writing `_flags`, for the reason #603/#612 give. That branch is not reachable from a caller — `mptcp_data_selector` rejects a flagless MP_JOIN with a `FieldError` first — so it is latent, but latent only by virtue of a guard in a different file, and the string is still what a caller reaching it sees. Noted in the test rather than left to be rediscovered.
- **Sweeps the source** for the pattern and pins the count at 32, so a fifth site cannot be added unnoticed *and* "fixing" the four by deleting them would not pass. Three of the four sites were themselves found by grepping rather than by reading the one that was reported, which is precisely why the absence is asserted over the whole file. That exact count is knowingly a tripwire on the whole ~3000-line module rather than on these four lines, so the assertion carries a note telling a future editor to update the number rather than loosen it; the `== 0` on the doubled form above it is the one that must hold forever. Raised by the cross-review as a footgun and kept deliberately, with the warning added.

## Failing, then passing

Same harness as its sibling PR: `__editable__*` stripped from `sys.meta_path`, worktree at `sys.path[0]`, `pcapkit.__file__` asserted and printed before any other import, and pytest's exit code **read from a file** rather than a pipeline.

```
MEASURED pcapkit.__file__ = …/.claude/worktrees/agent-a3c59f390d4ed0e74/pcapkit/__init__.py
```

**On `main` (fix reverted, test present):**

```
AssertionError: 'TCP: : [OptNo 30] 1: invalid flags combination'
             != 'TCP: [OptNo 30] 1: invalid flags combination'
AssertionError: 'TCP: : [OptNo 30] 2: missing required fields'
             != 'TCP: [OptNo 30] 2: missing required fields'
AssertionError: "f'{self.alias}: : " unexpectedly found in <source>

5 failed, 1 passed
exit code (from file): 1
```

All four tests fail, covering all four sites.

**With the fix:**

```
4 passed, 2 subtests passed
exit code (from file): 0
```

**Scoped directory run**, `tests/protocols/transport/` unit tier:

```
140 passed, 105 subtests passed
exit code (from file): 0
```

The full directory additionally shows 7 failures, all in `test_tcp_runtime.py` / `test_udp_runtime.py`, all `FileNotFoundError` for generated captures in a tree where `make samples` has not been run. That is the fixture-dependent tier documented in `tests/_tiers.py`, and it is unaffected by this change.

The suite was never run whole; every run was scoped.

## Coverage

`pcapkit/protocols/transport/tcp.py`, statement and branch:

| | statements | branches | cover | tests | subtests |
|---|---|---|---|---|---|
| before | 594 | 198 | **100%** | 136 | 103 |
| after | 594 | 198 | **100%** | 140 | 105 |

The four changed lines already executed, so per the usual rule the subtest count is the number that moves: **103 → 105**, with tests **136 → 140**. The statement and branch counts are necessarily unchanged because the change is to string literals inside existing `raise` statements, and coverage holds at 100%.

## Provenance

All four lines arrived together in `3bba8a1748` (2023-04-10, *"revised TCP schemas with OptionField/etc. & redesigned read funcs"*) and were untouched by any later change.

## Label

`fix` alone. Not `breaking`: no public API, wire output or dump output changes — only the text of an exception message, and nothing in the package or the tests parses it.

## Cross-review

Dispatched on a different model, briefed to falsify rather than bless; verdict posted as a comment below.


JarryShaw:

Independent cross-review, run on a **different model (Sonnet)** from the one that authored the change, in its own worktree, briefed to falsify rather than bless and to treat the issue's own measurements as claims rather than ground truth. It was read-only with respect to this PR: no amend, no push, no commit to the branch.

**Tree it measured:** `…/.claude/worktrees/agent-a5561bffde954eb30/pcapkit/__init__.py`. Worth recording how it got there, because it found a sharper version of the editable-install trap than the brief described: the finder is registered as a **bare class**, so `type(f).__name__` reads `'type'` rather than anything containing `__editable__`, and only checking `f.__name__` / `f.__module__` identifies it. It confirmed the finder maps `pcapkit` to the **main checkout** (still carrying the pre-fix doubled separator) and that `PathFinder` wins only because pytest's prepend import mode puts the worktree at `sys.path[0]`.

### Per-claim verdicts

| claim | verdict | evidence it obtained |
|---|---|---|
| **A** four sites, now zero, none missed repo-wide | TRUE | `source.count("f'{self.alias}: : ")` = 4 before, 0 after. Broader sweeps for `: : ` and `\}: : ` across `pcapkit/`, `tests/`, `docs/`, `examples/` found nothing outside the new test's own documentary literals. No near-miss variant anywhere. |
| **B** the exact before/after strings, by execution | TRUE | Confirmed `Option.Multipath_TCP`→`30`, `MPTCPOption.MP_JOIN`→`1`, `MPTCPOption.DSS`→`2`, and reproduced both messages by triggering the real `ProtocolError`s — via the public constructor and via a parsed flagless segment — not by reading the f-strings. |
| **C** no space lost after `OptNo`, nothing else touched | **TRUE, byte-for-byte** | Against the PR's true recorded base (`baseRefOid` `0c7f2b7c9`): exactly 4 changed lines in `pcapkit/`, at 1577 / 2831 / 2997 / 2999. Each line shrinks by exactly 2 bytes, the sole divergence is `"}: : ["` → `"}: ["`, and `before.replace("self.alias}: : ", "self.alias}: ")` reconstructs `after` exactly — `True` for all four. **No missing space, no doubled space, no other line altered.** |
| **D** 28→32 and 4→0 | TRUE | Derived independently on both trees. Matches the test's hardcoded `32`. Flagged as a footgun — see below. |
| **E** fails on `main`, passes with the fix | TRUE, exact match | Pre-fix: **5 failed / 1 passed**, exit file `1`. Post-fix: **4 passed / 2 subtests**, exit file `0`. Restored its worktree immediately after; `git status` clean. |
| **F** all four `raise` lines actually execute | **TRUE — the important one** | `coverage run --branch` against the new module alone: lines 1577, 2831, 2997 **and** 2999 all `EXECUTED`. Both `_make_mptcp_dss` guards fire independently despite emitting an identical string. |
| **G** the pre-existing substring assertion could not have caught this | TRUE, **and incomplete** | `…join_flag_ordering_unit.py:404` is exactly as quoted. But it found a **second** such assertion I had not named — see below. Confirmed neither file was touched. |
| **H** the source-sweep test is robust | TRUE | A read failure raises rather than passing vacuously; an empty read would still fail `== 32`; and (via E) it genuinely fails on unfixed code. Its dependence on `tcp_module.__file__` resolving to the worktree is a pre-existing property of every test in this suite, not introduced here. |
| **I** nothing else in the repo depends on these strings | TRUE | Swept `invalid flags combination` / `missing required fields` repo-wide: only the four `raise` sites, the two substring assertions, and the new test's literals. No `.rst` or docstring quotes the old doubled form. |
| **J** `fix` alone, not `breaking` | Agreed | An exception's `str()` is not a documented output contract and nothing in-repo parses it. It explicitly contrasted this with sibling #670, which *does* carry `breaking` because it changes dumped JSON/Tree/PLIST — a format downstream tooling parses. It judged the two cases legitimately different and the repo's own convention to support both self-assessments. |

Structural checks also passed: one commit, author `Jarry Shaw <jarryshaw@icloud.com>`, `Fixes #649`, `ProtocolError` from `pcapkit.utilities.exceptions`, only `.py` touched, no `GH-nnn`. Directory run reproduced my own numbers exactly: 140 passed / 105 subtests, with 7 `FileNotFoundError` failures in the fixture-dependent tier.

Both findings acted on rather than argued away. The commit was amended and the branch force-pushed; the fix itself is unchanged.

1. **My narrative said "the existing assertion" when there are two.** Besides `…join_flag_ordering_unit.py:404` (`assertIn`, against `_make_mptcp_join`) there is `test_tcp_udp_unit.py:1629` (`assertRaisesRegex(ProtocolError, 'invalid flags combination')`, against `_read_mptcp_join`) — the same limitation, equally unable to catch this. I verified both exist before accepting it. **The module docstring, the commit message and the PR body now name both.** Neither file is touched.

2. **The hardcoded `32` is a tripwire on the whole module, not on these four lines.** The reviewer called it "a minor NEEDS-CHANGES-caliber nit rather than a blocker": any future change that adds or removes any `f'{self.alias}: [OptNo …]'` message in this ~3000-line file will fail the assertion for an unrelated reason. I kept the exactness deliberately — asserting only the absence would pass if the four lines were deleted outright, which is the failure mode it guards — but **added a note at the assertion** telling a future editor to update the number rather than loosen it, and marking the `== 0` on the doubled form as the one that must hold forever.

It also observed that this branch looked 5 commits behind `main`. **That one I checked and did not act on:** `origin/main` is still `0c7f2b7c9`, exactly the commit this branch is cut from. The `8cfd6ab01` it compared against is a *local, unpushed* `main`, so no rebase is needed and GitHub reports the PR mergeable.

Nothing in A–J was left unverified.

